PR #2, opened 2016-08-15
span context

oibe:
- added span_context and associated tests
    - added configuation files and package.json that will be needed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,15 +19,13 @@ services:
 
 env:
   global:
-  - DOCKER_VERSION=1.11.1-0~trusty
+  - DOCKER_VERSION=1.12.0-0~trusty
   - DOCKER_COMPOSE_VERSION=1.7.0
   - COMMIT=${TRAVIS_COMMIT::8}
 
 before_install:
-- apt-cache madison docker-engine
-- sudo apt-get -o Dpkg::Options::="--force-confnew" install -y docker-engine=${DOCKER_VERSION}
 - docker version
-- sudo rm /usr/local/bin/docker-compose
+- sudo rm -rf /usr/local/bin/docker-compose
 - curl -L https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-`uname
   -s`-`uname -m` > docker-compose
 - chmod +x docker-compose
--- a/src/span_context.js
+++ b/src/span_context.js
@@ -1,0 +1,180 @@
+// @flow
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import Int64 from 'node-int64';
+import Utils from './util.js';
+
+let SAMPLED_MASK = 0x1;
+let DEBUG_MASK = 0x2;
+
+export default class SpanContext {
+    _traceId: any;
+    _spanId: any;
+    _parentId: any;
+    _flags: number;
+    _baggage: any;
+    _baggageHeaderCache: any;
+
+    constructor(traceId: any,
+                spanId: any,
+                parentId: any,
+                flags: number,
+                baggage: any = {}) {
+        this._traceId = traceId;
+        this._spanId = spanId;
+        this._parentId = parentId;
+        this._flags = flags;
+        this._baggage = baggage;
+    }
+
+    get traceId(): any {
+        return this._traceId;
+    }
+
+    get spanId(): any {
+        return this._spanId;
+    }
+
+    get parentId(): any {
+        return this._parentId;
+    }
+
+    get flags(): number {
+        return this._flags;
+    }
+
+    get baggage(): any {
+        return this._baggage;
+    }
+
+    static getBaggageHeaderCache() {
+        if (!this._baggageHeaderCache) {
+            this._baggageHeaderCache = {};
+        }
+
+        return this._baggageHeaderCache;
+    }
+
+    /**
+     * Returns a normalize key.
+     *
+     * @param {string} key - The key to be normalized for a particular baggage value.
+     * @return {string} - The normalized key (lower cased and underscores replaced, along with dashes.)
+     **/
+    // todo(oibe) moved to span in newest opentracing version
+    _normalizeBaggageKey(key: string) {
+        let baggageHeaderCache = this.constructor.getBaggageHeaderCache();
+        if (key in baggageHeaderCache) {
+            return baggageHeaderCache[key];
+        }
+
+        // This stackoverflow claims this approach is slower than regex
+        // http://stackoverflow.com/questions/1144783/replacing-all-occurrences-of-a-string-in-javascript
+        let normalizedKey: string = key.toLowerCase().split('_').join('-');
+
+        if (Object.keys(baggageHeaderCache).length < 100) {
+            baggageHeaderCache[key] = normalizedKey;
+        }
+
+        return normalizedKey;
+    }
+
+    /**
+     * Sets a baggage value with an associated key.
+     *
+     * @param {string} key - The baggage key.
+     * @param {string} value - The baggage value.
+     **/
+    // todo(oibe) moved to span in newest opentracing version
+    setBaggageItem(key: string, value: string): void {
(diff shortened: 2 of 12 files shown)


src/thrift.js
```diff
@@ -0,0 +1,43 @@
...
+
+export const CLIENT_SEND = 'cs';
+export const CLIENT_RECV = 'cr';
+export const SERVER_SEND = 'ss';
+export const SERVER_RECV = 'sr';
+export const WIRE_SEND = 'ws';
+export const WIRE_RECV = 'wr';
+export const CLIENT_SEND_FRAGMENT = 'crf';
+export const SERVER_SEND_FRAGMENT = 'ssf';
+export const SERVER_RECV_FRAGMENT = 'srf';
```
yurishkuro: when would we use 26-30?

yurishkuro: or is it auto-generated?

src/span_context.js
```diff
@@ -0,0 +1,139 @@
...
+        // jshint validthis:true
+        if (Object.keys(this._baggageHeaderCache).length < 100) {
+            this._baggageHeaderCache[key] = normalizedKey;
+        }
+        // jshint validthis:false
+
+        return normalizedKey;
+    }
+
+    setBaggageItem(key: string, value: string): void {
```
yurishkuro: add todo to move these to Span, looks like JS API has not been upgraded to immutable Span Context. I asked ben and bcronin if they're planning to do that, if not we may need to do a PR to OT, as I do not want to release Node client that needs to change the API again shortly.

yurishkuro:
There's a bit odd assortment of files, e.g. partial thrift factories, but no span factory. The main question I have is what data type we're getting from Thrift for trace ID.

src/thrift.js
oibe: They should be removed i copied them from zipkin thrift file thinking we would need them.

oibe:
When we want to serialize to thrift we have been using es5 Buffers.  When we convert from string to id we convert to a Buffer.  Es6 may have better types for this, but I want to verify which types can be reported through a  udp/thrift binary protocol reporter first.

src/span_context.js
```diff
@@ -0,0 +1,163 @@
...
+    constructor(traceId: any,
+                spanId: any,
+                parentId: any,
+                flags: number) {
+        this._traceId = traceId;
+        this._spanId = spanId;
+        this._parentId = parentId;
+        this._flags = flags;
+        this._baggage = {};
+        this._baggageHeaderCache = {};
```
yurishkuro: cache only makes sense as static cache, but this looks like an instance variable, which will be unique for each instance of SpanContext

src/interfaces/mocha.js
```diff
@@ -0,0 +1,27 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+declare var it: Function;
+declare var describe: Function;
+declare var beforeEach: Function;
+declare var before: Function;
+declare var afterEach: Function;
+declare var after: Function;
```
yurishkuro: what are these for?

src/span_context.js
```diff
@@ -0,0 +1,163 @@
...
+    getBaggageItem(key: string): string {
+        let normalizedKey = this._normalizeBaggageKey(key);
+        return this._baggage[normalizedKey];
+    }
+
+    /**
+     * @return {boolean} - returns whether or not this span context was sampled.
+     **/
+    IsSampled() {
+        return !!(this.flags & constants.SAMPLED_MASK);
```
yurishkuro: what's with double negation?

yurishkuro: a preferred way to check bitmasks is `value & mask == mask`

src/decls/types.js
```diff
@@ -0,0 +1,49 @@
...
+declare type BinaryAnnotation = {
+    key: string,
+    value: any,
+    annotationType: string,
+    host?: Endpoint
+};
+
+declare class Span {};
+
+declare type spanFields = {
```
yurishkuro: what are these? they don't look like anything in the data model

test/span_context.js
```diff
@@ -0,0 +1,101 @@
...
+            );
+        assert.isOk(context.IsSampled());
+
+        context._flags = 0;
+        assert.isNotOk(context.IsSampled());
+    });
+
+    it ('format strings properly with toString', () => {
+        let ctx1 = new SpanContext(Utils.encodeInt64(0x100), Utils.encodeInt64(0x7f), null, 1).toString();
+        assert.equal(ctx1, '0000000000000100:000000000000007f:0:1');
```
yurishkuro: in other languages we do not pad IDs with zeros.

test/span_context.js
```diff
@@ -0,0 +1,101 @@
...
+    it ('normalized key correctly', () => {
+        let context = new SpanContext();
+        let unnormalizedKey = 'SOME_KEY';
+        let key = context._normalizeBaggageKey(unnormalizedKey);
+
+        assert.equal(key, 'some-key');
+        assert.isOk(unnormalizedKey in context._baggageHeaderCache);
+    });
+
+});
```
yurishkuro: add tests for large numbers

src/span_context.js
oibe: This is not double negation, but the operator to cast to a boolean instead of returning a number.

.travis.yml
```diff
@@ -19,7 +19,7 @@ services:
 
 env:
   global:
-  - DOCKER_VERSION=1.11.1-0~trusty
+  - DOCKER_VERSION=1.12.0-0~trusty
```
yurishkuro: I cannot comment on lines below, but I don't think you need lines 27-28, because Travis already comes with docker installed (in fact the attempt to reinstall was failing for me in other PRs). We only need to install docker-compose, and only because we want a later version.

yurishkuro: change line 30 to `rm -rf`

src/constants.js
```diff
@@ -0,0 +1,24 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+export const SAMPLED_MASK = 0x1;
+export const DEBUG_MASK = 0x2;
```
yurishkuro: this are only needed in span.js, we do not need to expose them as public

src/constants.js
```diff
@@ -0,0 +1,24 @@
...
+export const TRACER_STATE_HEADER_NAME = 'UBER-TRACE-ID';
+export const TRACER_BAGGAGE_HEADER_PREFIX = 'UberCtx-';
```
yurishkuro: these are only needed in propagation/codecs, and we do not want to expose them as public

src/span_context.js
```diff
@@ -0,0 +1,170 @@
...
+        let normalizedKey = this._normalizeBaggageKey(key);
+        return this._baggage[normalizedKey];
+    }
+
+    /**
+     * @return {boolean} - returns whether or not this span context was sampled.
+     **/
+    IsSampled() {
+        return this.flags & constants.SAMPLED_MASK === constants.SAMPLED_MASK;
+    }
```
yurishkuro: also add IsDebug

src/span_context.js
```diff
@@ -0,0 +1,170 @@
...
+     **/
+    static fromString(serializedString: string): any {
+        var bufferValueOne: any = Utils.encodeInt64(1);
+
+        let headers:  any = serializedString.split(':');
+        if (headers.length !== 4) {
+            return null;
+        }
+
+        let traceId: any = Utils.encodeInt64(headers[0]);
```
yurishkuro: can this method throw exceptions? let's document, we'll need to revisit in the tracer.extract()

src/span_context.js
```diff
@@ -0,0 +1,170 @@
...
+
+    constructor(traceId: any,
+                spanId: any,
+                parentId: any,
+                flags: number) {
+        this._traceId = traceId;
+        this._spanId = spanId;
+        this._parentId = parentId;
+        this._flags = flags;
+        this._baggage = {};
```
yurishkuro: baggage should be passed in

test/span_context.js
```diff
@@ -0,0 +1,114 @@
...
+// THE SOFTWARE.
+
+import {assert} from 'chai';
+import bufferEqual from 'buffer-equal';
+import * as constants from '../src/constants.js';
+import SpanContext from '../src/span_context.js';
+import Utils from '../src/util.js';
+
+// 2^53 is the biggest number in javascript
+const JS_MAX_NUMBER =  0x1fffffffffffff;
```
yurishkuro: why do we care what the largest number is in JS?  We never represent these as numbers, but only as strings or byte buffers. The test must deal with max numbers that can come from other tracers, i.e. all 64bits

src/util.js
```diff
@@ -0,0 +1,68 @@
...
+     **/
+    static encodeInt64(numberValue: any): any {
+        return new Int64(numberValue).toBuffer();
+    }
+
+    /**
+     * @param {string} input - the input for which leading zeros should be removed.
+     * @return {string} - returns the input string without leading zeros.
+     **/
+    static removeLeadingZeros(input: string): string {
```
yurishkuro: 😭

src/util.js
```diff
@@ -0,0 +1,68 @@
...
+     * @param {string} prefix - the prefix to search for in the text given.
+     * @return {boolean} - boolean representing whehter or not the
+     * string contains the prefix.
+     **/
+    static startsWith(text: string, prefix: string): boolean {
+        return text.indexOf(prefix) === 0;
+    }
+
+    /**
+     * @param {string|number} numberValue - a string or number to be encoded
```
yurishkuro: Should this ever be a number, given that JS cannot represent more than 53bits?

oibe: it will always be a string when being deserialized, but its useful to have it take numbers for tests.  I can make it documented as only taking integers, but the truth is there are not types, so it can take numbers.

yurishkuro:
:shipit:

PR #3, opened 2016-08-17
Tracer and span

oibe:


src/constants.js
```diff
@@ -0,0 +1,22 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+export const SAMPLED_MASK = 0x1;
+export const DEBUG_MASK = 0x2;
```
yurishkuro: didn't you move these to spanContext?

src/_interfaces/reporter.js
```diff
@@ -0,0 +1,26 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+declare interface Reporter {
+    report(span: Span): void;
+    flush(callback:Function): void;
+    close(callback: Function): void;
```
yurishkuro: what's the difference between flush and close?

src/decls/types.js
```diff
@@ -0,0 +1,51 @@
...
+declare class Span {};
+
+declare class Trcer {};
+
+declare type spanFields = {
+    operationName: string,
+    timestamp?: number,
+    event: string,
+    payload: any,
+    peer: Endpoint
```
yurishkuro: what about debug flag?

src/decls/types.js
```diff
@@ -0,0 +1,51 @@
...
+declare type BinaryAnnotation = {
+    key: string,
+    value: any,
+    annotationType: string,
+    host?: Endpoint
+};
+
+declare class Span {};
+
+declare class Trcer {};
```
yurishkuro: misspelled

src/propagators/text_map_codec.js
```diff
@@ -0,0 +1,89 @@
...
+    _decodedValue(value: string): string {
+        if (this._urlEncoding) {
+            return decodeURIComponent(value);
+        }
+
+        return value;
+    }
+
+    extract(carrier: any): any {
+        let header = carrier[`${this._contextKey}`];
```
yurishkuro: what's with the escaping? why not `carrier[this._contextKey]`?

src/propagators/text_map_codec.js
```diff
@@ -0,0 +1,89 @@
...
+
+        let spanContext;
```
yurishkuro: don't need to pre-declare, since you're returning in the success branch, you don't need `else`

src/propagators/text_map_codec.js
yurishkuro: what happens if the key does not exist? we must avoid exceptions

src/propagators/text_map_codec.js
```diff
@@ -0,0 +1,89 @@
...
+        if (!header) {
+            return null;
+        } else {
+            spanContext = SpanContext.fromString(this._decodedValue(header));
+        }
+
+        for (let key in carrier) {
+            if (carrier.hasOwnProperty(key) && Utils.startsWith(key, this._baggagePrefix)) {
```
yurishkuro: why `carrier.hasOwnProperty(key)`? what if we didn't check?

src/propagators/text_map_codec.js
```diff
@@ -0,0 +1,89 @@
...
+                let keyWithoutPrefix = key.substring(this._baggagePrefix.length);
+                let value = carrier[key];
```
yurishkuro: pass through _decodeValue?

src/propagators/text_map_codec.js
```diff
@@ -0,0 +1,89 @@
...
+
+    inject(spanContext: SpanContext, carrier: any): any {
+        let stringSpanContext = spanContext.toString();
+        carrier[`${this._contextKey}`] = this._encodedValue(stringSpanContext);
+
+        let baggage = spanContext.baggage;
+        for (let key in baggage) {
+            if (baggage.hasOwnProperty(key)) {
+                let value = spanContext.getBaggageItem(key);
+                carrier[`${this._baggagePrefix}${key}`] = value;
```
yurishkuro: _encoded value?

src/propagators/text_map_codec.js
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    inject(spanContext: SpanContext, carrier: any): any {
+        let stringSpanContext = spanContext.toString();
+        carrier[`${this._contextKey}`] = this._encodedValue(stringSpanContext);
+
+        let baggage = spanContext.baggage;
+        for (let key in baggage) {
+            if (baggage.hasOwnProperty(key)) {
+                let value = spanContext.getBaggageItem(key);
```
yurishkuro: why go though spanContext again, if we already have baggage dictionary right here?

src/reporters/in_memory_reporter.js
```diff
@@ -0,0 +1,45 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+export default class InMemoryReporter {
```
yurishkuro: add clear() method, useful in tests sometimes

src/reporters/noop_reporter.js
```diff
@@ -0,0 +1,31 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+export default class NoopReporter {
```
yurishkuro: Q: isn't there a way in ES6 to indicate that this implements the Reporter interface?

src/reporters/in_memory_reporter.js
```diff
@@ -0,0 +1,45 @@
...
+    report(span: Span): void {
+        this._spans.push(span);
+    }
+
+    get spans(): Array<Span> {
+        return this._spans;
+    }
+
+    flush(callback: Function): void {
+    }
```
yurishkuro: questionable

src/samplers/noop_sampler.js
```diff
@@ -0,0 +1,33 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+export default class NoopSampler {
```
yurishkuro: not needed

src/span.js
```diff
@@ -0,0 +1,163 @@
...
+import Utils from './util.js';
+import SpanContext from './span_context.js';
+
+export default class Span {
+    _tracer: any;
+    _name: string;
+    _spanContext: SpanContext;
+    _start: number;
+    _duration: number;
+    _port: number;
```
yurishkuro: what port?

src/span.js
```diff
@@ -0,0 +1,163 @@
...
+    _isClient: boolean;
+    _peer: Endpoint;
+    _ended: boolean;
```
yurishkuro: why?

src/constants.js
oibe: Yes, but because I reference SAMPLED_MASK In tests, and in the future (after tags are added) DEBUG_MASK I felt they belong here afterall.

src/_interfaces/reporter.js
oibe: close calls "close" on reporter and sampler, but I question whether we should have a close at all.  The tracer is not really meant to be closed.

src/decls/types.js
oibe: That should probably be added here.  I guess debug is an implementation detail that should not exist in opentracing (i'm asking because it wasn't there either.).

src/decls/types.js
oibe: I do, but forgot to run flow check for a bit.  It does complain about this.

src/propagators/text_map_codec.js
oibe: No exception is thrown, it just returns undefined.

src/propagators/text_map_codec.js
oibe: Its best practice to use hasOwnProperty because this makes sure its actually on the object instead of the object's prototype property.  javascript :(

src/reporters/in_memory_reporter.js
oibe: questionable as in we should remove it?  Or that his flush should do more?

src/reporters/noop_reporter.js
oibe: Object and Interface types are structural. Classes implement interfaces, and satisfy object shapes, on a structural basis.  Also interfaces aren't really part of ES6, they are part of flow.  Flow is really a development tool that provides you some type checking, but once you compile it, or run the code in es6 mode it gets stripped out.  I'm starting to think we shouldn't use it.

src/_interfaces/reporter.js
yurishkuro: yes, we need close() on Tracer for clean shutdown during unit tests. But flush() is not intrinsic to all reporters, I don't think it belongs in the interface - tracer would never call it.

src/decls/types.js
yurishkuro: it does not exist in opentracing indeed, but this is a structure for our internal data model. Actually, I don't even know what this structure is for - what are "Event" and "payload" at the span level?

src/reporters/in_memory_reporter.js
yurishkuro: remove it

src/span.js
yurishkuro: this is where we need debug flag, perhaps? This struct gets converted to thrift, correct?

src/span.js
```diff
@@ -0,0 +1,163 @@
...
+                isClient: boolean = false,
+                peer: Endpoint,
+                annotations: Array<Annotation> = [],
+                binaryAnnotations: Array<BinaryAnnotation> = []
+    ) {
+        this._tracer = tracer;
+        this._name = name;
+        this._spanContext = spanContext;
+        this._start = start;
+        this._duration = 0;
```
yurishkuro: maybe undefined?

src/span.js
```diff
@@ -0,0 +1,163 @@
...
+        this._name = name;
+    }
+
+    /**
+     * Finishes a span which has the effect of reporting it and
+     * setting the finishTime on the span.
+     *
+     * @param {number} finishTime - The time on which this span finished.
+     **/
+    finish(finishTime: number): void {
```
yurishkuro: I take it args in JS are optional?

src/span.js
```diff
@@ -0,0 +1,163 @@
...
+        assert(!this._ended, "You can only call finish() on span once");
+        this._ended = true;
```
yurishkuro: can use `duration == undefined` instead

src/span.js
```diff
@@ -0,0 +1,163 @@
...
+
+        if (this._spanContext.IsSampled() && this._start) {
```
yurishkuro: this._start seems superfluous

src/span_context.js
```diff
@@ -125,14 +123,14 @@ export default class SpanContext {
      * @return {boolean} - returns whether or not this span context was sampled.
      **/
     IsSampled(): boolean {
```
yurishkuro: lowercase `is`, same for isdebug

src/tracer.js
```diff
@@ -0,0 +1,218 @@
...
+            spanContext,
+            Utils.getTimestampMicros(),
+            isClient,
+            peer
+        );
+
+        span.log({
+            'event': isClient ? thrift.CLIENT_SEND : thrift.SERVER_RECV,
+            'timestamp': span._timestamp
+        });
```
yurishkuro: I think we should move these to reporter. They are artifacts of Zipkin representation, we don't need them to be in the main tracer code.

src/span.js
```diff
@@ -0,0 +1,163 @@
...
+            let endTime = finishTime || Utils.getTimestampMicros();
+            this._duration = endTime - this._start;
+            this.log({
+                event: this._isClient ? thrift.CLIENT_RECV : thrift.SERVER_SEND,
+                timestamp: endTime
+            });
```
yurishkuro: I think we should move these to reporter. They are artifacts of Zipkin representation, we don't need them to be in the main tracer code.

yurishkuro: also, we do not always log these annotations, e.g. for local spans, so this is wrong anyway - compare with go/python

src/tracer.js
```diff
@@ -0,0 +1,218 @@
...
+    *        should not modify this object after calling startSpan).
+    * @param {number} [fields.startTime] - a manually specified start time for
+    *        the created Span object. The time should be specified in
+    *        milliseconds as Unix timestamp. Decimal value are supported
+    *        to represent time values with sub-millisecond accuracy.
+    * @return {Span} - a new Span object.
+    **/
+    startSpan(fields: spanFields): Span {
+        let randomId = Utils.getRandom64();
+        let isClient = fields.isClient || !!fields.childOf || false;
```
yurishkuro: fields.isClient is not a thing. ChildOf has no bearing on isClient. Compare with go logic

src/tracer.js
```diff
@@ -0,0 +1,218 @@
...
+
+        if (this._sampler.isSampled()) {
+            flags = constants.SAMPLED_MASK;
+        }
+
+        let spanContext;
+        if (fields.childOf) {
+            spanContext = new SpanContext(
+                fields.childOf.traceId,
+                randomId,
```
yurishkuro: this is not true for RPC_SERVER spans, compare with go

src/propagators/text_map_codec.js
oibe: The way i've rewritten it, its still useful to predeclare.

src/tracer.js
```diff
@@ -0,0 +1,218 @@
...
+    *        passed to the created span for efficiency reasons (the caller
+    *        should not modify this object after calling startSpan).
+    * @param {number} [fields.startTime] - a manually specified start time for
+    *        the created Span object. The time should be specified in
+    *        milliseconds as Unix timestamp. Decimal value are supported
+    *        to represent time values with sub-millisecond accuracy.
+    * @return {Span} - a new Span object.
+    **/
+    startSpan(fields: spanFields): Span {
+        let randomId = Utils.getRandom64();
```
yurishkuro: we do not need random ID for RPC_SERVER span, so don't do it upfront

src/tracer.js
```diff
@@ -0,0 +1,218 @@
...
+                randomId,
+                randomId,
+                null,
+                flags,
+                this._tracerTags
+            );
+        }
+
+        // todo(oibe) in the new model how will the instrumentation specify the peer for
+        // a client, and server span?
```
yurishkuro: as peer tags

src/propagators/text_map_codec.js
oibe: spanContext normalizes the key as well.

src/tracer.js
```diff
@@ -0,0 +1,218 @@
...
+            this._sampler.close(callback);
+        });
+    }
+
+    /**
+     * Writes spans to the local agent, and executes a callback if necessary.
+     *
+     * @param {Function} [callback] - a callback that runs after spans have been flushed.
+     **/
+    flush(callback: Function): void {
```
yurishkuro: not sure what the use case is for flush on the tracer

src/span.js
oibe: So we can check if we called finish on the span, but I suppose it should not be part of the constructor.

src/propagators/text_map_codec.js
yurishkuro: and if you remove else nesting, it will be 4 lines instead of 6, without hanging variable declarations

src/propagators/text_map_codec.js
yurishkuro: you are already getting key from baggage on the spanContext

test/tracer.js
```diff
@@ -0,0 +1,148 @@
...
+        tracer._report(span);
+
+        assert.isOk(reporter.spans.length, 1);
+        assert.equal(span, reporter.spans[0]);
+    });
+
+    it ('start a root span with proper structure', () => {
+        let span = tracer.startSpan({
+            operationName: 'test-name',
+            isClient: true
```
yurishkuro: there is no isClient in OT API

test/tracer.js
```diff
@@ -0,0 +1,148 @@
...
+        });
+
+        assert.equal(span.context().traceId, span.context().spanId);
+        assert.isNotOk(span.context().parentId);
+        assert.equal(span.context().flags, constants.SAMPLED_MASK);
+    });
+
+    it ('start a child span with proper structure', () => {
```
yurishkuro: there are three ways to create a child span, we need to test all
  * client span (new span ID)
  * rpc server span (same span ID)
  * local span (new span ID, different zipkin annotations)

test/span.js
```diff
@@ -0,0 +1,116 @@
...
+            timestamp: timestamp,
+            event: 'some message'
+        });
+
+        assert.equal(span._annotations.length, 1);
+        assert.equal(span._annotations[0].timestamp, timestamp);
+        assert.equal(span._annotations[0].value, 'some message');
+    });
+});
+
```
yurishkuro: need tests for standard tags (can be diff PR, but add todo)

src/decls/types.js
oibe: event - a string that gets logged as a LogData object associated with a  span
payload - an object that gets turned into a json sring and logged as LogData on a span

src/span.js
oibe: 0 and undefined are both falsy, but since duration's type is a number it makes more sense for it to be 0.

src/span.js
oibe: 0 and undefined are both falsy, but since duration's type is a number it makes more sense for it to be 0.

src/span.js
oibe: Yes.

src/span.js
oibe: why?

src/decls/types.js
yurishkuro: those are the properties of an individual log message. How are they related to `span fields`?

src/span.js
yurishkuro: I think you replied on the wrong place.

But re duration, undefined makes more sense to me because at this point in time you do not know what duration of the span is, hence it is undefined. 0 is not undefined, it's an actual value. In Go it would be natural to use 0, but here `undefined` expresses the meaning better.

src/span.js
yurishkuro: Undefined makes more sense to me because at this point in time you do not know what duration of the span is, hence it is `undefined`. 0 is not undefined, it's an actual value. In Go it would be natural to use 0 because there's no other choice, but here `undefined` expresses the meaning better.

src/span.js
yurishkuro: because it makes the extra field `_ended` unnecessary, it does not carry any additional information.

src/tracer.js
oibe: I agree, but opentracing wants the ability to explicitly call flush.

yurishkuro: ah, interesting. That's not present in OT for any other languages.

Fine, then we need flush on the top-level reporter interface.

But we still need close, which is slightly different - it assumes flush as well, but also does clean-up like timers etc.

src/_interfaces/injector.js
```diff
@@ -0,0 +1,26 @@
+// @flow
```
yurishkuro: are these just pre-compile files for flow? Can we not combine all interfaces into one file?

src/decls/types.js
```diff
@@ -0,0 +1,54 @@
...
+
+declare type BinaryAnnotation = {
+    key: string,
+    value: any,
+    annotationType: string,
+    host: ?Endpoint
+};
+
+declare type Reference = {
+    type(): void;
```
yurishkuro: why void?

src/decls/types.js
```diff
@@ -0,0 +1,54 @@
...
+    referencedContext(): SpanContext;
+};
+
+declare type spanFields = {
```
yurishkuro: I still don't understand why this declaration has such an odd collection of field names. If it's supposed to represent the whole Span structure, then there are a lot more fields, like `logs` or `debug`, while at the same time `childOf` is **not** a field on the Span. And there is no equivalent declaration for the Tracer. What exactly is the purpose here?

src/span.js
```diff
@@ -0,0 +1,165 @@
...
+                binaryAnnotations: Array<BinaryAnnotation> = []
+    ) {
+        this._tracer = tracer;
+        this._name = name;
+        this._spanContext = spanContext;
+        this._start = start;
+        this._firstInProcess = firstInProcess;
+        this._duration;
+        this._isClient;
+        this._peer;
```
yurishkuro: what do these 3 lines do?

src/span.js
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    /**
+     * Finishes a span which has the effect of reporting it and
+     * setting the finishTime on the span.
+     *
+     * @param {number} finishTime - The time on which this span finished.
+     **/
+    finish(finishTime: number): void {
+        if (this._duration) {
```
yurishkuro: this should be something like `this._duration === undefined` or whatever the syntax is. Afaik we're only getting millisecond precision timestamps in JS, so it's quite possible for a duration of a finished span to be 0 if it was faster that 1ms.

src/util.js
```diff
@@ -65,4 +97,134 @@ export default class Utils {
...
+        return this.createBinaryAnnotation(key, value, thrift.annotationType.STRING);
+    }
+
+    /**
+     * Returns the timestamp in microseconds.
+     * @return {number} - The microseconds since the epoch.
+     **/
+    static getTimestampMicros(): number {
+        return Date.now() * 1000;
+    }
```
yurishkuro: please add a todo to investigate `process.hrtime()`, per http://stackoverflow.com/questions/11725691/how-to-get-a-microtime-in-node-js

src/span_context.js
```diff
@@ -79,8 +94,8 @@ export default class SpanContext {
      * @param {string} key - The key to be normalized for a particular baggage value.
      * @return {string} - The normalized key (lower cased and underscores replaced, along with dashes.)
      **/
-    // todo(oibe) moved to span in newest opentracing version
     _normalizeBaggageKey(key: string) {
+       // todo(oibe) should be moved to span in newest opentracing version
```
yurishkuro: todo's are customarily written in ALL CAPS. The point of TODO is to stand-out.

src/span.js
```diff
@@ -0,0 +1,165 @@
...
+    _tracer: any;
+    _name: string;
+    _spanContext: SpanContext;
+    _start: number;
+    _duration: number;
+    _firstInProcess: boolean;
+    _isClient: boolean;
+    _peer: Endpoint;
+    _annotations: Array<Annotation>;
+    _binaryAnnotations: Array<BinaryAnnotation>;
```
yurishkuro: given these declarations, what does spanFields in the flow file do?

src/tracer.js
```diff
@@ -0,0 +1,258 @@
...
+    _sampler: Sampler;
+    _host: Endpoint;
+    _tracerTags: any;
+    _injectors: any;
+    _extractors: any;
+
+    constructor(serviceName: string,
+            reporter: Reporter = new NoopReporter(),
+            sampler: Sampler = new ConstSampler(false),
+            port: number  = 0,
```
yurishkuro: port doesn't make sense. A single service may listen on multiple ports.

src/span.js
```diff
@@ -0,0 +1,165 @@
...
+            }
+            let value = fields.event || JSON.stringify(fields.payload);
+            let logData = Utils.createLogData(fields.timestamp, value);
+
+            logData.host = this._tracer._host;
+            this._annotations.push(logData);
+        }
+    }
+
+    setSamplingPriority(priority: number): void {
```
yurishkuro: should this be `_setSampingPriority`, for consistency with private fields?

src/tracer.js
```diff
@@ -0,0 +1,258 @@
...
+
+    registerExtractor(format: string, extractor: Extractor): void {
+        this._extractors[format] = extractor;
+    }
+
+    /**
+    * The method for creating a root or child span.
+    *
+    * @param {object} fields - the fields to set on the newly created span.
+    * @param {string} fields.operationName - the name to use for the newly
```
yurishkuro: comments below enclose attributes in square brackets, but this one is not, any reason?

src/tracer.js
```diff
@@ -0,0 +1,258 @@
...
+        let references = fields.references || [];
+        let startTime = fields.startTime;
+        if (!startTime) {
+            startTime = Utils.getTimestampMicros();
+        }
+
+        let parent: ?SpanContext;
+        for (let i = 0; i < references.length; i++) {
+            let ref: Reference = references[i];
+            if (ref.type() === opentracing.REFERENCE_CHILD_OF) {
```
yurishkuro: follows_from also defines the parent, as far as parentID goes

src/tracer.js
```diff
@@ -0,0 +1,258 @@
...
+            } else {
+                // todo(oibe) support other types of span references
+            }
+        }
+
+        let rpcServer = false;
+        let spanKindValue = tags[opentracing_tags.SPAN_KIND];
+        if (spanKindValue) {
+            rpcServer = (spanKindValue === opentracing_tags.SPAN_KIND_RPC_SERVER);
+        }
```
yurishkuro: isn't it equivalent to these 2 lines?
```
let spanKindValue = tags[opentracing_tags.SPAN_KIND];
let rpcServer = (spanKindValue === opentracing_tags.SPAN_KIND_RPC_SERVER);
```

src/tracer.js
```diff
@@ -0,0 +1,258 @@
...
+            ctx.flags = flags;
+        } else {
+            ctx.traceId = parent.traceId;
+            if (rpcServer) {
+                // Support Zipkin's one-span-per-RPC model
+                ctx.spanId = parent.spanId;
+                ctx.parentId = parent.parentId;
+            } else {
+                let randomId = Utils.getRandom64();
+                ctx.spanId = randomId;
```
yurishkuro: inline the call to random ID, variable is not reused

src/tracer.js
```diff
@@ -0,0 +1,258 @@
...
+                ctx.parentId = parent.spanId;
+            }
+            ctx.flags = parent.flags;
+
+            // copy baggage items
+            for (let key in parent.baggage) {
+                if (parent.baggage.hasOwnProperty(key)) {
+                    let value = parent.baggage[key];
+                    ctx.setBaggageItem(key, value);
```
yurishkuro: if this was python, all we really need is `ctx.baggage = dict(parent.baggage)`. I don't know if JS has a dict copy, but we certainly don't need to go through setBaggageItem method, because we already have a normalized dictionary from parent.

Once we (do)/land the PR for "immutable" SpanContext, this can simply become `ctx.baggage = parent.baggage`, because span.setBaggage() should be "copy on write", which happens a lot less often than creating new spans from parent spans.

yurishkuro: NB: why don't you do a PR for immutable SpanContext?  We're not going to release this Tracer version with setBaggage on SpanContext anyway, so it's a blocker (and Rooz can't use it like this).

test/span.js
```diff
@@ -0,0 +1,114 @@
...
+    it('return tracer when tracer() is called', () => {
+        assert.equal(span.tracer(), tracer);
+    });
+
+    it('set operation name correctly', () => {
+        span.setOperationName('operation-name');
+        assert.equal(span._name, 'operation-name');
+    });
+
+    it('finish reports span with proper log', () => {
```
yurishkuro: finish span with custom duration

test/tracer.js
```diff
@@ -0,0 +1,183 @@
...
+        let spanId = Utils.encodeInt64(2);
+        let parentId = Utils.encodeInt64(3);
+        let flags = 1;
+        let context = new SpanContext(traceId, spanId, parentId, flags);
+        let startTime = Utils.getTimestampMicros();
+
+        let span = tracer.startSpan({
+            operationName: 'test-name',
+            childOf: context,
+            references: [new opentracing.Reference(opentracing.REFERENCE_CHILD_OF, context)],
```
yurishkuro: reference is redundant in the presence of `childOf`. Should really be two tests, one with `childOf`, another the same but via reference.

test/tracer.js
```diff
@@ -0,0 +1,183 @@
...
+        let flags = 1;
+        let context = new SpanContext(traceId, spanId, parentId, flags);
+        let startTime = Utils.getTimestampMicros();
+
+        let tags = {};
+        tags[`${opentracing_tags.SPAN_KIND}`] = opentracing_tags.SPAN_KIND_RPC_SERVER;
+        let span = tracer.startSpan({
+            operationName: 'test-name',
+            childOf: context,
+            references: [new opentracing.Reference(opentracing.REFERENCE_CHILD_OF, context)],
```
yurishkuro: redundant, same as above

test/tracer.js
```diff
@@ -0,0 +1,183 @@
...
+            assert.isOk(bufferEqual(savedContext.traceId, extractedContext.traceId));
+            assert.isOk(bufferEqual(savedContext.spanId, extractedContext.spanId));
+            assert.isOk(bufferEqual(savedContext.parentId, extractedContext.parentId));
+            assert.equal(savedContext.flags, extractedContext.flags);
+            assert.equal(savedContext.getBaggageItem(keyOne), extractedContext.getBaggageItem(keyOne));
+            assert.equal(savedContext.getBaggageItem(keyTwo), extractedContext.getBaggageItem(keyTwo));
+        }
+
+        assertByFormat(opentracing.FORMAT_TEXT_MAP);
+        assertByFormat(opentracing.FORMAT_HTTP_HEADERS);
```
yurishkuro: http headers codec must be url-encoding values, so the test should be slightly different (or have an extra one)

src/_interfaces/injector.js
oibe: This tells flow to type check a particular file.  Flow doesn't look at files that don't have "// @flow" We could move all declared interfaces to a single file, but this seems more organized, and in line with what other projects i've looked at do.

src/decls/types.js
oibe: Its not the span structure.  I think the naming is bad, but its really the parameters to start Span, since opentracing's start span method called their parameters fieldsOrName (but they guarantee you the fields)  I used fields as the name, so then I called this type spanFields.  Maybe "startSpanFields"?

src/span.js
oibe: see above.

src/span.js
oibe: They are not needed.  Since declared above.

src/tracer.js
oibe: Yet an endpoint object takes a single port, so Should we allow a list of ports?

src/tracer.js
oibe: square brackets in jsdoc mean optional, but a span must always have an operationName, so its not in square brackets.

src/tracer.js
oibe: jaeger-go only references child_of, should it also be changed to accept follows_from?

src/_interfaces/injector.js
yurishkuro: just annoying to have 20 lines of license for 3 lines of code. but fine.

src/decls/types.js
yurishkuro: startSpanArgs is fairly descriptive (Fields not so much)

src/tracer.js
yurishkuro: let's accept both

src/span.js
```diff
@@ -0,0 +1,220 @@
...
+        this._annotations = annotations;
+
+        // tags
+        this._binaryAnnotations = binaryAnnotations;
+    }
+
+    static _getBaggageHeaderCache() {
+        if (!this._baggageHeaderCache) {
+            // $FlowIgnore - flow doesn't seem to like how statics are declared
+            this._baggageHeaderCache = {};
```
yurishkuro: I don't see what makes this static - it looks just like any other member variable on `this`

yurishkuro: http://babeljs.io/docs/plugins/transform-class-properties/ shows example of static variables

src/span.js
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    /**
+     * Returns a normalize key.
+     *
+     * @param {string} key - The key to be normalized for a particular baggage value.
+     * @return {string} - The normalized key (lower cased and underscores replaced, along with dashes.)
+     **/
+    _normalizeBaggageKey(key: string) {
+       // TODO(oibe) should be moved to span in newest opentracing version
```
yurishkuro: rm

src/span.js
```diff
@@ -0,0 +1,220 @@
...
+        let baggageHeaderCache = this.constructor._getBaggageHeaderCache();
+        if (key in baggageHeaderCache) {
+            return baggageHeaderCache[key];
+        }
+
+        // This stackoverflow claims this approach is slower than regex
+        // http://stackoverflow.com/questions/1144783/replacing-all-occurrences-of-a-string-in-javascript
```
yurishkuro: won't surprise me, you're creating an array of strings and then joining them into a new string. Just use regex. The whole point of your baggage header cache is that this is only done a few times

src/span.js
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    /**
+     * Sets a baggage value with an associated key.
+     *
+     * @param {string} key - The baggage key.
+     * @param {string} value - The baggage value.
+     **/
+    setBaggageItem(key: string, value: string): void {
+        // TODO(oibe) should be moved to span in newest opentracing version
```
yurishkuro: rm

src/span.js
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    /**
+     * Gets a baggage value with an associated key.
+     *
+     * @param {string} key - The baggage key.
+     * @return {string} value - The baggage value.
+     **/
+    getBaggageItem(key: string): string {
+        // TODO(oibe) should be moved to span in newest opentracing version
```
yurishkuro: rm

src/span.js
```diff
@@ -0,0 +1,220 @@
...
+    log(fields: any): void {
+        if (this._spanContext.isSampled()) {
+            if (!fields.timestamp) {
+                fields.timestamp = Utils.getTimestampMicros();
+            }
+
+            if (!fields.event && !fields.payload) {
+                throw new Error('log must be passed either an event of type string, or a payload of type object');
+            }
+            let value = fields.event || JSON.stringify(fields.payload);
```
yurishkuro: I don't think payload should be used as the message in the log. I recently made a change in Go client to store payloads are `log_payload_#` tags: https://github.com/uber/jaeger-client-go/pull/31

src/span_context.js
```diff
@@ -149,6 +111,13 @@ export default class SpanContext {
         ].join(':');
     }
 
+    withBaggageItem(key: string, value: string): SpanContext {
+        let newBaggage = _.assign(this._baggage);
```
yurishkuro: according to [this,](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) `assign()`'s first argument is the target of a copy.

src/span_context.js
```diff
@@ -159,7 +128,7 @@ export default class SpanContext {
             return null;
         }
 
-        // todo(oibe) is it better to throw an exception here, or return null?
+        // TODO(oibe) is it better to throw an exception here, or return null?
```
yurishkuro: we must always avoid exceptions, not only because they bubble up and can crash the app, but also because they confuse Node's optimizer

src/tags.js
```diff
@@ -0,0 +1,65 @@
+module.exports = {
```
yurishkuro: TODO to replace with official OT constants, I presume?

src/tracer.js
```diff
@@ -0,0 +1,255 @@
...
+                ctx.spanId = parent.spanId;
+                ctx.parentId = parent.parentId;
+            } else {
+                ctx.spanId = Utils.getRandom64();
+                ctx.parentId = parent.spanId;
+            }
+            ctx.flags = parent.flags;
+
+            // copy baggage items
+            ctx.baggage = _.assign(parent.baggage);
```
yurishkuro: there is no need to copy, baggage is immutable

src/util.js
```diff
@@ -45,6 +59,24 @@ export default class Utils {
...
 
+
+    /**
+     * @param {string} ip - a string representation of an ip address.
+     * @return {number} - a 32-bit number where each byte represents an
+     * octect of an ip address.
+     **/
+    static ipToInt(ip: string): number {
+        let ipl = 0;
+        let parts = ip.split('.');
```
yurishkuro: check for size==4, otherwise it can return nonsense number

src/util.js
```diff
@@ -45,6 +59,24 @@ export default class Utils {
...
+        for (let i = 0; i < parts.length; i++) {
+            ipl <<= 8;
+            ipl += parseInt(parts[i], 10);
+        }
+        // TODO(oibe) prove that this will never be grater than 2^31 because
```
yurishkuro: 169.x.x.x will be greater that 2^31

you just need to take a complement in this case, 2^32 - ipl

yurishkuro: add unit test

test/span.js
```diff
@@ -0,0 +1,133 @@
...
+        assert.equal(keys[0], 'numberTag');
+        assert.equal(keys[1], 'stringTag');
+        assert.equal(keys[2], 'booleanTag');
+    });
+
+    it('add logs as annotations', () => {
+        let timestamp = new Date(2016, 8, 12).getTime();
+        span.log({
+            timestamp: timestamp,
+            event: 'some message'
```
yurishkuro: more use cases:
  * no timestamp
  * event and payload
  * payload only

test/tracer.js
```diff
@@ -0,0 +1,230 @@
...
+            Utils.encodeInt64(2),
+            Utils.encodeInt64(3),
+            constants.SAMPLED_MASK,
+            baggage
+        );
+        let carrier = {};
+
+        tracer.inject(savedContext, opentracing.FORMAT_HTTP_HEADERS, carrier);
+        assert.equal(carrier['UberCtx-keyOne'], encodeURIComponent('leela'));
+        assert.equal(carrier['UberCtx-keyTwo'], encodeURIComponent('bender'));
```
yurishkuro: this validation is moot, because you picked the values that do not change when passed through url encoding.

Try values with spaces like `"Leela vs. Bender"` and verify that spaces will be represented as `%20`. Compare with literal expected string, not by delegating to encodeURI method again.

test/tracer.js
```diff
@@ -0,0 +1,230 @@
...
+            Utils.encodeInt64(2),
+            Utils.encodeInt64(3),
+            constants.SAMPLED_MASK
+        );
+
+        // subtle but expect wants a function to call not the result of a function call.
+        expect(() => {tracer.inject(context, 'fake-format', carrier)}).to.throw('Unsupported format: fake-format');
+        expect(() => {tracer.extract('fake-format', carrier)}).to.throw('Unsupported format: fake-format');
+    });
+});
```
yurishkuro: missing tests: what happens when malformed trace context string is given to extractor? https://github.com/uber/jaeger-client-go/blob/master/context_test.go#L10

todo.txt
```diff
@@ -0,0 +1,5 @@
++ support different types of spans: client span (new span ID), rpc server span (same span ID) , (component, and lc) local span (new span ID, different zipkin annotations)
```
yurishkuro: profile performance of Noop tracer vs. Jaeger tracer with const sampling off and on

src/tracer.js
oibe: technically if someone wants they can always modify fields.  No real immutable in javascript.

yurishkuro: we don't care if someone modifies the fields, it's their fault. The design is that the SpanContext contains a snapshot of baggage to which the tracer will never add anything, so we do not need to make copies of that map when creating children spans, we can point to the same map.

src/decls/types.js
```diff
@@ -21,8 +21,14 @@
...
 
+declare type LogData = {
+    timestamp: ?number,
+    event: ?string,
+    payload: ?any
+};
+
 declare type Endpoint = {
-    ipv4: number,
+    ipv4: ?number,
```
yurishkuro: it's not optional in thrift. We should always set it to 0.

src/span.js
```diff
@@ -19,52 +19,49 @@
...
+    static _baggageHeaderCache: any;
 
     constructor(tracer: any,
                 name: string,
                 spanContext: SpanContext,
                 start: number = Utils.getTimestampMicros(),
-                firstInProcess: boolean = false,
-                annotations: Array<Annotation> = [],
-                binaryAnnotations: Array<BinaryAnnotation> = []
+                logger: TracerLogger = new TracerLogger(),
```
yurishkuro: not a good idea to instantiate logger for every span

src/tracer.js
```diff
@@ -180,7 +183,7 @@ export default class Tracer {
             ctx.flags = parent.flags;
 
             // copy baggage items
```
yurishkuro: `// reuse parent's baggage as we'll never change it`

test/span.js
```diff
@@ -80,43 +81,52 @@ describe('span should', () => {
...
 
-    it('add tags as binary annotations', () => {
+    it('add tags', () => {
         span.addTags({
             numberTag: 7,
             stringTag: 'string',
             booleanTag: true,
         });
 
+        assert.equal(Object.keys(span._tags).length, 3);
```
yurishkuro: * nit: can we compare with another (expected) dictionary?
 * need a test for calling addTags twice, I think you're overriding the previous values.

src/span.js
```diff
@@ -19,52 +19,49 @@
...
+    _logger: TracerLogger;
     _duration: number;
     _firstInProcess: boolean;
     _isClient: boolean;
     _peer: Endpoint;
-    _annotations: Array<Annotation>;
-    _binaryAnnotations: Array<BinaryAnnotation>;
-    _baggageHeaderCache: any;
+    _logs: Array<LogData>;
+    _tags: any;
```
yurishkuro: I would really like to see some profiling results. iirc the reason we had the old structure here was to avoid copying the data into another set of objects when generating thrift message.  This new format will work better once we upgrade to a new data model (which we *could* do as part of UDP reporter).

src/tracer.js
```diff
@@ -31,13 +30,14 @@ import pjson from '../package.json';
 import Span from './span.js';
 import SpanContext from './span_context.js';
 import TextMapCodec from './propagators/text_map_codec.js';
-import * as thrift from './thrift.js';
+import TracerLogger from './tracer_logger.js';
```
yurishkuro: why not just `./logger.js`?

src/tracer_logger.js
```diff
@@ -0,0 +1,46 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+export default class TracerLogger {
```
yurishkuro: I would rather call it NullLogger, without any parameters, and instantiate it if real logger was not passed in.

src/span.js
yurishkuro: should just be taking it from tracer

src/tracer.js
```diff
@@ -46,13 +46,15 @@ export default class Tracer {
...
             sampler: Sampler = new ConstSampler(false),
+            logger: any,
             tracerTags: any = {}) {
         let port = 0;
         tracerTags.jaegerClient = `Node-${pjson.version}`;
         this._tracerTags = tracerTags
         this._serviceName = serviceName;
         this._reporter = reporter;
         this._sampler = sampler;
+        this._logger = new TracerLogger(logger);
```
yurishkuro: `logger || new NullLogger()`

src/span.js
yurishkuro: actually, why do we need ay defaults here? The span is only supposed to be instantiated by tracer

src/tracer.js
```diff
@@ -46,13 +46,15 @@ export default class Tracer {
     constructor(serviceName: string,
             reporter: Reporter = new NoopReporter(),
             sampler: Sampler = new ConstSampler(false),
```
yurishkuro: do defaults work if you pass null as an argument?

kriskowal:
Please do not Babel a Node.js service. That’s a lot of dependencies and workflow. It will confuse flame graphs and end in tears. Our existing Node.js backend services are all written for ES5.

src/tracer.js
oibe: They do not.

src/logger.js
```diff
@@ -19,28 +19,6 @@
...
-            this._logger.warn(msg);
-        }
-    }
-
-    error(msg: string): void {
-        if (this._logger) {
-            this._logger.error(msg);
-        }
-    }
+export default class NullLogger {
```
yurishkuro: I think you need to define noop methods that will be called, otherwise it will blow up

test/span.js
```diff
@@ -81,14 +81,43 @@ describe('span should', () => {
...
+
+    it('set debug and sampling version through sampling priority', () => {
+        span._setSamplingPriority(3);
+
+        assert.isOk(span.context().isDebug());
+        assert.isOk(span.context().isSampled());
+    });
+
+    it('unset sampling on span', () => {
+        span._setSamplingPriority(-1);
```
yurishkuro: hmm, the spec says nothing about negative numbers, only about 0

test/span.js
```diff
@@ -81,14 +81,43 @@ describe('span should', () => {
...
             stringTag: 'string',
             booleanTag: true,
-        });
-
-        assert.equal(Object.keys(span._tags).length, 3);
+        };
+        span.addTags(keyValuePairs);
+        span.addTags({numberTag: 8});
+
+        // test to make sure cosecutive calls with same key does not
```
yurishkuro: typo: consecutive

test/span.js
```diff
@@ -81,14 +81,43 @@ describe('span should', () => {
...
+        // overwrite the first key.
```
yurishkuro: that's not actually a requirement, but ok

test/span_context.js
```diff
@@ -91,4 +91,23 @@ describe ('SpanContext should', () => {
         assert.isOk(bufferEqual(Utils.encodeInt64(0x5), context.parentId));
         assert.equal(context.flags, 0x1);
     });
+
+    it ('return null on malformed traces', () => {
+        // TODO(oibe) do we need these tests?  We are the ones creating the serialized
+        // version of these fields.  They should never be corrupt.  We follow this practice
+        // in a lot of places.  Also it makes the code slower.
+        /*
```
yurishkuro: yes, we do. Headers can be corrupted by http routing layers. We need to make sure that we do not parse malformed IDs, and that we don't throw parsing exceptions.

yurishkuro:
aside from NullLogger() issue, and a couple nits, LGTM

PR #4, opened 2016-08-25
add samplers

oibe:


src/samplers/remote_sampler.js
```diff
@@ -0,0 +1,126 @@
...
+        this._sampler = options.sampler || new ProbabilisticSampler(DEFAULT_INITIAL_SAMPLING_RATE);
+        this._logger = options.logger || new NullLogger();
+        this._refreshInterval = options.refreshInterval || SAMPLING_REFRESH_INTERVAL;
+
+        let randomDelay: number = Math.random() * this._refreshInterval;
+        this._testCallback = options.testCallback;
+
+        // This branch is necessary because code in 'setTimeout' does not always run in the same tick
+        // thus in testing the code proceeds without calling 'options.callback', and the tests end without
+        // hitting all the proper assertions.
```
yurishkuro: not sure I follow what this comment is about. The actual code is about randomizing the poll moments across service instances, not about testing.

src/samplers/remote_sampler.js
```diff
@@ -0,0 +1,126 @@
...
+        if (options.firstRefreshDelay) {
```
yurishkuro: if it's not set in option, shouldn't this be the default behavior anyway? Unless this breaks the API with the old client, I would call this `disableFirstRefreshDelay`

src/samplers/remote_sampler.js
```diff
@@ -0,0 +1,126 @@
...
+            this._timeoutHandle = setTimeout(() => {
+                this._intervalHandle = setInterval(this._refreshSamplingStrategy, this._refreshInterval);
+            }, randomDelay)
+
+        } else {
+            this._refreshSamplingStrategy();
```
yurishkuro: shouldn't this also `setInterval`?

src/samplers/remote_sampler.js
```diff
@@ -0,0 +1,126 @@
...
+        }
+    }
+
+    _refreshSamplingStrategy() {
+        this._getSamplingStrategy(this._callerName);
+    }
+
+    _getSamplingStrategy(callerName: string): ?SamplingStrategyResponse  {
+        let encodedCaller: string = encodeURIComponent(callerName);
+        request.get(`http:\/\/127.0.0.1:${DEFAULT_SAMPLING_PORT}/?service=${encodedCaller}`, (err, response) => {
```
yurishkuro: no need to hardcode the host address, let's make it configurable and defaulted to loopback

src/samplers/remote_sampler.js
```diff
@@ -0,0 +1,126 @@
...
+                error: strategy
+            });
+        }
+
+        if (newSampler && (!this._sampler.equal(newSampler))) {
+            this._sampler = newSampler;
+        }
+
+        if (this._testCallback) {
+            this._testCallback(this._sampler);
```
yurishkuro: we can rename this to `onSamplerUpdate`

test/lib/sampler_server.js
```diff
@@ -0,0 +1,67 @@
...
+
+export default class SamplingServer {
+    constructor(port=5778) {
+        this._port = port
+        this._app = express();
+        this._setupResponses();
+    }
+
+    _setupResponses(){
+        this._app.get('/', function (req, res) {
```
yurishkuro: did we really attach the sampling manager endpoint to the root `/`?  My bad.

test/ratelimiting_sampler_test.js
```diff
@@ -0,0 +1,43 @@
...
+import {assert} from 'chai';
+import ProbabilisticSampler from '../src/samplers/probabilistic_sampler.js';
+import RateLimiter from '../src/samplers/ratelimiting_sampler.js';
+import sinon from 'sinon';
+
+describe ('leaky bucket ratelimiter should', () => {
+    it('block after threshold is met', () => {
+        let initialDate = new Date(2011,9,1).getTime();
+        let clock = sinon.useFakeTimers(initialDate);
+        let limiter = new RateLimiter(10);
```
yurishkuro: did you mean to test rate limiter or rate limiting sampler?

test/ratelimiting_sampler_test.js
```diff
@@ -0,0 +1,43 @@
...
+        for (let i = 0; i < 10; i++) {
+            limiter.isSampled(1);
+        }
+
+        assert.equal(limiter.maxTracesPerSecond, 10);
+        assert.isNotOk(limiter.equal(new ProbabilisticSampler(0.5)));
+        assert.equal(limiter.isSampled(1), false, 'expected checkCredit to be false');
+        clock = sinon.useFakeTimers(initialDate + 1000);
+        assert.equal(limiter.isSampled(1), true, 'expected checkCredit to be true');
```
yurishkuro: for rate limiter itself, we should also test how it behaves with frequency <1

test/remote_sampler.js
```diff
@@ -0,0 +1,65 @@
...
+        });
+    });
+
+    it('throw error on bad sampling strategy', () => {
+        let sampler = new RemoteSampler('error-service', {
+            firstRefreshDelay: false,
+            logger: new NullLogger()
+        });
+        // TODO(oibe) complete the assertion here by logging a metric
+    });
```
yurishkuro: looks like unfinished test

src/samplers/remote_sampler.js
oibe: It would break the api with the old client.

src/samplers/remote_sampler.js
oibe: The point of this branch is to not set Interval, but call refresh immediately for testing.

src/samplers/remote_sampler.js
yurishkuro: That's actually not a strong argument, because we do not have a need to maintain the same initialization API.IN order to use the new client the apps would need to change their initialization anyway, even if they want to use the legacy API.

src/samplers/remote_sampler.js
yurishkuro: why couldn't the test just pass a very short refresh interval like 1ms? It's not a clean design to build code branches only exec-ed in testing.

test/lib/sampler_server.js
oibe: yep.

test/ratelimiting_sampler_test.js
oibe: The leaky_bucket_rate_limiter is tested for this, but really "isSampled" should not take any values.  The parameter passed to initialize a rate limiter should always reflect the max traces per second assuming a credit of 1.0 is used.

test/ratelimiting_sampler_test.js
oibe: I cleaned this up to make it clearer.

test/ratelimiting_sampler_test.js
yurishkuro: that wasn't my point, I am saying that the rate limiting should also work for things like `maxTracesPerSecond=0.25`, i.e. one trace in 4 seconds.

src/samplers/remote_sampler.js
oibe: Short refresh intervals don't work in node js.  Even if the refresh interval is very short functions like setInterval, and setTimeout don't run in the same "tick" so the code will always proceed, and execute them later.  This makes testing difficult, and design more obnoxious.

yurishkuro: you're already passing callback invoked on the refresh, so it makes no difference how long JS takes until it runs the refresh, the only impact is how long your test will be blocked waiting for the callback (but it still going to be millis).

yurishkuro: I the remote sampler test I made this change:
```
-            firstRefreshDelay: false,
+            refreshInterval: 0.001,
```
and the tests pass. So we don't need it after all.

oibe: In the above it should be "firstRefreshDelay: true, refreshInterval: 0.001," that runs, but then you encounter a problem where the testCallback calls "done" multiple times.

src/samplers/remote_sampler.js
```diff
@@ -0,0 +1,126 @@
...
+        this._logger = options.logger || new NullLogger();
+        this._refreshInterval = options.refreshInterval || SAMPLING_REFRESH_INTERVAL;
+
+        let randomDelay: number = Math.random() * this._refreshInterval;
+        this._onSamplerUpdate = options.onSamplerUpdate;
+        this._host = options.host || DEFAULT_SAMPLING_HOST;
+        this._port = options.port || DEFAULT_SAMPLING_PORT;
+
+        // setTimeout and setInterval have delays, so for testing purposes I allow an alternative path
+        // that immediately executes the _refreshSamplingStrategy
```
yurishkuro: why can we not keep the production code free of "for testing only" code? The tests can always be written like this (just needs to be converted into callback format):
```
rs = RemoteSampler(delay=1ms)
for i in range(1, 1000) {
  if rs.sampler has changed {
    break;
  }
  sleep(1ms)
}
assert rs.sampler has changed
```
under normal conditions this test will complete within a few ms.

Right now we're messing up public API of the sampler with a flag that's only used for testing.

PR #6, opened 2016-08-25
remove my-local-ip dependency in favor of std utilities.

oibe:


src/tracer.js
```diff
@@ -54,7 +53,7 @@ export default class Tracer {
         this._reporter = reporter;
         this._sampler = sampler;
         this._logger = logger || new NullLogger();
-        this._host = Utils.createEndpoint(serviceName, myLocalIp(), port);
+        this._host = Utils.createEndpoint(serviceName, Utils.myIp(), port);
```
yurishkuro: if we are switching to the new data model, there's no need to e `endpoint` and for `port`. Simply add IP as a tracer tag.

yurishkuro:
aside from using the endpoint, LGTM

PR #8, opened 2016-09-08
adapt code to use opentracing interface

oibe:
Works with locally tested opentracing, but will need a new release of opentracing from https://github.com/opentracing/opentracing-javascript/pull/57.

### DIFF
--- a/src/span.js
+++ b/src/span.js
@@ -145,7 +145,6 @@ export default class Span {
      **/
     finish(finishTime: number): void {
         if (this._duration !== undefined) {
-        console.log('coolio');
             throw new Error('You can only call finish() on a span once.');
         }
 
@@ -192,11 +191,11 @@ export default class Span {
      * Adds a log event, or payload to a span.
      *
      * @param {Object} fields - an object that represents the fields to log.
-     * @param {number} [fields.timestamp] - the starting timestamp of a span.
+     * @param {number} [timestamp] - the starting timestamp of a span.
      * @param {string} [fields.event] - a string message to be logged on a span.
      * @param {string} [fields.payload] - an object to be logged on a span as a json string.
      **/
-    log(fields: any): void {
+    log(fields: any, timestamp: ?number): void {
         if (this._spanContext.isSampled()) {
             if (!fields.event && !fields.payload) {
                 this._logger.error('log must be passed either an event of type string, or a payload of type object');
@@ -204,7 +203,7 @@ export default class Span {
             }
 
             if (!fields.timestamp) {
-                fields.timestamp = Utils.getTimestampMicros();
+                fields.timestamp = timestamp || Utils.getTimestampMicros();
             }
 
             let value = fields.event;
@@ -242,6 +241,8 @@ export default class Span {
         let tags = [];
         for (let i = 0; i < this._tags.length; i++) {
             let tag = this._tags[i];
+            let emptyBuffer = new Buffer(1);
+            emptyBuffer.fill(0);
             tags.push({
                 key: tag.key,
                 tagType: 'STRING', // TODO(oibe) support multiple tag types
@@ -251,7 +252,7 @@ export default class Span {
                 vInt16: 0,
                 vInt32: 0,
                 vInt64: 0,
-                vBinary: new Buffer(1)
+                vBinary: emptyBuffer
             });
         }
 
--- a/src/span.js
+++ b/src/span.js
@@ -195,6 +195,24 @@ export default class Span {
         }
     }
 
+    /**
+     * Logs a event with an optional payload.
+     *
+     * @param  {string} eventName - string associated with the log record
+     * @param  {object} [payload] - arbitrary payload object associated with the
+     *         log record.
+     */
+    logEvent(eventName: string, payload: any): void {
+        return this.log({
+            event: eventName,
+            payload: payload
+        });
+    }
+
+    setTag(key: string, value: string): Span {
+        // TODO(oibe) implement
+    }
+
     _setSamplingPriority(priority: number): void {
         if (priority > 0) {
             this._spanContext.flags = this._spanContext.flags | constants.SAMPLED_MASK | constants.DEBUG_MASK;
--- a/src/span.js
+++ b/src/span.js
@@ -166,6 +166,7 @@ export default class Span {
     addTags(keyValuePairs: any): Span {
         if (opentracing_tags.SAMPLING_PRIORITY in keyValuePairs) {
             this._setSamplingPriority(keyValuePairs[opentracing_tags.SAMPLING_PRIORITY]);
+            delete keyValuePairs[opentracing_tags.SAMPLING_PRIORITY];
         }
 
         if (this._spanContext.isSampled()) {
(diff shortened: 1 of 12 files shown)


src/tracer.js
```diff
@@ -125,7 +126,25 @@ export default class Tracer {
...
+
+        if (fields.childOf) {
+            // Convert from a Span or a SpanContext into a Reference.
+            let childOf = opentracing.childOf(fields.childOf);
+            if (fields.references) {
+                fields.references.push(childOf);
+            } else {
+                fields.references = [childOf];
+            }
+            delete(fields.childOf);
```
yurishkuro: why?

src/tracer.js
```diff
@@ -125,7 +126,25 @@ export default class Tracer {
...
+        }
+        return this._startSpan(fields);
```
yurishkuro: nothing else calls _startSpan afaik, so could simply continue the function.

src/tracer.js
yurishkuro: also, doesn't seem to make sense in light of L155. I suggest rewriting it into a single piece of logic, skipping the call to `opentracing.childOf(fields.childOf)` altogether

src/tracer.js
```diff
@@ -125,7 +126,25 @@ export default class Tracer {
...
+    }
+
+    _startSpan(fields: startSpanArgs): Span {
```
yurishkuro: for L171 - what happens if span kind is not in the hash?

test/remote_reporter.js
```diff
@@ -49,7 +49,7 @@ describe('Remote Reporter should', () => {
     });
 
     it ('report span, and flush', () => {
-        let span = tracer.startSpan({operationName: 'operation-name'});
+        let span = tracer._startSpan({operationName: 'operation-name'});
```
yurishkuro: why?

test/span.js
```diff
@@ -123,7 +123,7 @@ describe('span should', () => {
     it('add logs with timestamp, and event', () => {
         let timestamp = new Date(2016, 8, 12).getTime();
         let event = 'some messgae';
-        span.log({ timestamp, event });
+        span._log({ timestamp, event });
```
yurishkuro: I don't follow - what are all these _ methods, why do we need them, and why would tests run against them instead of the public API?

test/span.js
```diff
@@ -143,7 +143,7 @@ describe('span should', () => {
         // mock global clock
         let clock = sinon.useFakeTimers(expectedTimestamp);
         let event = 'some messgae';
-        span.log({ event });
+        span._log({ event });
```
yurishkuro: isn't this supposed to be `{ event: 'some message' }`?

src/span.js
```diff
@@ -140,7 +145,8 @@ export default class Span {
      **/
     finish(finishTime: number): void {
         if (this._duration !== undefined) {
-            throw 'You can only call finish() on a span once.';
+        console.log('coolio');
```
yurishkuro: rm debug line

src/span.js
```diff
@@ -155,16 +161,31 @@ export default class Span {
...
+     * Adds a single tag to a span
+     *
+     * @param {string} key - The key for the tag added to this span.
+     * @param {string} value - The value corresponding with the key 
+     * for the tag added to this span.
+     * @return {Span} - returns this span.
+     * */
+    setTag(key: string, value: string): Span {
+        if (this._spanContext.isSampled()) {
+            this._tags.push({key, value});
```
yurishkuro: doesn't look like `sampling.priority` is handled, add TODO

yurishkuro: actually there's a method below, so we can fix it now

src/span.js
```diff
@@ -195,6 +216,20 @@ export default class Span {
         }
```
yurishkuro: the log() method will likely be different: https://github.com/opentracing/opentracing-javascript/pull/53/files#diff-a334dd7a261db56bf2558d532ee01111R200

we may want to implement the new format, and not require just event/payload to be in the fields.

oibe: I would rather do that in another diff, because the ability to use "setTags" is blocking my new sampler tags diff.

src/span.js
```diff
@@ -192,19 +191,19 @@ export default class Span {
...
                 this._logger.error('log must be passed either an event of type string, or a payload of type object');
                 return;
             }
 
             if (!fields.timestamp) {
-                fields.timestamp = Utils.getTimestampMicros();
+                fields.timestamp = timestamp || Utils.getTimestampMicros();
             }
 
             let value = fields.event;
```
yurishkuro: 3 lines below - you simply want to push `fields` hash, not a new hash. There's nothing special about event/payload keys anymore.

src/span.js
```diff
@@ -242,6 +241,8 @@ export default class Span {
         let tags = [];
         for (let i = 0; i < this._tags.length; i++) {
             let tag = this._tags[i];
+            let emptyBuffer = new Buffer(1);
+            emptyBuffer.fill(0);
             tags.push({
                 key: tag.key,
                 tagType: 'STRING', // TODO(oibe) support multiple tag types
```
yurishkuro: next line - you need to convert the value to string if it's not a string already

src/span.js
```diff
@@ -155,35 +160,50 @@ export default class Span {
...
      **/
-    addTags(keyValuePairs: any): void {
-        if (this._spanContext.isSampled()) {
-            for (let key in keyValuePairs) {
-                if (keyValuePairs.hasOwnProperty(key)) {
-                    let value = keyValuePairs[key];
-                    this._tags.push({key, value});
-                }
+    addTags(keyValuePairs: any): Span {
+        for (let key in keyValuePairs) {
```
yurishkuro: why did you loose isSampled check?

src/span.js
```diff
@@ -155,35 +160,50 @@ export default class Span {
...
+            if (keyValuePairs.hasOwnProperty(key)) {
+                let value = keyValuePairs[key];
+                this.setTag(key, value);
```
yurishkuro: this is not very efficient, I would append to _tags directly here, or use a helper method _setTag which does not do the check for isSampled over and over

src/span.js
oibe: because setTag does the isSampled check.

src/span.js
```diff
@@ -155,46 +161,80 @@ export default class Span {
...
+     * for the tag added to this span.
+     * @return {Span} - returns this span.
+     * */
+    setTag(key: string, value: any): Span {
+        if (key === opentracing_tags.SAMPLING_PRIORITY) {
+            this._setSamplingPriority(value);
+        }
+
+        if (this._spanContext.isSampled()) {
+            this._tags.push({key, value});
```
yurishkuro: {'key': key, 'value': value}  ?

src/span.js
```diff
@@ -155,46 +161,80 @@ export default class Span {
...
+     * Adds a single tag to a span
+     *
+     * @param {string} key - The key for the tag added to this span.
+     * @param {string} value - The value corresponding with the key 
+     * for the tag added to this span.
+     * @return {Span} - returns this span.
+     * */
+    setTag(key: string, value: any): Span {
+        if (key === opentracing_tags.SAMPLING_PRIORITY) {
+            this._setSamplingPriority(value);
```
yurishkuro: our other tracers do not store the sampling priority tag itself

src/span.js
oibe: es6 short hand for {'key': key, 'value': value} is {key, value}.

PR #10, opened 2016-09-14
Record sampling strategy as Span tags

oibe:


src/constants.js
```diff
@@ -18,5 +18,36 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
+// SAMPLED_MASK represents the power of two that indicates we have set the sampled
```
yurishkuro: a common term is "bitmap mask". And explanation shouldn't refer back to "flag", as it makes is a tautology.

SAMPLED_MASK is a bitmap mask indicating that the trace has been sampled for collection into storage.

src/samplers/const_sampler.js
```diff
@@ -19,11 +19,18 @@
...
+
 export default class ConstSampler {
     _decision: boolean;
+    _tags: Array<Tag>;
 
     constructor(decision: boolean) {
         this._decision = decision;
+        this._tags = [
+            {'key': constants.SAMPLER_TYPE_TAG_KEY, 'value': constants.SAMPLER_TYPE_CONST},
+            {'key': constants.SAMPLER_PARAM_TAG_KEY, 'value': this._decision.toString()}
```
yurishkuro: you do not need `toString()` it, that will be done by the span if needed, and in the future won't be needed at all because we can preserve the type of the tag explicitly in jaeger model.

src/samplers/probabilistic_sampler.js
```diff
@@ -19,14 +19,21 @@
...
+    _tags: Array<Tag>;
 
     constructor(samplingRate: number) {
         if (samplingRate < 0.0 || samplingRate > 1.0) {
             throw new Error(`The sampling rate must be less than 0.0 and grater than 1.0. Received ${samplingRate}`);
         }
         this._samplingRate = samplingRate;
+        this._tags = [
+            {'key': constants.SAMPLER_TYPE_TAG_KEY, 'value': constants.SAMPLER_TYPE_PROBABILISTIC},
+            {'key': constants.SAMPLER_PARAM_TAG_KEY, 'value': `${this._samplingRate}`}
```
yurishkuro: don't stringify

src/samplers/ratelimiting_sampler.js
```diff
@@ -37,6 +39,10 @@ export default class RateLimitingSampler {
 
         this._maxTracesPerSecond = maxTracesPerSecond;
         this._rateLimiter = new RateLimiter(maxTracesPerSecond);
+        this._tags = [
+            {'key': constants.SAMPLER_TYPE_TAG_KEY, 'value': constants.SAMPLER_TYPE_RATE_LIMITING},
+            {'key': constants.SAMPLER_PARAM_TAG_KEY, 'value': `${this._maxTracesPerSecond}`}
```
yurishkuro: don't stringify

src/span.js
```diff
@@ -229,6 +233,13 @@ export default class Span {
         });
     }
 
+    _setTracerTags(tags: Array<Tag>): void {
+        for (let i = 0; i < tags.length; i++) {
+            let tag = tags[i];
+            this._tags.push(tag);
```
yurishkuro: there's no addAll() like in Java?

src/tracer.js
```diff
@@ -74,6 +74,7 @@ export default class Tracer {
             spanContext: SpanContext,
             operationName: string,
             startTime: number,
+            internalTags: any = [],
```
yurishkuro: we're doing a lot of defaulting by creating new objects, I wonder how that impacts the benchmarks.

src/tracer.js
```diff
@@ -90,12 +91,20 @@ export default class Tracer {
             span._setSamplingPriority(tags[opentracing_tags.SAMPLING_PRIORITY])
         }
 
+        // set all sampler tags
+        for (let i = 0; i < internalTags.length; i++) {
```
yurishkuro: * why not use `span.addTags(internalTags)` ?
* the comment is redundant

src/span.js
yurishkuro: also, how is this different from span.addTags() ?

src/tracer.js
```diff
@@ -90,12 +91,20 @@ export default class Tracer {
...
+
+        // set all tags provided by instrumentation
         span.addTags(tags);
         return span;
     }
 
     _report(span: Span): void {
-        span.addTags(this._tracerTags);
+        if (span.firstInProcess) {
+            span._setTracerTags(this._tags);
```
yurishkuro: s/../span.addTags()/ ?

test/all_samplers.js
```diff
@@ -0,0 +1,48 @@
...
+
+describe('samplers should', () => {
+    it('return correct tags', () => {
+        var samplers = [
+            {'sampler': new ConstSampler(true), 'type': constants.SAMPLER_TYPE_CONST, 'param': 'true'},
+            {'sampler': new ConstSampler(false), 'type': constants.SAMPLER_TYPE_CONST, 'param': 'false'},
+            {'sampler': new ProbabilisticSampler(0.1), 'type': constants.SAMPLER_TYPE_PROBABILISTIC, 'param': '0.1'},
+            {'sampler': new RateLimitingSampler(2), 'type': constants.SAMPLER_TYPE_RATE_LIMITING, 'param': '2'},
+            {'sampler': new RemoteSampler('some-caller-name'), 'type': constants.SAMPLER_TYPE_PROBABILISTIC, 'param': '0.001'},
+        ];
```
yurishkuro: neat!

src/span.js
oibe: It takes an array, but I can make internal tags use an object instead.

src/tracer.js
oibe: I'm not worried about it, because a lot of people do this manually in es5 code.  Even in tchannel this is the pattern.

src/span.js
yurishkuro: less (methods) is better

oibe: i'm sure lodash has something like addAll, but is not more efficient than using a for loop.

src/tracer.js
```diff
@@ -86,12 +87,17 @@ export default class Tracer {
             firstInProcess
         );
 
+        // set all tags
```
yurishkuro: rm comment

oibe:
merged in another pr

PR #12, opened 2016-09-15
add long.js to jaeger-client-node

### LINKED ISSUE #7: Consider using Long instead of buffers for IDs
According to @Raynos:

> Keeping i64 in buffer is expensive for numerous reasons, prefer Long format.
>  - buffers should have a "short life time", keeping them around will bloat memory ( they are internally slab allocated )
>  - allocating a buffer slice is more expensive then an object with two integer fields

I see that Long has to/from hex string, so we should be using it instead of buffers:

```
Long#toString(radix=)
Long.fromString(str, unsigned=, radix=)
```

https://www.npmjs.com/package/long#longtostringradix

oibe:
Fixes #7

### DIFF
--- a/src/span_context.js
+++ b/src/span_context.js
@@ -20,19 +20,19 @@
 // THE SOFTWARE.
 
 import * as constants from './constants.js';
-import Int64 from 'node-int64';
 import Utils from './util.js';
+import Long from 'long';
 
 export default class SpanContext {
-    _traceId: any;
-    _spanId: any;
-    _parentId: any;
+    _traceId: Long;
+    _spanId: Long;
+    _parentId: Long;
     _flags: number;
     _baggage: any;
 
-    constructor(traceId: any,
-                spanId: any,
-                parentId: any,
+    constructor(traceId: Long,
+                spanId: Long,
+                parentId: Long,
                 flags: number,
                 baggage: any = {}) {
         this._traceId = traceId;
@@ -42,15 +42,16 @@ export default class SpanContext {
         this._baggage = baggage;
     }
 
-    get traceId(): any {
+    get traceId(): Long {
         return this._traceId;
     }
 
-    get spanId(): any {
+
+    get spanId(): Long {
         return this._spanId;
     }
 
-    get parentId(): any {
+    get parentId(): Long {
         return this._parentId;
     }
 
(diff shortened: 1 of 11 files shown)


src/span_context.js
```diff
@@ -101,11 +102,11 @@ export default class SpanContext {
...
      **/
     toString(): string {
-        var parentId = this._parentId ? this._parentId.toString('hex') : '0';
+        var parentId = this._parentId ? this._parentId.toString(16) : '0';
 
         return [
-            Utils.removeLeadingZeros(this._traceId.toString('hex')),
-            Utils.removeLeadingZeros(this._spanId.toString('hex')),
+            Utils.removeLeadingZeros(this._traceId.toString(16)),
+            Utils.removeLeadingZeros(this._spanId.toString(16)),
```
yurishkuro: is removeLeadingZeros() still needed even with Long?

src/span_context.js
```diff
@@ -128,26 +129,29 @@ export default class SpanContext {
...
 
         if (NaNDetected) {
             return null;
         }
 
-        let parentId = null;
-        if (headers[2] !== '0') {
-            parentId = Utils.encodeInt64(headers[2]);
+        // The collector expects null instead of zero.
+        if (parentId.toNumber() === 0) {
```
yurishkuro: what happens if parent ID is not 0 but a number > 2^53? Will toNumber() work still?  It seems like we can just check if the string was 0 to begin with.

src/util.js
```diff
@@ -147,4 +119,17 @@ export default class Utils {
         }
     }
 
+    static parseHex64(inputStr: string): any {
+        let hexLookup = {'0': '0', '1': '1', '2': '2', '3': '3', '4': '4', '5': '5', '6': '6', '7': '7', '8': '8',
```
yurishkuro: shouldn't this be static?

src/util.js
```diff
@@ -147,4 +119,17 @@ export default class Utils {
...
+            '9': '9', 'a': 'a', 'b': 'b', 'c': 'c', 'd': 'd', 'e': 'e', 'f': 'f'};
+        let unsigned = true;
+
+        for (let i = 0 ; i < inputStr.length; i++) {
+            if(!(inputStr[i] in hexLookup)) {
```
yurishkuro: you don't need the map above, simply a condition
```
let c = inputStr[i]
if (!(c >= '0' && c <= '9' || c >= 'A' && c <= 'F' || c >= 'a' && c <= 'f')) {
  return Number.NaN;
}
```

yurishkuro:
Could you please add the before and after benchmarks? What improvement are we getting with this change?

PR #13, opened 2016-09-15
Add benchmark code

oibe:


benchmarks/core_flows.js
```diff
@@ -0,0 +1,91 @@
...
+        {'format': opentracing.FORMAT_TEXT_MAP, 'tracer': constTracer, 'carrier': textCarrier, description: ' with const tracer and text carrier'},
+        {'format': opentracing.FORMAT_HTTP_HEADERS, 'tracer': probabilisticTracer, 'carrier': httpCarrier, description: ' with probabilistic tracer and http carrier'},
+        {'format': opentracing.FORMAT_TEXT_MAP, 'tracer': probabilisticTracer, 'carrier': textCarrier, description: ' with probabilistic tracer and text carrier'},
+    ];
+
+    function run(tracer, carrier, format, description) {
+        var suite = new Benchmark.Suite(tracer, carrier)
+            .add('Fundamental Request flow', function() {
+                var context = tracer.extract(format, carrier);
+                var fields = {
```
yurishkuro: s/fields/options/

package.json
```diff
@@ -19,13 +19,16 @@
     "xorshift": "^0.2.0"
```
yurishkuro: * L15 - is lodash a required runtime dependency? Can we avoid it?
* what is node-int64 used for?
* do we need assert as runtime dep?
* what is "request" module?

benchmarks/core_flows.js
```diff
@@ -0,0 +1,91 @@
...
+    var textCarrier = {};
+    constTracer.inject(span.context(), opentracing.FORMAT_TEXT_MAP, textCarrier);
+
+    var probabilisticTracer = new Tracer('probabilistic-tracer', new InMemoryReporter(), new ProbabilisticSampler(0.01));
+    var params = [
+        {'format': opentracing.FORMAT_HTTP_HEADERS, 'tracer': constTracer, 'carrier': httpCarrier, description: ' with const tracer and http carrier'},
+        {'format': opentracing.FORMAT_TEXT_MAP, 'tracer': constTracer, 'carrier': textCarrier, description: ' with const tracer and text carrier'},
+        {'format': opentracing.FORMAT_HTTP_HEADERS, 'tracer': probabilisticTracer, 'carrier': httpCarrier, description: ' with probabilistic tracer and http carrier'},
+        {'format': opentracing.FORMAT_TEXT_MAP, 'tracer': probabilisticTracer, 'carrier': textCarrier, description: ' with probabilistic tracer and text carrier'},
+    ];
```
yurishkuro: We probably also need start trace vs. join trace dimension

benchmarks/core_flows.js
```diff
@@ -0,0 +1,91 @@
...
+
+    function run(tracer, carrier, format, description) {
+        var suite = new Benchmark.Suite(tracer, carrier)
+            .add('Fundamental Request flow', function() {
+                var context = tracer.extract(format, carrier);
```
yurishkuro: is this context intentionally unused?

benchmarks/flame_graph_loop.js
```diff
@@ -0,0 +1,61 @@
...
+}
+
+var tracer = new Tracer(
+    'flamegraph-tracer',
+    new InMemoryReporter(),
+    new ConstSampler(true)
+);
+
+var carrier = {
+    "UBER-TRACE-ID": encodeURIComponent('10:10:0:3')
```
yurishkuro: note: ":3" at the end means this cycle is not making any sampling decisions

benchmarks/span.js
```diff
@@ -0,0 +1,84 @@
...
+function benchmarkSpan() {
+    var tracer = new Tracer('benchmark-tracer', new InMemoryReporter(), new ConstSampler(true));
+    var spanWithBaggage = tracer.startSpan('op-name');
+    spanWithBaggage.setBaggageItem('key', 'value');
+
+    console.log('Beginning Span Benchmark...');
+    function run(span) {
+        var suite = new Benchmark.Suite()
+            .add('Span:setBaggageItem', function() {
+                span.setBaggageItem('key', 'value');
```
yurishkuro: this suite doesn't make sense to me. If you continue adding tags or baggage to the same span, the internal collections of the span will keep growing, getting slower and slower.

benchmarks/span.js
```diff
@@ -0,0 +1,84 @@
...
+var Benchmark = require('benchmark');
+var benchmarks = require('beautify-benchmark');
+var ConstSampler = require('../dist/src/samplers/const_sampler.js').default;
+var InMemoryReporter = require('../dist/src/reporters/in_memory_reporter.js').default;
+var SpanContext = require('../dist/src/span_context.js').default;
+var Tracer = require('../dist/src/tracer.js').default;
+var opentracing = require('opentracing');
+
+function benchmarkSpan() {
+    var tracer = new Tracer('benchmark-tracer', new InMemoryReporter(), new ConstSampler(true));
```
yurishkuro: calling ops on sampled and non-sampled span has different perf. If anything, we're probably more interested in the perf of un-sampled spans since they occur a lot more often,

benchmarks/tracer.js
```diff
@@ -0,0 +1,89 @@
...
+
+    function run(tracer, carrier, format, description) {
+        var suite = new Benchmark.Suite()
+            .add('Tracer:startSpan', function() {
+                tracer.startSpan({operationName: 'test-op'});
+            })
+            .add('Tracer:startSpan:childOf', function() {
+                // note for multiple outbound client calls context wouldn't be allocated
+                // multiple times.
+                var context = SpanContext.fromString('10:19:0:1');
```
yurishkuro: if you're measuring startSpan function, you should use a const context defined outside

benchmarks/core_flows.js
oibe: no it should be part of childOf

package.json
oibe: - lodash can be removed
- node-int64 will be removed in another diff
- assert is not needed
- request is for making http calls in the sampler

benchmarks/span.js
oibe: not if its the same key each time.

benchmarks/core_flows.js
```diff
@@ -54,13 +54,14 @@ function benchmarkCoreFlows() {
         var suite = new Benchmark.Suite(tracer, carrier)
             .add('Fundamental Request flow', function() {
                 var context = tracer.extract(format, carrier);
-                var fields = {
+                var optionss = {
```
yurishkuro: typo - to ss

yurishkuro: how is it that the checker did not catch this?

benchmarks/flame_graph_loop.js
```diff
@@ -23,10 +23,10 @@ var InMemoryReporter = require('./dist/src/reporters/in_memory_reporter.js').def
 var Tracer = require('./dist/src/tracer.js').default;
 var opentracing = require('opentracing');
 
-function server_client_span_loop(tracer, carrier) {
+function server_client_span_loop(tracer) {
     for ( var i = 0; i <= 0; i--) {
         // deserialize parent context from carrier
-        var parentContext = tracer.extract(opentracing.FORMAT_HTTP_HEADERS, carrier);
+        var parentContext = tracer.extract(opentracing.FORMAT_HTTP_HEADERS, {});
```
yurishkuro: empty carrier makes this a no-op, did you mean that?

oibe: no...

benchmarks/core_flows.js
oibe: this is es5, and not type checked.  Nor does it need to be.

benchmarks/flame_graph_loop.js
oibe: yeah but, the way I coded it is a bit strange, because I want context to be null.  So I could have just emitted this line.

PR #16, opened 2016-09-26
add additional benchmarks for thrift seralization

oibe:


benchmarks/span.js
```diff
@@ -34,22 +34,24 @@ function benchmarkSpan() {
...
-        probabilisticTracer
+        { 'tracer': constTracer, 'description': ' with constant tracer' },
+        { 'tracer': probabilisticTracer, 'description': ' with probabilitic tracer' }
     ];
 
     function createSpanFactory(tracer) {
         return function() {
             var span = tracer.startSpan('op-name');
             span.setBaggageItem('key', 'value');
+            span.finish();
```
yurishkuro: this doesn't make sense. The only legal method to call after finish() is span.context(), so all the benchmarks are doing the "Wrong" things

PR #17, opened 2016-09-26
Add crossdock tests for node

oibe:


Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM woorank/docker-node-babel
+EXPOSE 8080-8082
+
+ADD node_modules/ /node_modules
```
yurishkuro: Hm, I am not sure if this doesn't introduce a conflict between host OS and container OS in case these modules contain pre-compiled native code

Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+ADD package.json /
+ADD src/ /src
+ADD crossdock/src /crossdock/src
+ADD .babelrc /
+
+RUN npm install tchannel
```
yurishkuro: to illustrate the point above, why do we need to re-install tchannel if we already copied all modules? Is it because tchannel (or its dependencies like thriftrw) uses native code?

crossdock/rules.mk
```diff
@@ -0,0 +1,27 @@
...
+
+.PHONY: crossdock-logs
+crossdock-logs:
+	docker-compose -f $(XDOCK_YAML) logs
+
+#foo.bak: foo.bar
+#	    echo "foo"
+#		    if [ -d "~/Dropbox" ]; then \
+#				        echo "Dir exists"; \
+#						    fi
```
yurishkuro: rm

crossdock/src/driver.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+/node_modules/.bin/babel-node /crossdock/src/http_server.js &
+/node_modules/.bin/babel-node /crossdock/src/tchannel_server.js &
```
yurishkuro: this is weird. Can we not start these in the same process?

crossdock/src/http_server.js
```diff
@@ -0,0 +1,211 @@
...
+import SpanContext from '../../src/span_context.js';
+import Tracer from '../../src/tracer.js';
+import Utils from '../../src/util.js';
+
+export default class HttpServer {
+    _tracer: Tracer;
+
+    constructor() {
+        let app = express();
+        this._tracer = new Tracer('crossdock', new InMemoryReporter(), new ConstSampler(false));
```
yurishkuro: s/crossdock/node/ ?

crossdock/src/http_server.js
```diff
@@ -0,0 +1,211 @@
...
+
+            // create or deserialize a span from the headers
+            let spanContext = this._tracer.extract(opentracing.FORMAT_HTTP_HEADERS, req.headers);
+            console.log('start_trace_extract_span_context:', spanContext);
+            let serverSpan = this._tracer.startSpan('start-trace', { childOf: spanContext });
+
+            // set sampling and baggage on the span
+            serverSpan.setBaggageItem(constants.BAGGAGE_KEY, startRequest.baggage);
+            if (startRequest.sampled) {
+                serverSpan._setSamplingPriority(1);
```
yurishkuro: crossdock tests must use public API

crossdock/src/http_server.js
```diff
@@ -0,0 +1,211 @@
...
+        this._tracer = new Tracer('crossdock', new InMemoryReporter(), new ConstSampler(false));
+
+        // json responses need bodyParser when working with express
+        app.use(bodyParser.json());
+
+        app.post('/start_trace', (req, res) => {
+            let startRequest = req.body;
+            console.log('start_trace_request:', startRequest);
+
+            // create or deserialize a span from the headers
```
yurishkuro: need to DRY this - beyond this point there's no difference in the logic between start and join, except for additional span initialization in startTrace that can be done via a lambda passed to a shared function.

crossdock/src/http_server.js
```diff
@@ -0,0 +1,211 @@
...
+                    resolve(response);
+                });
+            });
+
+            // then return result of prepareResponse
+            promise.then((response) => {
+                let traceResponse = JSON.stringify(response);
+                console.log('join_trace_response:', traceResponse);
+                res.send(traceResponse);
+            });
```
yurishkuro: what happens on error?

crossdock/src/http_server.js
```diff
@@ -0,0 +1,211 @@
...
+            dns.lookup(downstream.host, (err, address, family) => {
+                if (err) {
+                    return;
+                }
+
+                let port = parseInt(downstream.port);
+                let downstreamUrl = `http:\/\/${address}:${port}/join_trace`;
+
+                let clientSpan = this._tracer.startSpan('client-span', { childOf: context.span.context() });
+                context.span = clientSpan;
```
yurishkuro: wouldn't this mess up the server span that the server endpoint might try to call Finish on?

crossdock/src/http_server.js
```diff
@@ -0,0 +1,211 @@
...
+                        'downstream': downstream.downstream
+                    })
+                }, (err, response) => {
+                    if (err) {
+                        console.log('error in downstream call:', err);
+                        reject(err);
+                        return;
+                    }
+
+                    clientSpan.finish();
```
yurishkuro: should be finished in case of the error too

crossdock/src/tchannel_server.js
```diff
@@ -0,0 +1,97 @@
...
+        callback(null, {
+                'ok': true,
+                'body': {
+                    'notImplementedError': 'TChannel crossdock not implemented for node.',
+                    'span': span
+                }
+            });
+    }
+
+    observeSpan(context: any): ObservedSpan {
```
yurishkuro: duplicated from http server

yurishkuro:
Travis build needs to include crossdock run and crossdock publish. The latter could be in separate PR, but the main run of the crossdock tests should be here, otherwise we're adding dead code.

yurishkuro:
I am not seeing any tests for crossdock code itself

Dockerfile
oibe: only for tchannel, and when other our discovered I will add them separately to the docker file.

crossdock/src/driver.sh
oibe: we could, but why not start them each as a separate process?  This implementation has better performance technically.

yurishkuro: you already have index.js below that seems to start them all in one process, so why two different ways?

oibe: index.js is used for local testing, i actually mean to delete that.

crossdock/src/http_server.js
oibe: I just print the error on an http request, or the logs show a stack trace.

crossdock/src/http_server.js
oibe: I don't really need this line actually because context won't be used again.

PR #18, opened 2016-09-26
update to the latest jaeger-idl

oibe:


src/reporters/in_memory_reporter.js
```diff
@@ -24,10 +24,13 @@ import Span from '../span.js';
 export default class InMemoryReporter {
     _spans: Array<Span>;
     _flushed: Array<Span>;
+    _process: Process;
+    _sender: Sender
 
-    constructor() {
+    constructor(sender: Sender) {
```
yurishkuro: only remote reporter needs a sender. What's the point of having it here?

src/reporters/noop_reporter.js
```diff
@@ -22,6 +22,8 @@
 import Span from '../span.js';
 
 export default class NoopReporter {
+    _process: Process;
+    _sender: Sender;
```
yurishkuro: same

src/constants.js
```diff
@@ -49,3 +49,7 @@ export const SAMPLER_TYPE_RATE_LIMITING = 'ratelimiting';
 
 // SAMPLER_TYPE_REMOTE is the type of sampler that polls Jaeger agent for sampling strategy.
 export const SAMPLER_TYPE_REMOTE = "remote"
+
+// EMIT_SPAN_BATCH_OVERHEAD is the rough estimate of header space taken up in a
+// thrift frame.
+export const EMIT_SPAN_BATCH_OVERHEAD = 30;
```
yurishkuro: why does this need to be public? only udp sender cares about this value

src/reporters/udp_sender.js
```diff
@@ -46,7 +47,7 @@ export default class UDPSender {
                 maxPacketSize: number = UDP_PACKET_MAX_LENGTH) {
         this._hostPort = hostPort;
         this._maxPacketSize = maxPacketSize;
-        this._maxSpanBytes = this._maxPacketSize - EMIT_SPAN_BATCH_OVERHEAD;
+        this._maxSpanBytes = this._maxPacketSize - constants.EMIT_SPAN_BATCH_OVERHEAD;
```
yurishkuro: this old logic is now completely wrong, because the overhead depends on the size of the Process thrift.

src/span.js
```diff
@@ -37,7 +41,12 @@ export default class Span {
     _peer: Endpoint;
     _logs: Array<LogData>;
     _tags: Array<Tag>;
+    static emptyBuffer: Buffer = new Buffer(8);
     static _baggageHeaderCache: any;
+    static _thrift = new Thrift({
+        source: fs.readFileSync(path.join(__dirname, './jaeger-idl/thrift/jaeger.thrift'), 'ascii'),
```
yurishkuro: is this file (from submodule) going to be available when the library is used as a dependency?

src/span.js
```diff
@@ -207,23 +216,25 @@ export default class Span {
...
+    log(keyValuePairs: any, timestamp: ?number): void {
         if (this._spanContext.isSampled()) {
-            if (!fields.event && !fields.payload) {
-                this._logger.error('log must be passed either an event of type string, or a payload of type object');
-                return;
-            }
-
-            if (!fields.timestamp) {
-                fields.timestamp = timestamp || Utils.getTimestampMicros();
+            let tags = [];
```
yurishkuro: s/tags/fields/

src/span.js
```diff
@@ -207,23 +216,25 @@ export default class Span {
     /**
      * Adds a log event, or payload to a span.
      *
-     * @param {Object} fields - an object that represents the fields to log.
+     * @param {Object} keyValuePairs - an object that represents the keyValuePairs to log.
      * @param {number} [timestamp] - the starting timestamp of a span.
-     * @param {string} [fields.event] - a string message to be logged on a span.
-     * @param {string} [fields.payload] - an object to be logged on a span as a json string.
+     * @param {string} [keyValuePairs.event] - a string message to be logged on a span.
+     * @param {string} [keyValuePairs.payload] - an object to be logged on a span as a json string.
```
yurishkuro: I don't think we need to keep these two

src/span.js
```diff
@@ -249,35 +260,97 @@ export default class Span {
...
-                vInt16: 0,
-                vInt32: 0,
-                vInt64: 0,
-                vBinary: emptyBuffer
+    static _getTagType(tag: Tag): string {
+        if (tag.value instanceof Number) {
+            return Span._thrift.TagType.DOUBLE;
+        } else if (tag.value instanceof Boolean) {
+            return Span._thrift.TagType.BOOL;
+        } else if (tag.value instanceof Long) { //TODO(oibe) how else to recognize a long?
```
yurishkuro: seems reasonable

src/span.js
```diff
@@ -249,35 +260,97 @@ export default class Span {
...
+            return Span._thrift.TagType.LONG;
+        } else if (tag.value instanceof Buffer) {
+            return Span._thrift.TagType.BUFFER;
```
yurishkuro: buffer? We only have these: 

`enum TagType { STRING, DOUBLE, BOOL, LONG, BINARY }`

Do you have tests exercising all branches of this if?

src/span.js
```diff
@@ -249,35 +260,97 @@ export default class Span {
...
+        }
+    }
+
+    static _getThriftTags(initialTags: Array<Tag>): Array<any> {
+        let thriftTags = [];
+        for (let i = 0; i < initialTags.length; i++) {
+            let tag = initialTags[i];
+
+            let key: string = tag.key;
+            let vType: string = Span._getTagType(tag);
```
yurishkuro: this is doing double work of switching on the type, why not combine it in a single if?

src/span.js
```diff
@@ -249,35 +260,97 @@ export default class Span {
...
+                'fields': Span._getThriftTags(log.fields)
             });
         }
 
+        return thriftLogs;
+    }
+
+    _toThrift() {
+        let tags = Span._getThriftTags(this._tags);
+        let logs = this._getThriftLogs();
```
yurishkuro: can we be consistent and have all _thrift functions static? Also, I would move them to a separate `thrift.js` file

src/span.js
```diff
@@ -249,35 +260,97 @@ export default class Span {
...
+        let unsigned = true;
+
         return {
-            traceId : this._spanContext.traceId,
+            traceIdLow: this._spanContext.traceId,
+            traceIdHigh: Span.emptyBuffer,  // TODO(oibe) implement 128 bit ids
             spanId: this._spanContext.spanId,
+            parentSpanId: this._spanContext.parentId || Span.emptyBuffer, //TODO(oibe) how should thrift send null?
```
yurishkuro: we can send it as all zeroes

src/tracer.js
```diff
@@ -68,6 +68,23 @@ export default class Tracer {
...
+        for (let key in this._tags) {
+            let value = this._tags[key];
+            if (this._tags.hasOwnProperty(key)) {
+                tags.push({ 'key': key, 'value': value });
+            }
+        }
+
+        this._reporter.setProcess({
+            'serviceName': this._serviceName,
+            'tags': Span._getThriftTags(tags)
```
yurishkuro: here's example why _thrift methods belong to a separate file, not span.js

src/tracer.js
```diff
@@ -68,6 +68,23 @@ export default class Tracer {
...
+        });
```
yurishkuro: similar to comment in span.js, the thrift logic does not belong in Tracer. The API to the reporter can be `setProcess(serviceName, tags)`, and thrift construction can be deferred to as late as needed, e.g. in the sender, or at least the reporter (since interface to sender is already thrift)

src/reporters/in_memory_reporter.js
oibe: I have it here for testing.

src/span.js
oibe: it will now.

src/constants.js
yurishkuro: ?

src/reporters/in_memory_reporter.js
yurishkuro: ?

src/reporters/udp_sender.js
```diff
@@ -19,34 +19,37 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
+import * as constants from '../constants.js';
```
yurishkuro: don't think this needed

src/reporters/udp_sender.js
```diff
@@ -19,34 +19,37 @@
...
     _byteBufferSize: number;
     _thrift: any;
+    _batch: Batch;
 
     constructor(hostPort: string = DEFAULT_UDP_SPAN_SERVER_HOST_PORT,
                 maxPacketSize: number = UDP_PACKET_MAX_LENGTH) {
         this._hostPort = hostPort;
         this._maxPacketSize = maxPacketSize;
-        this._maxSpanBytes = this._maxPacketSize - EMIT_SPAN_BATCH_OVERHEAD;
+        this._maxSpanBytes = this._maxPacketSize;
```
yurishkuro: it would be safer not to initialize it here (to incorrect vaue), but only do it in setProcess

src/reporters/udp_sender.js
```diff
@@ -57,12 +60,30 @@ export default class UDPSender {
...
 
+    setProcess(process: Process): void {
+        // This function is only called once during reporter construction, and thus will
+        // give us the length of the batch before any spans have been added to _spanBuffer.
+        this._process = process;
+        this._batch = {
+            'process': this._process,
+            'spans': this._spanBuffer
+        };
+        this._emitSpanBatchOverhead = this._calcBatchSize(this._batch);
```
yurishkuro: I don't think this is correct. The actual UDP packet also contains a wrapper struct, something like `emitBatchArg`

src/span.js
```diff
@@ -207,23 +207,23 @@ export default class Span {
...
-            }
-
-            if (!fields.timestamp) {
-                fields.timestamp = timestamp || Utils.getTimestampMicros();
+            let fields = [];
+            for (let key in keyValuePairs) {
+                let value = keyValuePairs[key];
+                if (keyValuePairs.hasOwnProperty(key)) {
+                    fields.push({'key': key, 'value': value});
+                }
```
yurishkuro: what's the purpose of this conversion? Why not simply store `keyValuePairs` as `fields` in L225?  You're still going to iterate and re-convert to thrift log fields later.

src/tracer.js
```diff
@@ -68,6 +68,20 @@ export default class Tracer {
...
+    }
+
+    _setProcess(): void {
+        let tags = [];
+        for (let key in this._tags) {
+            let value = this._tags[key];
+            if (this._tags.hasOwnProperty(key)) {
+                tags.push({ 'key': key, 'value': value });
+            }
+        }
```
yurishkuro: what's the purpose of this conversion? It's less critical than with log fields, but I'd rather have fewer lines of code, and this repackaging of the tags/fields seems unnecessary.

yurishkuro:
please fix the build (I was looking for code coverage changes)

src/span.js
oibe: The function that takes in tags, and converts to thrift expects everything to be in {'key': 'some-key', 'value': 'some-value'} format.  So right here, and below for the tracer they need to become key value objects.  An alternative is to store everything including tags as a map, and then convert them all to key value pairs in the end.

src/tracer.js
oibe: see above.

src/span.js
yurishkuro: My preference is to keep tracer and span tags as array({key, value}), rather than a dictionary, because the former allows duplicate keys. However, there are a few places where the API accepts KVs as a map, e.g. options to startSpan, the fields in KV logging, etc. In those places we should call a util function to convert map to an array. But we shouldn't keep repeating the same code over and over.

src/reporters/remote_reporter.js
```diff
@@ -57,4 +59,15 @@ export default class RemoteReporter {
...
         clearInterval(this._intervalHandle);
     }
+
+    setProcess(serviceName: string, tags: Array<Tag>): void {
+        this._process = {
+            'serviceName': serviceName,
+            'tags': ThriftUtils.getThriftTags(tags)
+        };
+
+        if (this._sender) {
```
yurishkuro: I think reporter's constructor should panic if the sender is null, rather than checking here

src/reporters/udp_sender.js
```diff
@@ -57,12 +58,30 @@ export default class UDPSender {
         });
     }
 
+    _calcBatchSize(batch: Batch) {
+        let thriftBatch = this._thrift.getType('Agent::emitBatch_args');
```
yurishkuro: can we cache this?

src/reporters/udp_sender.js
```diff
@@ -57,12 +58,30 @@ export default class UDPSender {
...
+        let buffer: Buffer = thriftBatch.toBufferResult({'batch': batch}).value;
+        return buffer.length;
+    }
+
     _calcSpanSize(span: any): number {
         let thriftJaegerSpan = this._thrift.getType('Span');
```
yurishkuro: can we cache this?

src/reporters/udp_sender.js
```diff
@@ -57,12 +58,30 @@ export default class UDPSender {
...
+    setProcess(process: Process): void {
+        // This function is only called once during reporter construction, and thus will
+        // give us the length of the batch before any spans have been added to _spanBuffer.
+        this._process = process;
+        this._batch = {
+            'process': this._process,
+            'spans': this._spanBuffer
+        };
+        this._emitSpanBatchOverhead = this._calcBatchSize(this._batch);
+        this._maxSpanBytes -= this._emitSpanBatchOverhead;
```
yurishkuro: `this._maxPacketSize - this._emitSpanBatchOverhead;`

yurishkuro: in the current form this looks like it's negative or NaN. Do you check it in unit tests?

oibe: So this bug was masked by the fact that each of my tests happens to set maxSpanBytes for its own purposes.

yurishkuro: tests should manipulate the public API, e.g. the sender constructor can accept the maxPacketSize as argument.

src/reporters/udp_sender.js
```diff
@@ -90,33 +87,36 @@ export default class UDPSender {
...
+        this._batch.spans.push(span);
         this._byteBufferSize = spanSize;
         return flushResponse;
     }
 
     flush(testCallback: ?Function): SenderResponse {
-        let numSpans: number = this._spanBuffer.length;
+        let numSpans: number = this._batch.spans.length;
         if (numSpans == 0) {
             return {err: false, numSpans: 1}
```
yurishkuro: why `numSpans: 1` ?

yurishkuro: is there a unit test for this case?

src/reporters/remote_reporter.js
yurishkuro: ?

src/reporters/udp_sender.js
```diff
@@ -107,8 +127,34 @@ export default class UDPSender {
...
+        }
+
+        let tagMessages = [];
+        for (let j = 0; j < this._batch.process.tags.length; j++) {
+            let tag = this._batch.process.tags[j];
+            tagMessages.push(new this._thrift.Tag(tag));
+        }
+
+        return new this._thrift.Agent.emitBatch.ArgumentsMessage({
+            version: 1,
```
yurishkuro: is this a field that gets sent in the datagram?

src/reporters/udp_sender.js
```diff
@@ -107,8 +127,34 @@ export default class UDPSender {
...
+            id: 0,
+            body: {batch: new this._thrift.Batch({
+                    process: new this._thrift.Process({
```
yurishkuro: can't we do this once from setProcess?

src/reporters/udp_sender.js
oibe: I think so, the docs say to set version to 1 for enveloping.

src/reporters/udp_sender.js
yurishkuro: you need to assign _something_ to `process: ` every time, no argument. But why does that _something_ needs to be reconstructed every time, given that the arguments to the construction never change?

PR #19, opened 2016-09-27
add jaeger-debug-header with correlation id

oibe:


src/constants.js
```diff
@@ -48,4 +48,10 @@ export const SAMPLER_TYPE_PROBABILISTIC = 'probabilistic';
 export const SAMPLER_TYPE_RATE_LIMITING = 'ratelimiting';
 
 // SAMPLER_TYPE_REMOTE is the type of sampler that polls Jaeger agent for sampling strategy.
-export const SAMPLER_TYPE_REMOTE = "remote"
+export const SAMPLER_TYPE_REMOTE = 'remote';
+
+// JaegerDebugHeader is the name of an HTTP header or a TextMap carrier key which,
```
yurishkuro: name mismatch

src/propagators/text_map_codec.js
```diff
@@ -56,21 +57,28 @@ export default class TextMapCodec {
...
-            for (let key in carrier) {
-                if (carrier.hasOwnProperty(key) && Utils.startsWith(key, this._baggagePrefix)) {
+        // $FlowIgnore - I just want an empty span context.
+        let spanContext = new SpanContext();
+        let baggage = {};
+
+        for (let key in carrier) {
+            if (carrier.hasOwnProperty(key)) {
+                if (key === this._contextKey) {
+                    spanContext = SpanContext.fromString(this._decodedValue(carrier[this._contextKey]));
```
yurishkuro: depending on the order of keys, this can override what you've already processed for other keys.

src/span_context.js
```diff
@@ -82,6 +89,14 @@ export default class SpanContext {
         this._baggage = baggage;
     }
 
+    set debugId(debugId: string): void {
+        this._debugId = debugId;
+    }
+
+    isDebugIDContainerOnly(): boolean {
+        return !this._traceId && this._debugId !== '';
```
yurishkuro: does `!` work correctly on Long?

test/text_map_codec.js
```diff
@@ -0,0 +1,59 @@
...
+        let tracer = new Tracer(
+            'test-tracer',
+            new InMemoryReporter(),
+            new ConstSampler(true)
+        );
+        let headers = {};
+        headers[constants.JAEGER_DEBUG_HEADER] = encodeURIComponent('value1');
+
+        let context = tracer.extract(opentracing.FORMAT_HTTP_HEADERS, headers);
+        assert.isNotOk(context.parentId);
```
yurishkuro: why parentId? traceId would've made more sense. Also, you can call the new isDebugIDContainer method to validate.

src/span_context.js
oibe: I use ! because when i initialize an empty span context, all of its values will be undefined.  So traceId will always be undefined.

src/propagators/text_map_codec.js
```diff
@@ -56,21 +57,28 @@ export default class TextMapCodec {
...
-            let baggage = {};
-            for (let key in carrier) {
-                if (carrier.hasOwnProperty(key) && Utils.startsWith(key, this._baggagePrefix)) {
+        // $FlowIgnore - I just want an empty span context.
+        let spanContext = new SpanContext();
+        let baggage = {};
+
+        for (let key in carrier) {
+            if (carrier.hasOwnProperty(key)) {
+                if (key === this._contextKey) {
```
yurishkuro: before comparing, convert the key to lower case, since HTTP does not guarantee that the case is preserved

src/propagators/text_map_codec.js
```diff
@@ -56,21 +57,28 @@ export default class TextMapCodec {
...
                     let keyWithoutPrefix = key.substring(this._baggagePrefix.length);
-                    let value = this._decodedValue(carrier[key]);
-                    baggage[keyWithoutPrefix] = value;
+                    baggage[keyWithoutPrefix] = this._decodedValue(carrier[key]);
                 }
             }
-            spanContext.baggage = baggage;
         }
 
+        if (constants.JAEGER_DEBUG_HEADER in carrier) {
```
yurishkuro: similarly, it is safer to have this inside the loop and compare against lowerCase(key)

test/text_map_codec.js
```diff
@@ -0,0 +1,59 @@
...
+import InMemoryReporter from '../src/reporters/in_memory_reporter.js';
+import opentracing from 'opentracing';
+import Tracer from '../src/tracer.js';
+
+describe('Text Map Codec should', () => {
+    it('set debug flag when debug-id-header is received', () => {
+        let tracer = new Tracer(
+            'test-tracer',
+            new InMemoryReporter(),
+            new ConstSampler(true)
```
yurishkuro: should be false, otherwise the check for isSampled() is inconclusive

src/span_context.js
yurishkuro: ok

src/constants.js
yurishkuro: s/JaegerDebugHeader/JAEGER_DEBUG_HEADER

src/propagators/text_map_codec.js
oibe: putting this in the loop means we could set debug, and then later find the tracer header, and overwrite the debug that we set.  This comes from the fact that we initialize span context in case we just want to propagate the debug flag without anything else.

yurishkuro: if you store debugID in a variable defined outside of the loop, similar to baggage, then nothing can override it

src/propagators/text_map_codec.js
```diff
@@ -56,21 +57,34 @@ export default class TextMapCodec {
...
+
+        for (let key in carrier) {
+            if (carrier.hasOwnProperty(key)) {
+                let lowerKey = key.toLowerCase();
+                if (lowerKey === this._contextKey) {
+                    spanContext = SpanContext.fromString(this._decodedValue(carrier[key]));
+                }
+
+                if (lowerKey === constants.JAEGER_DEBUG_HEADER) {
+                    debugId = this._decodedValue(carrier[constants.JAEGER_DEBUG_HEADER]);
```
yurishkuro: this must be `carrier[key]`

src/span_context.js
yurishkuro: I would do `return !this._traceId && this._debugId;`, don't need to check against empty string

PR #22, opened 2016-10-06
Add test utils to abstract away implementation details

oibe:


src/test_util.js
```diff
@@ -0,0 +1,113 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import _ from 'lodash';
+import bufferEqual from 'buffer-equal';
+import Span from './span.js';
+import Utils from './util.js';
+
+export default class TestUtils {
+    static traceIdEqual(span: Span, traceId: number): boolean {
```
yurishkuro: (like)

yurishkuro:
you may also consider methods like `hasTag(span, key, value)` and similar for the logs. Can also have plural versions of those (i.e. input is a subset of span data). Comparing the full set of tags/logs is often problematic because it's a bit implementation dependent (e..g root span might have more tags).

PR #27, opened 2016-10-10
Yaml config loading for Jaeger Tracer

oibe:


src/index.js
```diff
@@ -20,10 +20,185 @@
...
+                'agentHost': {'type': 'string'},
+                'agentPort': {'type': 'number'},
+                'flushIntervalMs': {'type': 'number'}
+            },
+            'additionalProperties': false
+        }
+    }
+};
+
+let getSampler = (options) => {
```
yurishkuro: can we create a Configuration class instead and encapsulate all methods in there?

src/index.js
```diff
@@ -20,10 +20,185 @@
...
+        throwError: true
+    });
+
+    options = options.jaeger;
+
+    let reporter = null;
+    let sampler = null;
+    if (options.disable) {
+        reporter = new NoopReporter();
+        sampler = new ConstSampler(false);
```
yurishkuro: let's log something that tracing is disabled and all other configuration is ignored.

Also, in other libs disable causes a NoopTracer to be used (from opentracing), not Jaeger tracer with null reporter, because they want to disable the whole tracing (including wire serialization). If they only want to disable reporter, we can have another disabled flag under reporter which will turn it to NullReporter.

src/index.js
```diff
@@ -20,10 +20,185 @@
...
+    } else {
+        let sender;
+        let reporterOptions = {};
+        if (options.reporter) {
+            if (options.reporter.logSpans) {
+                // use logging reporter
```
yurishkuro: let's add this now

src/index.js
```diff
@@ -20,10 +20,185 @@
...
+            if (options.reporter.agentHost && options.reporter.agentPort) {
+                let hostPort = `${options.reporter.agentHost}:${options.reporter.agentPort}`;
+                sender = new UDPSender(hostPort);
+            } else {
+                // use default sender
+                sender = new UDPSender();
+            }
+        } else {
+            // use default sender
+            sender = new UDPSender();
```
yurishkuro: DRY - define `hostPort = null` instead of `sender` at the top and create Sender only once

src/index.js
```diff
@@ -20,10 +20,185 @@
...
+        }
+
+
+        reporter = new RemoteReporter(sender, reporterOptions);
```
yurishkuro: we can use the same pattern as with sampler - if `options.reporter` is defined, use it, otherwise create remote. This is helpful when people are testing and *only* want a logging reporter. This came up in Java lib once, where there's no way to override the reporter via config.

src/index.js
```diff
@@ -20,10 +20,185 @@
...
+        } else {
+            sampler = new RemoteSampler(options.serviceName);
+        }
+    }
+
+    return new Tracer(
+        options.serviceName,
+        reporter,
+        sampler
+    );
```
yurishkuro: let's log something about what we're using. You may need to add toString() to all components.

src/index.js
```diff
@@ -20,10 +20,185 @@
...
+        'managerHostPort': {'type': 'string'}
+    },
+    'required': ['type', 'param'],
+    'additionalProperties': false
+};
+
+let jaegerSchema = {
+    'id': '/jaeger',
+    'type': 'object',
+    'required': ['serviceName'],
```
oibe: My problem with this is that all new instrumentation will use "initTracer" and pass in a serviceName that can be incorrect, or fall out of sync from what they have in their yaml file.  So it makes sense for it to be in the config.  The internal repo can still read service name from an env variable, and modify the config it receives.  Basically since most apps determine there own serviceName (which can be wrong) and keep it in yaml files.  I think it makes sense to keep serviceName in the config.

src/index.js
```diff
@@ -20,10 +20,185 @@
...
+        'serviceName': {'type': 'string'},
+        'disable': {'type': 'boolean'},
+        'sampler': {
+            'type': 'object',
+            // should be 'oneOf', but that doesn't work properly with this implementation.
+            'anyOf': [
+                {'$ref': 'const'},
+                {'$ref': 'probabilistic'},
+                {'$ref': 'ratelimiting'},
+                {'$ref': 'remote'}
```
oibe: fixed.

src/index.js
oibe: but options.reporter is a config section not a reporter.  Do you mean if "options.reporter.reporterImpl" is defined?

src/index.js
yurishkuro: Having a brand new entry in the config provides no more guarantee that the service name will be correct vs. if we require them to pass it to initTracer as a literal string. But the latter allows us to build a wrapper in the internal repo that will look up the service name from env variable.

Also, our code should be able to function with a completely empty config, which by default should be completely empty in production, to use the internal tracer defaults. The philosophy of an empty config does not work well if there is a required `serviceName` field in that config.

src/configuration.js
```diff
@@ -0,0 +1,164 @@
...
+import ProbabilisticSampler from './samplers/probabilistic_sampler';
+import RateLimitingSampler from './samplers/ratelimiting_sampler';
+import RemoteReporter from './reporters/remote_reporter';
+import CompositeReporter from './reporters/composite_reporter';
+import LoggingReporter from './reporters/logging_reporter';
+import RemoteSampler from './samplers/remote_sampler';
+import Tracer from './tracer';
+import UDPSender from './reporters/udp_sender';
+import {Validator} from 'jsonschema';
+import Logtron from 'logtron';
```
oibe: This is a default logger, for our uses.  So the user can still pass in a logger that supports "info", "error", "debug", "warn".  I'll move it to prod dependencies though.

yurishkuro: that is my point - we do not want prod dependency on logtron, in case other people use other loggers.

src/configuration.js
```diff
@@ -101,7 +100,7 @@ export default class Configuration {
         return sampler;
     }
 
-    static initTracer(options) {
+    static initTracer(options, logger) {
```
yurishkuro: logger should be in the `options` imo, defaulted to NullLogger

src/configuration.js
```diff
@@ -0,0 +1,158 @@
...
+
+export default class Configuration {
+
+    static _getSampler(config) {
+        let type = config.sampler.type;
+        let param = config.sampler.param;
+
+        let sampler;
+        if (type === 'probabilistic') {
+        sampler = new ProbabilisticSampler(param);
```
yurishkuro: formatting

PR #29, opened 2016-10-11
Add metrics support

oibe:


src/_interfaces/timer.js
```diff
@@ -0,0 +1,24 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+declare interface Timer {
+    timing(timestamp: number): void;
```
yurishkuro: `s/timing(timestamp)/record(value)/`

src/metrics/metrics.js
```diff
@@ -0,0 +1,35 @@
...
+
+export default class Metrics {
+    _factory: MetricFactory;
+
+    constructor(factory: MetricFactory) {
+        this._factory = factory;
+    }
+
+    get tracesStartedSampled = () => {
+        return this._factory.createCounter('traces', {
```
yurishkuro: this will call the factory on every call to the property. You need to init the respective variables in the constructor.

yurishkuro:
overall on the right path

src/_interfaces/timer.js
yurishkuro: This is our own API, independent of other libraries / systems. And in our API (e.g. in Go) it's called Record

```go
// Timer tracks how long an event has occurred.
type Timer interface {
	// Records the time passed in.
	Record(time.Duration)
}
```

btw, please add comments to all interfaces similar to https://github.com/uber/jaeger/tree/master/pkg/metrics

src/reporters/remote_reporter.js
```diff
@@ -39,19 +42,24 @@ export default class RemoteReporter {
...
         let response: SenderResponse = this._sender.append(ThriftUtils.spanToThrift(span));
         if (response.err) {
             this._logger.error('Failed to append spans in reporter.');
+            this._metrics.reporterDropped.increment(1);
         }
     }
 
     flush(callback: ?Function): void {
         let response: SenderResponse = this._sender.flush(callback);
         if (response.err) {
```
yurishkuro: this should be a string, not a boolean - filed #32

src/samplers/remote_sampler.js
```diff
@@ -72,6 +76,10 @@ export default class RemoteControlledSampler {
         request.get(`http:\/\/${this._host}:${this._port}/?service=${encodedCaller}`, (err, response) => {
```
yurishkuro: why do you need to escape // ?  The slash after the port is not escaped.

src/samplers/remote_sampler.js
```diff
@@ -93,11 +101,14 @@ export default class RemoteControlledSampler {
             let maxTracesPerSecond = strategy.rateLimitingSampling.maxTracesPerSecond;
             newSampler = new RateLimitingSampler(maxTracesPerSecond);
         } else {
+            this._metrics.samplerParsingFailure.increment(1);
             this._logger.error('Unrecognized strategy type: ' + JSON.stringify({error: strategy}));
```
yurishkuro: `JSON.stringify({error: strategy})` - should our logging API be key-value based instead, so that we don't need to stringify fields?  I thought that's what logtron supports.

src/test_util.js
```diff
@@ -156,4 +157,10 @@ export default class TestUtils {
 
         return true;
     }
+
+    static counterEquals(counter: LocalCounter, value: number): boolean {
+        let valueEqual = counter._backend._counterValues[counter._name] === value;
```
yurishkuro: I don't think this should be in the public utils. The point of local backend is that you can query it for the current state of counters (which would return some well defined dict-based structure), rather than peeking into implementation details of the counters.

src/tracer.js
```diff
@@ -40,28 +42,30 @@ export default class Tracer {
...
         this._serviceName = serviceName;
         this._reporter = reporter;
         this._sampler = sampler;
-        this._logger = logger || new NullLogger();
+        this._logger = options.logger || new NullLogger();
         this._injectors = {};
         this._extractors = {};
 
-        let textCodec = new TextMapCodec(false);
+        let textCodec = new TextMapCodec(false, { metrics: this._metrics });
```
yurishkuro: while at it, let's move the first argument into the options too.

src/tracer.js
```diff
@@ -95,10 +99,30 @@ export default class Tracer {
 
         span.addTags(tags);
         span.addTags(internalTags);
+
+        // emit metrics
+        this._metrics.spansStarted.increment(1);
+        if (span.context().isSampled()) {
+            this._metrics.spansSampled.increment(1);
+            if (rpcServer) {
```
yurishkuro: s/rpcServer/not firstInProcess/

src/tracer.js
```diff
@@ -95,10 +99,30 @@ export default class Tracer {
...
+                this._metrics.tracesJoinedSampled.increment(1);
+            } else if (!span.context().parentId) {
+                this._metrics.tracesStartedSampled.increment(1);
+            }
+        } else {
+            this._metrics.spansNotSampled.increment(1);
+            if (rpcServer) {
```
yurishkuro: s/rpcServer/not firstInProcess/

src/tracer.js
yurishkuro: this is a logical error, if your unit tests are not catching it, they are incomplete.

test/text_map_codec.js
```diff
@@ -24,8 +24,29 @@ import ConstSampler from '../src/samplers/const_sampler.js';
 import InMemoryReporter from '../src/reporters/in_memory_reporter.js';
 import opentracing from 'opentracing';
 import Tracer from '../src/tracer.js';
+import MetricsContainer from '../src/metrics/metrics.js';
+import LocalMetricFactory from '../src/metrics/local/metric_factory.js';
+import TestUtils from '../src/test_util.js';
 
 describe('Text Map Codec should', () => {
+    it ('report metric when failing to decode tracer stawte', () => {
```
yurishkuro: typo

yurishkuro: don't we already have a unit test for failure to extract context? I'd rather add metrics check there than repeat the boilerplate

test/text_map_codec.js
```diff
@@ -24,8 +24,29 @@ import ConstSampler from '../src/samplers/const_sampler.js';
...
+            new InMemoryReporter(),
+            new ConstSampler(false), {
+                metrics: metrics
+            }
+        );
+
+        let headers = {
+            'uber-trace-id': 'bad-value'
+        };
+        tracer.extract(opentracing.FORMAT_HTTP_HEADERS, headers);
```
yurishkuro: what do we expect as a return value here?

test/remote_reporter.js
```diff
@@ -25,17 +25,25 @@ import MockLogger from './lib/mock_logger.js';
 import RemoteReporter from '../src/reporters/remote_reporter.js';
 import Tracer from '../src/tracer.js';
 import UDPSender from '../src/reporters/udp_sender.js';
+import MetricsContainer from '../src/metrics/metrics.js';
```
yurishkuro: where is the "container" name coming from?  If it's simply importing the Metrics class, I think it's less confusing to call it Metrics.

src/samplers/remote_sampler.js
oibe: logtron supports strings, and even if it did support key value pairs the clients don't have to use logtron, and most other loggers support strings.

test/text_map_codec.js
oibe: not one that has a faulty "uber-trace-id"

test/text_map_codec.js
oibe: null

test/remote_reporter.js
oibe: I did this to avoid name clashing with the "metrics" variable.

src/tracer.js
oibe: I don't think that makes sense since the first parameter is required.  Generally only optional arguments get put into a json object.

test/remote_reporter.js
yurishkuro: there's no clash between `Metrics` and  `metrics`

src/metrics/local/backend.js
```diff
@@ -31,6 +34,12 @@ export default class LocalBackend {
         this.reset();
     }
 
+    static counterEquals(counter: LocalCounter, value: number): boolean {
```
yurishkuro: just define this in test file somewhere

src/tracer.js
```diff
@@ -104,14 +104,14 @@ export default class Tracer {
         this._metrics.spansStarted.increment(1);
         if (span.context().isSampled()) {
             this._metrics.spansSampled.increment(1);
-            if (rpcServer) {
+            if (!firstInProcess) {
```
yurishkuro: wait, this is still wrong. If think we want
 * if parentId is not null
 ** trace started
 * else if rpcServer
 ** trace joined

firstInProcess can be both new and join

src/tracer.js
```diff
@@ -104,14 +104,14 @@ export default class Tracer {
...
                 this._metrics.tracesJoinedSampled.increment(1);
             } else if (!span.context().parentId) {
```
yurishkuro: change to hasParent which should return `this.parentId != null || this.parentIdStr != null`. Otherwise you may unnecessarily trigger Buffer construction

src/tracer.js
yurishkuro: I don't think the first parameter is required either. It's really an option that enables additional behavior of url-encoding, but the codec will work just fine without that option.

src/tracer.js
```diff
@@ -95,10 +100,30 @@ export default class Tracer {
...
+        if (span.context().isSampled()) {
+            this._metrics.spansSampled.increment(1);
+            if (!hadParent) {
+                this._metrics.tracesStartedSampled.increment(1);
+            } else if (rpcServer) {
+                this._metrics.tracesJoinedSampled.increment(1);
+            }
+        } else {
+            this._metrics.spansNotSampled.increment(1);
+            if (!hadParent) {
```
yurishkuro: the `parent` in L179 is not exactly the right condition, because with jaeger-debug-id you would still have a parent context, but it's not a real parent. You may notice how in Go I have a separate variable which is only set to true if the parent context is "valid"

yurishkuro: you can add a check for emitted metrics in the unit test where you test jaeger-debug-id header - the span created in that test should emit "trace started" metrics, but I think your code will emit "trace joined".

test/text_map_codec.js
yurishkuro: so let's assert that

test/tracer.js
```diff
@@ -249,4 +252,47 @@ describe('tracer should', () => {
...
     });
+
+    describe('Metrics', () => {
+        it ('startInternalSpan', () => {
+            let params = [
+                { 'hasParent': false, 'context': '1:2:0:1', 'sampled': true, 'metrics': ['spansStarted', 'spansSampled', 'tracesStartedSampled']},
+                { 'hasParent': true, 'context': '1:2:100:1', 'sampled': true, 'metrics': ['spansStarted', 'spansSampled', 'tracesJoinedSampled']},
+                { 'hasParent': false, 'context': '1:2:0:0', 'sampled': false, 'metrics': ['spansStarted', 'spansNotSampled', 'tracesStartedNotSampled']},
+                { 'hasParent': true, 'context': '1:2:100:0', 'sampled': false, 'metrics': ['spansStarted', 'spansNotSampled', 'tracesJoinedNotSampled']},
+            ];
```
yurishkuro: this is the problem with writing unit tests against internal methods - you fake the arguments to satisfy the unit test, but you're not checking what really happens when the public API is used. Can you augment this test to actually use the public `startSpan()`?

Issue #23, opened 2016-10-10
README is wrong


yurishkuro:
Instructions in the readme are broken.
- submodule does not init
- tests don't work
- I believe we need specific versions of Node and npm

```
$ cd jaeger-client-node
$ git submodule init update
error: pathspec 'update' did not match any file(s) known to git.
```

so fix it by 

```
$ git submodule init
$ git submodule update

# now try running tests
$ npm test
...
sh: flow: command not found
sh: eslint: command not found
...
npm WARN Local package.json exists, but node_modules missing, did you mean to install?
```

frankgreco:
I also experienced this. My workout was running all commands as root:

```sh
$ sudo su
# ...
```

yurishkuro:
I think this was resolved in 7804b0ef6c8c0cec15e8238292d954718438b09f

PR #30, opened 2016-10-12
Use lazy dual string/Buffer representation for trace IDs

yurishkuro:
According to the basic server-client loop `benchmarks/flame_graph_loop.js` flamegraph, to/from strings ops on the span context are among the most expensive portions.  This change introduces a dual representation of trace IDs as `string` and `Buffer`. Whichever is available first is stored, and the other is only generated as needed. For most (unsampled) spans the `fromString` operation produces string IDs, and the outbound `toString` operation can use them as is, without even converting to `Buffer` representation, which is more expensive.

## To/FromString Benchmarks

The micro-benchmarks below show almost 10 times improvement for to/from string methods.

### Before benchmark

```
$ node ./benchmarks/span_context.js

Beginning Span Context Benchmark...
  2 tests completed.

  SpanContext:fromString x   274,035 ops/sec ±1.03% (90 runs sampled)
  SpanContext:toString   x 1,649,536 ops/sec ±1.01% (93 runs sampled)
```

### After benchmark

```
$ node benchmarks/span_context.js

Beginning Span Context Benchmark...
  2 tests completed.

  SpanContext:fromString x  2,463,840 ops/sec ±1.33% (91 runs sampled)
  SpanContext:toString   x 12,618,966 ops/sec ±1.21% (94 runs sampled)
```

## Core Flows Benchmarks

The `core_flow` benchmark shows moderate to no improvement

### Before benchmark

$ node benchmarks/core_flows.js

Beginning Main Request Flows...
  Fundamental Request flow with const tracer and http carrier x 41,873 ops/sec ±5.00% (73 runs sampled)
  Fundamental Request flow with const tracer and text carrier x 60,004 ops/sec ±10.43% (76 runs sampled)
  Fundamental Request flow with probabilistic tracer and http carrier x 42,920 ops/sec ±12.71% (76 runs sampled)
  Fundamental Request flow with probabilistic tracer and text carrier x 51,835 ops/sec ±28.55% (74 runs sampled)

### After benchmark

$ node benchmarks/core_flows.js

Beginning Main Request Flows...
  Fundamental Request flow with const tracer and http carrier x 47,763 ops/sec ±4.87% (75 runs sampled)
  Fundamental Request flow with const tracer and text carrier x 64,508 ops/sec ±11.43% (72 runs sampled)
  Fundamental Request flow with probabilistic tracer and http carrier x 42,601 ops/sec ±21.58% (66 runs sampled)
  Fundamental Request flow with probabilistic tracer and text carrier x 52,418 ops/sec ±33.48% (54 runs sampled)

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,7 @@
 -include crossdock/rules.mk
 
 .PHONY: publish
-publish:
-	npm run compile
+publish: build-node
 	npm version $(shell ./scripts/version_prompt.sh)
 	git push  --set-upstream origin $(shell bash -c "git rev-parse --abbrev-ref HEAD")
 	git push origin --tags
--- a/README.md
+++ b/README.md
@@ -10,9 +10,10 @@ with Zipkin-compatible data model.
 
 ## Developing
 
- 1. `git submodule init update`
- 2. `npm test`
- 3. `make build-node`
+ 1. `git submodule update --init`
+ 2. `npm install`
+ 3. `npm test`
+ 4. `make build-node`
 
   [ci-img]: https://travis-ci.org/uber/jaeger-client-node.svg?branch=master
   [cov-img]: https://coveralls.io/repos/github/uber/jaeger-client-node/badge.svg?branch=master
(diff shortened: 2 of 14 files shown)


yurishkuro:
Also cleans up README, fixes #23

PR #37, opened 2016-10-14
fix imports from jaeger-client

oibe:


oibe:
Work in progress. No testing yet.  Is this what we want to export?

yurishkuro:
what's the matter with code coverage dropping by 30%?

PR #38, opened 2016-10-17
Add support for span references

oibe:


src/thrift.js
```diff
@@ -90,6 +91,30 @@ export default class ThriftUtils {
...
+        for (let i = 0; i < refs.length; i++) {
+            let refEnum;
+            let ref = refs[i];
+            let context = refs[i].referencedContext();
+
+            if (ref.type() === opentracing.REFERENCE_CHILD_OF) {
+                refEnum = ThriftUtils._thrift.SpanRefType.CHILD_OF;
+            } else if (ref.type() === opentracing.REFERENCE_FOLLOWS_FROM) {
+                refEnum = ThriftUtils._thrift.SpanRefType.FOLLOWS_FROM
+            }
```
yurishkuro: else continue

src/tracer.js
```diff
@@ -177,21 +178,21 @@ export default class Tracer {
...
-        let references = fields.references || [];
         let startTime = fields.startTime;
         if (!startTime) {
             startTime = Utils.getTimestampMicros();
         }
 
-        // TODO(oibe) support use of references
         let parent: ?SpanContext = fields.childOf instanceof Span ? fields.childOf.context(): fields.childOf;
-        if (!parent) {
+        if (!parent && fields.references) {
```
yurishkuro: s/fields/options/ to be consistent with other functions

src/tracer.js
```diff
@@ -177,21 +178,21 @@ export default class Tracer {
...
             // If there is no childOf in fields, then search list of references
-            for (let i = 0; i < references.length; i++) {
-                let ref: Reference = references[i];
+            for (let i = 0; i < fields.references.length; i++) {
+                let ref: Reference = fields.references[i];
                 let ref_type = ref.type();
                 if ((ref_type === opentracing.REFERENCE_CHILD_OF) ||
                     (ref_type === opentracing.REFERENCE_FOLLOWS_FROM)) {
+                    // TODO(oibe) which refference to use as parent when we have a list of
+                    // references? Should a new parent be created as when we receive no 'childOf'?
```
yurishkuro: strictly speaking references are a replacement for the notion of the parent, but we decided to keep the parent as the most frequently used type of reference. It should only be created from the CHILD_OF type of reference.

yurishkuro: NB: we do not need to keep the child_of reference in the span since we already represent it with the `parent` field.

NB: we should also guard against child_of reference that has a different trace ID - that would indicate a mistake in the instrumentation. I think we can convert such reference to FOLLOWS_FROM instead

test/udp_sender.js
```diff
@@ -92,6 +93,31 @@ describe('udp sender should', () => {
...
+            ]
+        });
+        spanOne.finish(); // finish to set span duration
+        spanOne = ThriftUtils.spanToThrift(spanOne);
+
+        server.on('message', (msg, remote) => {
+            let thriftObj = thrift.Agent.emitBatch.argumentsMessageRW.readFrom(msg, 0);
+            let batch = thriftObj.value.body.batch;
+            assert.isOk(batch);
+            assert.isOk(TestUtils.thriftSpanEqual(spanOne, batch.spans[0]));
```
yurishkuro: this is not actually testing that the refs were serialized, e.g. if there's a bug in startSpan that ignores them then the spans will still be equal

src/tracer.js
yurishkuro: bug: if there's no parent then you don't process references at all

src/tracer.js
```diff
@@ -177,21 +178,21 @@ export default class Tracer {
...
 
         let tags = fields.tags || {};
-        let references = fields.references || [];
         let startTime = fields.startTime;
         if (!startTime) {
             startTime = Utils.getTimestampMicros();
         }
 
-        // TODO(oibe) support use of references
         let parent: ?SpanContext = fields.childOf instanceof Span ? fields.childOf.context(): fields.childOf;
```
yurishkuro: need to normalize this with child_of being in the `references`.

src/tracer.js
oibe: If there's no parent only then do we look at span references.

yurishkuro: it's a bug

src/thrift.js
```diff
@@ -94,14 +94,20 @@ export default class ThriftUtils {
     static spanRefsToThriftRefs(refs: Array<Reference>): Array<any> {
         let thriftRefs = [];
         for (let i = 0; i < refs.length; i++) {
-            let refEnum;
             let ref = refs[i];
+            if (!ref) {
```
yurishkuro: why would this ever be the case?  These are refs from the span, we should prevent adding nulls to that array.

test/udp_sender.js
```diff
@@ -108,6 +110,12 @@ describe('udp sender should', () => {
         server.on('message', (msg, remote) => {
             let thriftObj = thrift.Agent.emitBatch.argumentsMessageRW.readFrom(msg, 0);
             let batch = thriftObj.value.body.batch;
+            let span = batch.spans[0];
+            let ref = span.references[0];
```
yurishkuro: assert len(span.references) == 1, not 2

test/udp_sender.js
```diff
@@ -108,6 +110,12 @@ describe('udp sender should', () => {
...
+
+            assert.isOk(bufferEqual(span.traceIdLow, ref.traceIdLow));
```
yurishkuro: unclear what the business meaning of this check is. I'd rather expect a check that the span's traceID/parentID came from childOf reference.

test/udp_sender.js
```diff
@@ -94,12 +95,13 @@ describe('udp sender should', () => {
...
-        let context2 = tracer.startSpan('just-used-for-context').context();
+        let childOfContext = tracer.startSpan('just-used-for-context').context();
+        let followsFromContext = tracer.startSpan('just-used-for-context').context();
+        followsFromContext.traceId = childOfContext.traceId;
         let spanOne = tracer.startSpan('operation-one', {
             references: [
-                new opentracing.Reference(opentracing.REFERENCE_CHILD_OF, context),
-                new opentracing.Reference(opentracing.REFERENCE_FOLLOWS_FROM, context2)
+                new opentracing.Reference(opentracing.REFERENCE_CHILD_OF, childOfContext),
+                new opentracing.Reference(opentracing.REFERENCE_FOLLOWS_FROM, followsFromContext)
```
yurishkuro: I suggest a table-driven test for various permutations, e.g.
 * childOf=null, refs=[]
 * childOf + refs=[]
 * childOf + refs=[follows]
 * childOf + refs=[childOf, follows]
 * childOf=null + refs=[childOf]
 * childOf=null + refs=[follows]  // this should still inherit parentID from the follows ref
 * childOf=null + refs=[follows, childOf]

test/udp_sender.js
```diff
@@ -95,35 +96,51 @@ describe('udp sender should', () => {
...
+        let parentContext = tracer.startSpan('just-used-for-context').context();
+        let childOfRef = new opentracing.Reference(opentracing.REFERENCE_CHILD_OF, tracer.startSpan('just-used-for-context').context());
+        let followsFromRef = new opentracing.Reference(tracer.startSpan('just-used-for-context').context());
+
+        var options = [
+            { 'childOf': null, 'references': [], 'expectedTraceId': null, 'expectedParentId': null },
+            { 'childOf': parentContext, 'references': [], 'expectedTraceId': parentContext.traceId, 'expectedParentId': parentContext.parentId },
+            { 'childOf': parentContext, 'references': [followsFromRef], 'expectedTraceId': parentContext.traceId, 'expectedParentId': parentContext.parentId },
+            { 'childOf': parentContext, 'references': [childOfRef, followsFromRef], 'expectedTraceId': parentContext.traceId, 'expectedParentId': parentContext.parentId},
+            { 'childOf': null, 'references': [childOfRef], 'expectedTraceId': childOfRef.traceId, 'expectedParentId': childOfRef.parentId },
```
yurishkuro: `childOfRef.parentId` - references do not have parentID

test/udp_sender.js
```diff
@@ -95,35 +96,51 @@ describe('udp sender should', () => {
...
-        let childOfContext = tracer.startSpan('just-used-for-context').context();
-        let followsFromContext = tracer.startSpan('just-used-for-context').context();
-        followsFromContext.traceId = childOfContext.traceId;
-        let spanOne = tracer.startSpan('operation-one', {
-            references: [
-                new opentracing.Reference(opentracing.REFERENCE_CHILD_OF, childOfContext),
-                new opentracing.Reference(opentracing.REFERENCE_FOLLOWS_FROM, followsFromContext)
-            ]
+        let parentContext = tracer.startSpan('just-used-for-context').context();
+        let childOfRef = new opentracing.Reference(opentracing.REFERENCE_CHILD_OF, tracer.startSpan('just-used-for-context').context());
```
yurishkuro: why start a new span? I would just create a reference from `parentContext`

test/udp_sender.js
```diff
@@ -95,35 +96,51 @@ describe('udp sender should', () => {
...
+                let thriftObj = thrift.Agent.emitBatch.argumentsMessageRW.readFrom(msg, 0);
+                let batch = thriftObj.value.body.batch;
+                let span = batch.spans[0];
+                let ref = span.references[0];
+
+                assert.isOk(batch);
+                assert.isOk(TestUtils.thriftSpanEqual(span, batch.spans[0]));
+                if (o.expectedTraceId) {
+                    assert.isOk(bufferEqual(span.traceIdLow, o.expectedTraceId));
+                }
```
yurishkuro: else check that it's not defined

test/udp_sender.js
```diff
@@ -95,35 +96,51 @@ describe('udp sender should', () => {
...
+
+                if (o.expectedparentId) {
+                    assert.isOk(bufferEqual(span.parentId, o.expectedParentId));
```
yurishkuro: else check that parent is not defined

src/tracer.js
```diff
@@ -185,6 +185,11 @@ export default class Tracer {
         }
 
         let parent: ?SpanContext = options.childOf instanceof Span ? options.childOf.context(): options.childOf;
+
+        if (!parent && options.references.length > 0) {
+            parent = options.references[0].referencedContext();
+        }
```
yurishkuro: this is already (mostly) taken care of by L198, and doing this unconditionally here is wrong anyway, because the ref type can be something that should not become a parent.

test/udp_sender.js
```diff
@@ -95,51 +95,63 @@ describe('udp sender should', () => {
...
+    describe('span reference tests', () => {
+        let tracer = new Tracer(
+            'test-service-name',
+            new InMemoryReporter(),
+            new ConstSampler(true)
+        );
         let parentContext = tracer.startSpan('just-used-for-context').context();
-        let childOfRef = new opentracing.Reference(opentracing.REFERENCE_CHILD_OF, tracer.startSpan('just-used-for-context').context());
-        let followsFromRef = new opentracing.Reference(tracer.startSpan('just-used-for-context').context());
+        let childOfContext = tracer.startSpan('just-used-for-context').context();
```
yurishkuro: I don't see why you need a new span for childOfContext instead of just using `parentContext`, but whatever

PR #44, opened 2016-10-25
Add source maps for debugging

oibe:


.vscode/launch.json
```diff
@@ -0,0 +1,39 @@
...
+            "type": "node",
+            "request": "launch",
+            "program": "${workspaceRoot}/node_modules/.bin/_mocha",
+            "stopOnEntry": false,
+            // change to the name of the generated sourcemapped file in order
+            // to debug that file.
+            "args": ["-t", "10000000", "--debug-brk", "./dist/crossdock/test/tchannel_server.js"],
+            "cwd": "${workspaceRoot}",
+            "preLaunchTask": null,
+            "runtimeExecutable": "/Users/onwukikeibe/.nvm/v0.10.32/bin/node",
```
black-adder: not sure what this file is for, is this hardcoded line needed?

package.json
```diff
@@ -21,6 +21,7 @@
     "js-yaml": "^3.6.1",
     "jsonschema": "^1.1.0",
     "request": "^2.74.0",
+    "tchannel": "^3.9.0",
```
yurishkuro: wow... not the runtime. Can it be in test deps?

oibe: we need it for the bridge...

.vscode/launch.json
oibe: good point.  Overall this file is used as the debuging setup for vscode.

package.json
oibe: we could vendor the tchannel/trace/span class

oibe:
removed tchannel from this diff.

PR #48, opened 2016-10-27
Crossdock tchannel

oibe:


crossdock/src/helpers.js
```diff
@@ -20,10 +20,188 @@
...
+
+                let port = parseInt(downstream.port);
+                let downstreamUrl = `http:\/\/${address}:${port}/join_trace`;
+
+                let clientSpan = this._tracer.startSpan('client-span', { childOf: context.span.context() });
+
+                // Save client span to tchannel headers
+                let headers = { 'Content-Type': 'application/json' };
+                TChannelBridge.saveTracerStateToCarrier(clientSpan, headers);
+                TChannelBridge.saveBaggageToCarrier(clientSpan, headers);
```
yurishkuro: should be just one function

crossdock/src/helpers.js
```diff
@@ -20,10 +20,188 @@
...
+
+                let request = this._thriftChannel.request({
+                    timeout: 2000,
+                    parent: { span: TChannelBridge.getTChannelParentSpan() },
+                    headers: {
+                        cn: 'tcollector-requestor'
+                    },
+                    trace: true,
```
yurishkuro: ?

crossdock/src/tchannel_server.js
```diff
@@ -46,17 +57,24 @@ export default class TChannelServer {
     }
 
     handleTChannelRequest(context: any, req: any, head: any, body: any, callback: Function) {
-        callback(null, {
+        let startRequest: boolean = false;
+        let traceRequest = body.request;
+        let spanContext = TChannelBridge.getSpanFromTChannelRequest(this._tracer, 'tchannel-server-span', head).context();
```
yurishkuro: Two problems here
  * the span is, apparently, never finished, thus is a bad example of using the API
  * I would prefer if we defined an API where (a) the user did not have to deal with the span manually, and (b) the user **had** to accept with some "context" object instead

Rather than making the user call getSpanFromTChannelRequest, I think we can create a decorator around the whole request handler (which incidentally already takes a context). Specifically, in L52 you're registering   `handleTChannelRequest` as a handler with tchannel. Instead you should be registering

`TracedHander(handleTChannelRequest)`

where TracedHander would take care of the span etc., and then delegate to the actual handler.

crossdock/src/helpers.js
oibe: This forces trace, so that in tests we can look at tcap, and verify that a traceid of 0 is being sent.  (Not sure how to test that condition in a unit test, since I don't think I have access to the outgoing frame).

yurishkuro: please document in comments, it's important

for unit test, you should be able to get access to tchannel span in the `context` in the server handler and verify its trace ID is 0

Issue #40, opened 2016-10-21
Remove dependency on yaml and jsonschema


yurishkuro:
As we discussed, we can leave the schema definition as a public constant in this project, should someone wants to use it, but we should not force these extra dependencies on users who may not need them.

yurishkuro:
Fixed by #95

PR #51, opened 2016-11-07
Removed yaml parsing of config to initialize tracer, along with json

oibe:
schema validation.


oibe:
fixes https://github.com/uber/jaeger-client-node/issues/40

src/configuration.js
```diff
@@ -31,47 +31,8 @@ import LoggingReporter from './reporters/logging_reporter';
...
-    'required': ['jaeger'],
-    'properties': {
-        'jaeger': {
-            'type': 'object',
-            '$ref': '/jaeger'
-        }
-    },
-};
-
-let jaegerSchema = {
```
yurishkuro: why remove this?  My proposal in another review was to keep the schema definition, just not do the validation (only do it in tests).

oibe: what's the point of doing in tests, but not in prod?

yurishkuro: (a) it provides an authoritative schema for the config (not just a file, but the actual data structure used in the code)
(b) it lets users do validation if they choose to

PR #52, opened 2016-11-07
Boost test coverage for reporters

oibe:


src/reporters/composite_reporter.js
```diff
@@ -52,5 +59,12 @@ export default class CompositeReporter {
         }
```
yurishkuro: this is incorrect. Each call to nested reporter `close()` is async, so the callback will be called almost immediately without waiting for the nested reporters to close. We need to implement a simple wait group, e.g. (assuming `count` is captured by reference):
```
let count = 0;
cb = function() {
  count++;
  if (count >= this._reporters.length) {
    if (callback) {
      callback(); // call the real callback
    }
  }
};
for (let i = 0; i < this._reporters.length; i++) {
             let reporter = this._reporters[i];
             reporter.close(cb);
}

```

test/all_reporters.js
```diff
@@ -0,0 +1,94 @@
...
+            let closeCallback = () => { closeCalled = true; };
+
+            // These next few lines exist for the sake of
+            // code coverage. The actual code does nothing.
+            let spanMock = {};
+            reporter.report(spanMock);
+            let tagsMock = [];
+            reporter.setProcess('some-service-name', tagsMock);
+
+            // branch coverage
```
yurishkuro: rm - it's just coverage

test/remote_reporter.js
```diff
@@ -63,7 +66,7 @@ describe('Remote Reporter should', () => {
         span.finish();
         assert.equal(sender._batch.spans.length, 1);
 
-        reporter.flush()
+        reporter.flush();
```
yurishkuro: don't follow this - how did it compile/run before?

src/reporters/composite_reporter.js
oibe: This can be done with promises.  That is the convention.

src/reporters/composite_reporter.js
```diff
@@ -33,24 +33,50 @@ export default class CompositeReporter {
...
-        }
+        let count = 0;
+        let cb = () => {
+            count++;
+            if (count >= this._reporters.length) {
+                if (callback) {
+                    callback();
+                }
+            }
+        };
```
yurishkuro: DRY

src/reporters/composite_reporter.js
```diff
@@ -33,24 +33,43 @@ export default class CompositeReporter {
...
+    }
+
+    flush(callback: ?Function): void {
+        this._reporters.map((r) => {
+            r.flush(this._executeCallbackLast(callback));
```
yurishkuro: there is a bug here, and your unit tests are not catching it. I'll comment on the test.

test/all_reporters.js
```diff
@@ -0,0 +1,95 @@
...
+
+        sender = new UDPSender();
+        sender.setProcess(inMemoryReporter._process);
+        remoteReporter._sender = sender;
+
+        // covered without callbacks
+        reporter.flush();
+        reporter.close(() => {
+            assert.isOk(closeCalled);
+            assert.isOk(flushCalled);
```
yurishkuro: why are asserts inside a callback? If you have a bug and the callback is not called, your test succeeds.

test/all_reporters.js
```diff
@@ -0,0 +1,95 @@
...
+            inMemoryReporter,
+            noopReporter,
+            remoteReporter
+        ];
+
+        let reporter = new CompositeReporter(reporters);
+        let closeCalled = false;
+        let flushCalled = false;
+        let closeCallback = () => { closeCalled = true; };
+        let flushCallback = () => { flushCalled = true; };
```
yurishkuro: change these to numbers, increment in the callback, and assert that the value is 1

test/all_reporters.js
```diff
@@ -0,0 +1,111 @@
...
+
+        sender = new UDPSender();
+        sender.setProcess(inMemoryReporter._process);
+        remoteReporter._sender = sender;
+
+        // covered without callbacks
+        reporter.flush();
+        reporter.close();
+
+        RSVP.all(promises).then((callbacks) => {
```
yurishkuro: this is still wrong. Your `promises` array can be empty here, because it's only populated in the callbacks. Simplify this test, testing one thing only.
```
let reporters = makeAllReporters();
let cReporter = new CompositeReporter(reporters);
let calls = 0;
cReporter.close(() => {
    calls++;
    assert.equal(calls, 1);
    done();
});
```

I am not 100% certain, but above test should catch the callback not being called, AND the callback being called more than once. You need to **verify** that by changing the code so that the test fails for both scenarios. I am not 100% certain about the latter, calling more than once, because if that happens the assertion would certainly fail, but the `done()` would've been called by then, so not sure how the framework would react to that (in fact it could happen after all tests have finished). 

Another option is to use http://sinonjs.org/. The validation in the test then becomes
```
var callback = sinon.spy();
cReporter.close(callback);
assert(callback.calledOnce);
```

Again, fudge the reporter code to prove that the test WILL fail if the callback is not called or called twice.

src/reporters/composite_reporter.js
```diff
@@ -33,24 +33,30 @@ export default class CompositeReporter {
...
-            reporter.report(span);
-        }
+        this._reporters.forEach((r) => {
+            r.report(span);
+        });
+    }
+
+    flush(callback: ?Function): void {
+        this._reporters.forEach((r) => {
+            r.flush(callback);
```
yurishkuro: this is calling callback N times, instead of once on completion. Where did the waitgroup logic go?

oibe: I thought we didn't need that.  If you want to assert how may times something is called then sinon has that.

src/reporters/remote_reporter.js
```diff
@@ -65,6 +65,10 @@ export default class RemoteReporter {
         } else {
             this._metrics.reporterSuccess.increment(response.numSpans);
         }
+
+        if (callback) {
```
yurishkuro: you already called the callback from the sender

src/reporters/remote_reporter.js
```diff
@@ -65,6 +65,10 @@ export default class RemoteReporter {
...
+            callback();
+        }
```
yurishkuro: L76 - clear the timer first, then call sender.close()

src/reporters/udp_sender.js
```diff
@@ -132,10 +132,6 @@ export default class UDPSender {
         this._client.send(thriftBuffer, 0, thriftBuffer.length, PORT, HOST);
         this._reset();
 
-        if (testCallback) {
```
yurishkuro: This is correct - this method is effectively synchronous, UNLESS there is a way to pass a callback to `this._client.send`. If there is not, then there should be no callbacks in this method. So remove it from the arguments too.

yurishkuro: The close() method in L160 is broken semantically. It first needs to call flush

src/reporters/remote_reporter.js
yurishkuro: Also, before calling sender.close() we need to call sender.flush(). And then I would reconsider having the callback in sender.close(), because it becomes synch method.

src/reporters/composite_reporter.js
```diff
@@ -33,24 +33,39 @@ export default class CompositeReporter {
...
-        for (let i = 0; i < this._reporters.length; i++) {
-            let reporter = this._reporters[i];
-            reporter.report(span);
+        this._reporters.forEach((r) => {
+            r.report(span);
+        });
+    }
+
+    compositeCallback(callback: ?Function): ?Function {
+        if (callback) {
```
yurishkuro: no if here, it's checked inside the loop, which is the right place.

src/reporters/remote_reporter.js
```diff
@@ -65,11 +65,20 @@ export default class RemoteReporter {
         } else {
             this._metrics.reporterSuccess.increment(response.numSpans);
         }
+
```
yurishkuro: L61 - do not pass callback

test/all_reporters.js
```diff
@@ -0,0 +1,89 @@
...
+            loggingReporter,
+            inMemoryReporter,
+            noopReporter,
+            remoteReporter
+        ];
+        let reporter = new CompositeReporter(reporters);
+
+        let closeCallback = sinon.spy();
+
+        reporter.clear();
```
yurishkuro: what is clear?

test/all_reporters.js
```diff
@@ -0,0 +1,89 @@
...
+        reporter.close(closeCallback, reporters.length);
+
+        sender = new UDPSender();
+        sender.setProcess(inMemoryReporter._process);
+        remoteReporter._sender = sender;
```
yurishkuro: what is the point of these 3 lines?

test/all_reporters.js
```diff
@@ -0,0 +1,89 @@
...
+
+        // covered without callbacks
+        reporter.close();
+
+        assert.isOk(closeCallback.calledOnce);
+    });
+
+    describe('Logging reporter', () => {
+        it('report span, and do nothing on clear and close', () => {
```
yurishkuro: bad description

test/remote_reporter.js
```diff
@@ -40,10 +42,10 @@ describe('Remote Reporter should', () => {
         metrics = new Metrics(new LocalMetricFactory());
         sender = new UDPSender();
         logger = new MockLogger();
-        reporter = new RemoteReporter(sender, {
+        reporter = new CompositeReporter([new RemoteReporter(sender, {
```
yurishkuro: what is the point of using composite reporter with a single argument?

test/remote_reporter.js
```diff
@@ -81,6 +84,21 @@ describe('Remote Reporter should', () => {
         assert.isOk(LocalBackend.counterEquals(metrics.reporterDropped, 1));
     });
 
+    it ('should have coverage for simple code paths', () => {
+        let sender = new UDPSender();
+        let reporter = new RemoteReporter(sender);
+        assert.equal(reporter.name(), 'RemoteReporter');
```
yurishkuro: don't you need to close remote reporter? otherwise it will leak the timer

test/remote_reporter.js
```diff
@@ -81,6 +84,21 @@ describe('Remote Reporter should', () => {
...
+    });
+
+    it ('should set process without a sender', () => {
```
yurishkuro: don't follow this, there is still a sender, but without setProcess method, which makes little sense since the method is a requirement of the Sender API. Don't know why this test is needed in the first place, since remote reporter without sender is invalid combination, it can just throw exception from the constructor.

test/udp_sender.js
```diff
@@ -176,6 +176,39 @@ describe('udp sender should', () => {
...
+        let spanOne = tracer.startSpan('operation-one');
+        spanOne.finish(); // finish to set span duration
+        spanOne = ThriftUtils.spanToThrift(spanOne);
+        sender._maxSpanBytes = 1;
+        let spanSize = sender._calcSpanSize(spanOne);
+        sender._maxSpanBytes = spanSize * 2;
+
+        let spanThatExceedsCapacity = tracer.startSpan('bigger-span');
+        spanThatExceedsCapacity.finish(); // finish to set span duration
+        spanThatExceedsCapacity.setTag('some-key', 'some-value');
```
yurishkuro: test violates OT API - span methods cannot be called after finish()

test/udp_sender.js
```diff
@@ -176,6 +176,39 @@ describe('udp sender should', () => {
...
+        spanThatExceedsCapacity = ThriftUtils.spanToThrift(spanThatExceedsCapacity);
+
+        let responseOne = sender.append(spanOne);
+        let responseTwo = sender.append(spanThatExceedsCapacity);
+        let expectedBufferSize = sender._calcSpanSize(spanThatExceedsCapacity);
+
+        // sender state reset
+        assert.equal(sender._batch.spans.length, 1);
+        assert.equal(sender._byteBufferSize, expectedBufferSize);
```
yurishkuro: add assertions for response One/Two

test/udp_sender.js
```diff
@@ -176,6 +176,39 @@ describe('udp sender should', () => {
...
+    });
+
+    it('flush returns error, on failed buffer conversion', () => {
+        let span = tracer.startSpan('leela');
+        span.finish(); // finish to set span duration
+        span = ThriftUtils.spanToThrift(span);
+        span.flags = 'string'; // malform the span to create a serialization error
+        sender.append(span);
```
yurishkuro: wouldn't serialization error happen here already?

src/reporters/remote_reporter.js
```diff
@@ -65,11 +65,20 @@ export default class RemoteReporter {
...
+            callback();
+        }
     }
 
     close(callback: ?Function): void {
-        this._sender.close(callback);
+        this._sender.flush();
+        this._sender.close();
         clearInterval(this._intervalHandle);
+
```
oibe: clearInterval is the function to stop the timer.

test/all_reporters.js
oibe: A function to reset spans saved.  Only makes sense for InMemoryReporter, should be removed from other reporters.

test/all_reporters.js
oibe: remote reporter's close method call's the sender's flush method, but sender needs to be initialized.

test/udp_sender.js
oibe: Why would it happen here already?

yurishkuro: because append() needs to know the actual size of the serialized span before deciding whether to add it to the buffer or cause a flush.

src/reporters/remote_reporter.js
yurishkuro: I meant you should stop the timer first before flushing/closing the sender, as otherwise another flush might be called from the timer before you stop it.

test/all_reporters.js
```diff
@@ -0,0 +1,96 @@
...
+            reporter.close(o.callback, reporters.length);
+
+            // remoteReporter.close calls sender.flush, so
+            // sender should be initialized.
+            sender = new UDPSender();
+            sender.setProcess(inMemoryReporter._process);
+            remoteReporter._sender = sender;
+
+            // covered without callbacks
+            reporter.close();
```
yurishkuro: remove L72-79

test/udp_sender.js
oibe: append calculates the size.

oibe: Errors happen on writing a span, not on getting the length.  So the error occurs in flush.  In other jaeger implementations the error would happen on appending because we get the length by writing into a buffer, but with thrftrw exposes a method for getting length that doesn't thrown an error in this case.

PR #55, opened 2016-11-08
Boost sampler test coverage

oibe:


src/samplers/ratelimiting_sampler.js
```diff
@@ -33,7 +33,7 @@ export default class RateLimitingSampler {
         this._logger = logger || new NullLogger();
 
         if (maxTracesPerSecond < 0) {
-            this._logger._error(`maxTracesPerSecond must be greater than 0.0.  Received ${maxTracesPerSecond}`);
+            throw new Error(`maxTracesPerSecond must be greater than 0.0.  Received ${maxTracesPerSecond}`);
             return;
```
yurishkuro: rm return after throw

test/all_samplers.js
```diff
@@ -18,15 +18,84 @@
...
+    });
+
+    describe('ProbabilisticSampler', () => {
+        it ('throws error on out of range sampling rate', () => {
+            expect(() => { new ProbabilisticSampler(2.0); }).to.throw('The sampling rate must be less than 0.0 and greater than 1.0. Received 2');
+        });
+
+        it ('calls is Sampled, and returns false', () => {
+            let sampler = new ProbabilisticSampler(0.0);
+            assert.isNotOk(sampler.isSampled());
```
yurishkuro: not sure what it means to call isSampled without the trace id

test/all_samplers.js
```diff
@@ -18,15 +18,84 @@
...
+        it ('callback is called on close', () => {
+            let sampler = new ProbabilisticSampler(1.0);
+            let callbackCalled = false;
+            let closeCallback = () => {callbackCalled = true; };
+            sampler.close(closeCallback);
+
+            assert.isOk(callbackCalled);
+        });
+
+        it ('close called without callback', () => {
```
yurishkuro: add this to the table driven test instead

test/all_samplers.js
oibe: we use math.random(), because how do you implement an efficient less than on a buffer?

test/all_samplers.js
```diff
@@ -25,25 +25,30 @@ import ConstSampler from '../src/samplers/const_sampler.js';
...
+            { useCallback: [true, false], sampler: new ProbabilisticSampler(0.5) },
+            { useCallback: [true, false], sampler: new RateLimitingSampler(2)},
+            { useCallback: [true, false], sampler: new RemoteSampler('some-service-name') }
         ];
 
         _.each(samplers, (o) => {
             it (o.description + 'close calls callback', () => {
-                let callbackCalled = false;
-                let closeCallback = () => { callbackCalled = true; };
+                if (o.useCallback) {
```
yurishkuro: useCallback is always a non-empty array, so this condition is always true

test/all_samplers.js
```diff
@@ -18,15 +18,78 @@
...
+                new RateLimitingSampler(2),
+                new RemoteSampler('some-service-name')
+            ]
+        });
+
+        _.each(samplers, (o) => {
+            it (o.description + 'close calls callback', () => {
+                if (o.useCallback) {
+                    let callbackCalled = false;
+                    let closeCallback = () => { callbackCalled = true; };
```
yurishkuro: a better test is to use a number, increment in the callback, and assert `== 1`

PR #57, opened 2016-11-09
Crossdock for tchannel-node

oibe:


crossdock/src/http_server.js
```diff
@@ -21,175 +21,62 @@
...
         this._tracer = new Tracer('node', new InMemoryReporter(), new ConstSampler(false));
+        this._helpers = new Helpers(this._tracer);
 
         // json responses need bodyParser when working with express
         app.use(bodyParser.json());
 
         app.post('/start_trace', (req, res) => {
-            this.handleRequest(true, 'start_trace#', req, res);
+            let startRequest: boolean = true;
+            let spanContext = this._tracer.extract(opentracing.FORMAT_HTTP_HEADERS, req.headers);
```
yurishkuro: nit: s/spanContext/parentContext/

crossdock/src/helpers.js
```diff
@@ -20,10 +20,196 @@
...
+            source: crossdockSpec
+        });
+
+        let bridge = new TChannelBridge(this._tracer);
+        this._tracedChannel= bridge.tracedChannel(this._thriftChannel);
+    }
+
+    handleRequest(isStartRequest: boolean, traceRequest: any, spanOrContext: any, operationName: string): void {
+        let serverSpan;
+        if (spanOrContext instanceof Span) {
```
yurishkuro: I don't see why tchannel handler passes `context.openTracingSpan.context()` instead of just `context.openTracingSpan`. If it did the latter, you wouldn't have needed these type checks

crossdock/src/helpers.js
```diff
@@ -20,10 +20,196 @@
...
+            serverSpan = spanOrContext;
+        }
+
+        if (spanOrContext instanceof SpanContext) {
```
yurishkuro: this never makes sense. Server span has already been created and passed as argument.

crossdock/src/helpers.js
```diff
@@ -20,10 +20,196 @@
...
+            serverSpan.setBaggageItem(constants.BAGGAGE_KEY, traceRequest.baggage);
+            if (traceRequest.sampled) {
+                serverSpan.setTag(opentracing.Tags.SAMPLING_PRIORITY, 1);
+            }
+        }
+
+        // do async call to prepareResponse
+        return new RSVP.Promise((resolve, reject) => {
+            this.prepareResponse(traceRequest.downstream, {span: serverSpan}).then((response) => {
+                serverSpan.finish();
```
yurishkuro: not the business of the handler to start/finish spans

crossdock/src/helpers.js
```diff
@@ -20,10 +20,196 @@
...
+        if (isStartRequest) {
+            serverSpan.setBaggageItem(constants.BAGGAGE_KEY, traceRequest.baggage);
+            if (traceRequest.sampled) {
+                serverSpan.setTag(opentracing.Tags.SAMPLING_PRIORITY, 1);
+            }
+        }
+
+        // do async call to prepareResponse
+        return new RSVP.Promise((resolve, reject) => {
+            this.prepareResponse(traceRequest.downstream, {span: serverSpan}).then((response) => {
```
yurishkuro: not following this - prepareResponse already returns a promise, you just need to pass it up, not wrap it.

crossdock/src/helpers.js
```diff
@@ -20,10 +20,196 @@
...
+                    'headers': headers,
+                    'body': JSON.stringify({
+                        'serverRole': downstream.serverRole,
+                        'downstream': downstream.downstream
+                    })
+                }, (err, response) => {
+                    if (err) {
+                        console.log('error in downstream call:', err);
+                        reject(err);
+                        clientSpan.finish();
```
yurishkuro: I think span.Finish should be before the call to reject()

crossdock/src/helpers.js
```diff
@@ -20,10 +20,196 @@
...
+            dns.lookup(downstream.host, (err, address, family) => {
+                if (err) {
+                    console.log('dns_lookup_err', err);
+                    return;
+                }
+
+                let port = parseInt(downstream.port);
+                let downstreamUrl = `http:\/\/${address}:${port}/join_trace`;
+
+                let clientSpan = this._tracer.startSpan('client-span', { childOf: context.span.context() });
```
yurishkuro: if you are using wrapped/decorated tchannel, it should already start client span automatically

crossdock/src/helpers.js
oibe: see above.

crossdock/src/helpers.js
oibe: see above.

crossdock/src/tchannel_server.js
```diff
@@ -20,41 +20,61 @@
...
 
     handleTChannelRequest(context: any, req: any, head: any, body: any, callback: Function) {
-        callback(null, {
+        let isStartRequest: boolean = false;
+        let traceRequest = body.request;
+
+        let promise = this._helpers.handleRequest(
+            isStartRequest,
+            traceRequest,
+            context.openTracingSpan.context(),
```
oibe: So this is the incredibly strange bug.  If you remove context, and pass in a span then yo go down the code path just accepts a span.  The moment I do that outgoing tchannel requests begin to fail.  What is stranger to me though is that if I modify tchannel_bridge.js to create the context() in the handler then this still fails....

crossdock/src/helpers.js
oibe: I can't move this out of here until i figure out the bug above, that lets me pass in the span instead, of create it in this scope.

crossdock/src/helpers.js
yurishkuro: ?

yurishkuro:
We need to acc crossdock run to .travis, otherwise it's not being tested

PR #58, opened 2016-11-09
Remove yaml validation and add flow to configuration

oibe:


src/configuration.js
```diff
@@ -80,58 +54,58 @@ export default class Configuration {
...
-                hostPort = `${config.reporter.agentHost}:${config.reporter.agentPort}`;
+            if (config.agentHost && config.agentPort) {
+                hostPort = `${config.agentHost}:${config.agentPort}`;
             }
         }
 
+        // $FlowIgnore - disable type inference for udpsender.
         sender = new UDPSender(hostPort);
         reporters.push(new RemoteReporter(sender, reporterConfig));
         return new CompositeReporter(reporters);
```
yurishkuro: it's not necessary to always use composite reporter. You can do something like
```
reporter = new RemoteReporter();
if config.logSpans {
  reporter = new CompositeReporter([reporter, new LoggingReporter()])
}
```

package.json
```diff
@@ -19,8 +19,6 @@
     "buffer-equal": "^1.0.0",
     "long": "^3.2.0",
     "node-int64": "^0.4.0",
```
yurishkuro: Booked https://github.com/uber/jaeger-client-node/issues/59

src/configuration.js
```diff
@@ -80,58 +54,58 @@ export default class Configuration {
...
-    static initTracer(config, options = {}) {
-        let v = new Validator();
-        v.validate(config, jaegerSchema, {
-            throwError: true
-        });
-
-        let reporters = [];
-        let reporter;
-        let sampler;
+    static initTracer(config: TracerConfig, options: any = {}): Tracer {
```
yurishkuro: * please document `options`. Maybe add another type to `src/decls/types.js`? Or even better, why can we not use just the TracerConfig for that?  That's what we do in other languages - some fields of the config object are meant to be loaded from a file, but others can be assigned programmatically, like the logger.
* add comment `// see ./decls/types.js for TracerConfig`

src/decls/types.js
```diff
@@ -93,3 +93,22 @@ declare type TChannelSpan = {
     parentid: Array<number>,
     flags: number
 };
+
+declare type ReporterConfig = {
+    flushIntervalMs: ?number,
```
yurishkuro: is milliseconds consistent with the other languages? I know in Go the value is actually typed. In Python it's probably a float as seconds, which seems the most sensible default scale

src/reporters/composite_reporter.js
```diff
@@ -33,24 +33,50 @@ export default class CompositeReporter {
     }
 
     report(span: Span): void {
-        for (let i = 0; i < this._reporters.length; i++) {
-            let reporter = this._reporters[i];
-            reporter.report(span);
-        }
+        this._reporters.map((r) => {
+            r.report(span);
+        });
```
yurishkuro: is this decision based on profiling? I'd be surprised if a monad is faster than indexed array access.

src/reporters/composite_reporter.js
```diff
@@ -33,24 +33,50 @@ export default class CompositeReporter {
...
-        }
+        let count = 0;
+        let cb = () => {
+            count++;
+            if (count >= this._reporters.length) {
+                if (callback) {
+                    callback();
+                }
+            }
+        };
```
yurishkuro: DRY

src/reporters/remote_reporter.js
```diff
@@ -27,7 +27,7 @@ import NoopMetricFactory from '../metrics/noop/metric_factory';
 const DEFAULT_BUFFER_FLUSH_INTERVAL_MILLIS = 10000;
```
yurishkuro: is 10s consistent with other languages?  I rather thing 1s default would be preferable.

test/all_reporters.js
```diff
@@ -0,0 +1,95 @@
...
+import RemoteReporter from '../src/reporters/remote_reporter';
+import UDPSender from '../src/reporters/udp_sender';
+import MockLogger from './lib/mock_logger';
+import LoggingReporter from '../src/reporters/logging_reporter';
+
+describe('All Reporters should', () => {
+    it ('have proper names', () => {
+        let loggingReporter = new LoggingReporter();
+        let inMemoryReporter = new InMemoryReporter();
+        inMemoryReporter.setProcess('service-name', []);
```
yurishkuro: why does only this one deserve the honor?

test/all_reporters.js
```diff
@@ -0,0 +1,95 @@
...
+
+        sender = new UDPSender();
+        sender.setProcess(inMemoryReporter._process);
+        remoteReporter._sender = sender;
+
+        // covered without callbacks
+        reporter.flush();
+        reporter.close(() => {
+            assert.isOk(closeCalled);
+            assert.isOk(flushCalled);
```
yurishkuro: this doesn't make sense - the booleans are already set from L69. And even if this was a standalone test, you cannot put asserts in the callback when you're testing if the callback is actually being called

test/remote_reporter.js
```diff
@@ -40,10 +42,10 @@ describe('Remote Reporter should', () => {
         metrics = new Metrics(new LocalMetricFactory());
         sender = new UDPSender();
         logger = new MockLogger();
-        reporter = new RemoteReporter(sender, {
+        reporter = new CompositeReporter([new RemoteReporter(sender, {
```
yurishkuro: not clear why this needs to use composite reporter

src/reporters/composite_reporter.js
oibe: The popular opinion is that map is faster, and that this is the more javascript-centric style.

src/decls/types.js
oibe: its seconds in go, and python, but  milliseconds in java.  I think this needs to be different depending on what the timers for each language take.

test/all_reporters.js
oibe: so eventhough they are set in line 69 they are bound to the functions scope, because of rules of closure.  However I see your point about asserting in the callback.

src/reporters/composite_reporter.js
yurishkuro: Popular opinion just elected Trump. I only trust benchmarks.

If functional style is faster that for-loop, then at least this should be `forEach`, not `map`, because map is used to transform collection into another list, but here we're not interested in any return values, so `forEach` (which returns void) is the correct monad to use.

src/reporters/composite_reporter.js
```diff
@@ -33,24 +33,43 @@ export default class CompositeReporter {
...
+    }
+
+    flush(callback: ?Function): void {
```
yurishkuro: remind me again - why do we need flush() function? Why isn't close() enough?

src/reporters/composite_reporter.js
```diff
@@ -33,24 +33,43 @@ export default class CompositeReporter {
...
-        for (let i = 0; i < this._reporters.length; i++) {
-            let reporter = this._reporters[i];
-            reporter.close();
-        }
-
-        if (callback) {
-            callback();
-        }
+
+        this._reporters.map((r) => {
```
yurishkuro: forEach

src/reporters/composite_reporter.js
```diff
@@ -33,24 +33,43 @@ export default class CompositeReporter {
...
+            r.close(this._executeCallbackLast(callback));
```
yurishkuro: bug

src/reporters/composite_reporter.js
oibe: some reporters have flush, so i thought it made sense to have flush on all of them.

src/configuration.js
```diff
@@ -80,58 +54,58 @@ export default class Configuration {
         }
 
         if (type === 'remote') {
```
oibe: I don't follow.  type is checked for in the previous function and why != 'remote'?

src/reporters/composite_reporter.js
yurishkuro: why does any reporter needs flush()? I see that the only place it's called from is in `Tracer.flush()`, which is not an open-tracing API method. I think it's unnecessary & redundant functionality. We only need close().

src/configuration.js
yurishkuro: if at this point you have not created a sampler and the `type` is not empty and not `remote`, then the `type` is unsupported value, thus exception. 
```
if (!sampler) {
  if (type && type != 'remote') {
    throw
  }
  sampler = new RemoteSampler(serviceName, {
                  sampler: new ProbabilisticSampler(param)
  });  
}
```

src/decls/types.js
yurishkuro: Let's leave it

src/decls/types.js
```diff
@@ -93,3 +93,27 @@ declare type TChannelSpan = {
...
+    logSpans: ?boolean,
+    agentHost: ?string,
+    agentPort: ?number
+};
+
+declare type ReporterOptions = {
+    reporter: Reporter,
```
yurishkuro: when I removed it, tests were green

src/reporters/composite_reporter.js
```diff
@@ -33,24 +33,43 @@ export default class CompositeReporter {
...
+                if (callback) {
+                    callback();
+                }
+            }
+        };
+    }
+
+    flush(callback: ?Function): void {
+        this._reporters.map((r) => {
+            r.flush(this._executeCallbackLast(callback));
```
yurishkuro: `this._executeCallbackLast(callback)` creates a new wait-group style callback, that counts to N. You need to create that wait group **outside the loop**, not a new one on every iteration of the loop.

test/all_reporters.js
yurishkuro: how come the same new file is in two PRs?

PR #60, opened 2016-11-09
Extend span and tracer properties

oibe:
- expose property to set name
- expose property to get reporter form tracer


src/tchannel_bridge.js
```diff
@@ -139,7 +139,7 @@ export default class TChannelBridge {
         return channel;
     }
 
-    _getTChannelParentSpan(): TChannelSpan {
+    static getTChannelParentSpan(): TChannelSpan {
```
yurishkuro: are you sure about this? This is a bad method name anyway, would be more accurate to call it "makeFakeTChannelParentSpan()", because we return 0 trace id.

oibe: sure.

PR #62, opened 2016-11-11
Add finished flag

oibe:


src/span.js
```diff
@@ -37,6 +37,7 @@ export default class Span {
     _tags: Array<Tag>;
     static _baggageHeaderCache: any;
     _references: Array<Reference>
```
black-adder: not related to this diff but why does this guy not in in ";"?

src/span.js
```diff
@@ -154,11 +160,12 @@ export default class Span {
      * @param {number} finishTime - The time on which this span finished.
      **/
     finish(finishTime: ?number): void {
-        if (this._duration !== undefined) {
+        if (this.finished) {
             throw new Error('You can only call finish() on a span once.');
         }
 
         if (this._spanContext.isSampled()) {
+            this._finished = true;
```
black-adder: is this right? only if a span is sampled it counts as finished? Shouldn't it be finished if finish is called at least once?

src/span.js
```diff
@@ -64,6 +66,10 @@ export default class Span {
         return this._operationName;
     }
 
+    get finished(): boolean {
```
yurishkuro: why does this have to be public? do not extend the public interface unless you have to - much easier to add later than to remove.

src/span.js
```diff
@@ -154,11 +160,12 @@ export default class Span {
      * @param {number} finishTime - The time on which this span finished.
      **/
     finish(finishTime: ?number): void {
-        if (this._duration !== undefined) {
+        if (this.finished) {
             throw new Error('You can only call finish() on a span once.');
```
yurishkuro: how about we just log it? not only throwing exception breaks the 1st rule of monitoring "do no harm", but it also prevents optimization of the method by the JS engine.

src/span.js
oibe: it should have that.

src/span.js
```diff
@@ -155,7 +155,8 @@ export default class Span {
      **/
     finish(finishTime: ?number): void {
         if (this._duration !== undefined) {
-            throw new Error('You can only call finish() on a span once.');
+            this.tracer()._logger.error('You can only call finish() on a span once.');
```
yurishkuro: nit: the error should provide better context. At minimum log the operation name so that people know which span is being double-finished.

PR #63, opened 2016-11-11
Support jaeger-baggage header

oibe:


src/span_context.js
```diff
@@ -54,6 +54,11 @@ export default class SpanContext {
         this._debugId = debugId;
     }
 
+    // An empty context can be created in crossdock tests
+    get isEmptyContext(): any {
```
black-adder: why can't this return bool? is it a flow thing?

src/span_context.js
```diff
@@ -136,7 +141,9 @@ export default class SpanContext {
     }
 
     isDebugIDContainerOnly(): boolean {
-        return !this._traceId && this._debugId !== '';
+        return  !this._traceId &&
```
black-adder: why the extra space indentation?

black-adder: also, you could do `isEmptyContext && this._debugId !== '';`

src/tracer.js
```diff
@@ -206,16 +204,13 @@ export default class Tracer {
         let rpcServer = (spanKindValue === opentracing_tags.SPAN_KIND_RPC_SERVER);
 
 
-        // $FlowIgnore - I just want a span context up front.
         let ctx: SpanContext = new SpanContext();
         let samplerTags: any = {};
-        let debugRequest = (parent && parent.isDebugIDContainerOnly());
```
black-adder: nit: do you mind if we keep this variable even if we only use it once on line 212? it makes it slightly more readable

src/tracer.js
```diff
@@ -184,20 +184,18 @@ export default class Tracer {
             startTime = Utils.getTimestampMicros();
         }
 
-        let followsFromIsParent = false;
```
black-adder: good find

src/span_context.js
```diff
@@ -54,6 +54,11 @@ export default class SpanContext {
         this._debugId = debugId;
     }
 
+    // An empty context can be created in crossdock tests
```
black-adder: This comment doesn't have much relevance here

test/tracer.js
```diff
@@ -50,6 +50,20 @@ describe('tracer should', () => {
         tracer.close();
     });
 
+    it ('begin a new span given an empty context', () => {
```
yurishkuro: s/empty context/only baggage headers/

src/tracer.js
```diff
@@ -184,6 +184,8 @@ export default class Tracer {
             startTime = Utils.getTimestampMicros();
         }
 
+        // This boolean ensures that a parent with reference type 'follows_from'
+        // is selected when no childOf is passed in.
```
yurishkuro: the opposite - it ensures that CHILD_OF reference is preferred as a parent even if it comes after FOLLOWS_FROM

src/tracer.js
```diff
@@ -206,16 +208,13 @@ export default class Tracer {
...
-        if (!parent || debugRequest) {
+        if (!parent || !parent.isValid) {
             let randomId = Utils.getRandom64();
             let flags = 0;
-            if (debugRequest) {
+            if (parent && parent.isDebugIDContainerOnly()) {
                 flags |= (constants.SAMPLED_MASK | constants.DEBUG_MASK);
-                // $FlowIgnore - parent can't be null, if debugRequest is true. Flow doesn't realize this.
                 samplerTags[constants.JAEGER_DEBUG_HEADER] = parent.debugId;
             } else if (this._sampler.isSampled()) {
```
yurishkuro: add
```
} else if (parent) {
    ctx.baggage = parent.baggage;
}
```

test/tracer.js
```diff
@@ -50,6 +50,18 @@ describe('tracer should', () => {
...
+    it ('begin a new span given only baggage headers', () => {
+        // Users want to sometimes pass baggage even if there is no span, so we allow for
+        // an empty context to be created.  In this case we must ensure a new span is created.
+        let headers = {};
+        headers[constants.TRACER_BAGGAGE_HEADER_PREFIX + 'keyOne'] = 'leela';
+        let spanContext = tracer.extract(opentracing.FORMAT_TEXT_MAP, headers);
+        let rootSpan = tracer.startSpan('fry', { childOf: spanContext });
+
+        assert.isOk(rootSpan.context().traceId);
+        assert.isOk(rootSpan.context().spanId);
```
yurishkuro: test that the rootSpan has the correct baggage

PR #64, opened 2016-11-11
Do not double-finish server span

yurishkuro:


crossdock/src/helpers.js
```diff
@@ -65,24 +65,19 @@ export default class Helpers {
         }
 
         // do async call to prepareResponse
-        return new RSVP.Promise((resolve, reject) => {
-            this.prepareResponse(traceRequest.downstream, {span: serverSpan}).then((response) => {
-                serverSpan.finish();
```
yurishkuro: here was the bug

PR #71, opened 2016-11-15
Add flow to config

oibe:


src/_interfaces/reporter.js
```diff
@@ -24,7 +24,7 @@ import Span from '../span.js';
 declare interface Reporter {
     report(span: Span): void;
     close(callback: ?Function): void;
-    setProcess(serviceName: string, tags: Array<Tag>): void;
+    name(): string;
```
yurishkuro: is this rebased off master? don't we need setProcess method?

oibe: We don't need setProcess because then flow expects every reporter to have setProcess.

oibe: Only remote, and composite reporter really need setProcess though, because only they delegate anything to the sender.

yurishkuro: This is a false argument. Currently setProcess() is called by the Tracer, irrespective of what type of Reporter we give to the tracer, so this method is a part of the expected Reporter interface. It is also a semantic dependency, because Process information is a part of every span, but we decided not to keep passing this data to the  `report()` method every time as an optimization, since the process never changes after initialization.

yurishkuro:
Very old PR with unclear purpose, closing.

PR #74, opened 2016-11-17
Change samplers to return tags in isSampled() call

yurishkuro:


src/_interfaces/sampler.js
```diff
@@ -20,8 +20,22 @@
...
+    
+    /**
+     * Decides if a new trace starting with given operation name
+     * should be sampled or not. If the method returns true, it
+     * must populate the tags dictionary with tags that identify
+     * the sampler, namely sampler.type and sampler.param.
+     *
+     * @param {string} operation - Operation name of the root span.
+     * @param {Object} tags - output dictionary to store sampler tags.
+     * @return {booleab} - returns whether the trace should be sampled.
```
black-adder: s/booleab/boolean

test/all_samplers.js
```diff
@@ -86,27 +87,35 @@ describe('samplers should', () => {
...
-            {'sampler': new ConstSampler(true), 'type': constants.SAMPLER_TYPE_CONST, 'param': true},
-            {'sampler': new ConstSampler(false), 'type': constants.SAMPLER_TYPE_CONST, 'param': false},
-            {'sampler': new ProbabilisticSampler(0.1), 'type': constants.SAMPLER_TYPE_PROBABILISTIC, 'param': 0.1},
-            {'sampler': new RateLimitingSampler(2), 'type': constants.SAMPLER_TYPE_RATE_LIMITING, 'param': 2},
-            {'sampler': new RemoteSampler('some-caller-name'), 'type': constants.SAMPLER_TYPE_PROBABILISTIC, 'param': 0.001},
+            {sampler: new ConstSampler(true), 'type': constants.SAMPLER_TYPE_CONST, param: true, decision: true},
+            {sampler: new ConstSampler(false), 'type': constants.SAMPLER_TYPE_CONST, param: false, decision: false},
+            {sampler: new ProbabilisticSampler(1.0), 'type': constants.SAMPLER_TYPE_PROBABILISTIC, param: 1.0, decision: true},
+            {sampler: new RateLimitingSampler(2), 'type': constants.SAMPLER_TYPE_RATE_LIMITING, param: 2, decision: true},
+            // {sampler: new RemoteSampler('some-caller-name'), 'type': constants.SAMPLER_TYPE_PROBABILISTIC, param: 1.0, decision: true},
```
black-adder: ?

src/_interfaces/sampler.js
```diff
@@ -20,8 +20,22 @@
...
+    name(): string;
+    
+    /**
+     * Decides if a new trace starting with given operation name
+     * should be sampled or not. If the method returns true, it
+     * must populate the tags dictionary with tags that identify
+     * the sampler, namely sampler.type and sampler.param.
+     *
+     * @param {string} operation - Operation name of the root span.
+     * @param {Object} tags - output dictionary to store sampler tags.
```
black-adder: just curious, why did you choose this approach rather than passing back tags that you can initialize once in your constructors?

yurishkuro: I am trying to minimize the number of object allocations. Javascript does not support multi-value functions, so I would need to return something like `{sampled: bool, tags: map}` objects, which (a) need to be allocated, and (b) in most cases will say sampled=false where tags are not relevant.

PR #82, opened 2016-12-05
Change the tchannel bridge to use a context that supports 'setSpan', and 'getSpan'.

oibe:


README.md
```diff
@@ -33,8 +33,10 @@ a tchannel bridge, and wrap there encoded handler function with `tracedHandler`
 
     serverThriftChannel.register(server, 'Echo::echo', context, bridge.tracedHandler(
         (context, req, head, body, callback) => {
-            // context will contain an 'openTracingSpan' field that stores the tracing span for the inbound request.
-            // Your handler code goes here.
+            // context must be an object with the methods 'setSpan', and 'getSpan' which are responsible,
```
black-adder: "context must implement 'setSpan' and 'getSpan' which are responsible for setting and getting the span on the context respectively."

Reads easier.

src/default_context.js
```diff
@@ -0,0 +1,32 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+export default class DefaultContext {
```
black-adder: Do we want to create a context interface with the setSpan and getSpan functions that DefaultContext implements?

src/tchannel_bridge.js
```diff
@@ -76,8 +77,7 @@ export default class TChannelBridge {
                 span.setTag('as', request.headers.as);
             }
 
-            // In theory may overwrite tchannel span, but thats what we want anyway.
-            context.openTracingSpan = span;
+            context.setSpan(span);
```
yurishkuro: what is the type of `context` object here?

test/tchannel_bridge.js
```diff
@@ -89,7 +92,7 @@ describe ('test tchannel span bridge', () => {
             });
 
             // register the server request function.
-            let context = {};
+            let context = new DefaultContext();
             encodedChannel.register(server, 'Echo::echo', context, bridge.tracedHandler(handleServerReq));
```
yurishkuro: does tchannel use this `context` object as a prototype? A single instance cannot be shared by all inbound requests.

src/default_context.js
oibe: Since we will only have one type of context the simplest thing is to annotate context variables with DefaultContext.

src/tchannel_bridge.js
```diff
@@ -19,6 +19,7 @@
 // THE SOFTWARE.
 
 import * as constants from './constants';
+import DefaultContext from '../src/default_context';
```
yurishkuro: why not `./default_context`?

src/tchannel_bridge.js
```diff
@@ -76,8 +78,7 @@ export default class TChannelBridge {
                 span.setTag('as', request.headers.as);
             }
 
-            // In theory may overwrite tchannel span, but thats what we want anyway.
-            context.openTracingSpan = span;
+            context.setSpan(span);
 
             // remove headers prefixed with $tracing$
             let headerKeys = Object.keys(headers);
```
yurishkuro: L93: `handlerFunc(context, request, headers, body, wrappingCallback);`

Aren't you making it backwards incompatible? Didn't we say it should be this?
```
request.context = context;
handlerFunc(context, request, headers, body, wrappingCallback);
```

README.md
```diff
@@ -31,10 +31,13 @@ a tchannel bridge, and wrap there encoded handler function with `tracedHandler`
         entryPoint: path.join(__dirname, 'thrift', 'echo.thrift') // file path to a thrift file
     });
 
-    serverThriftChannel.register(server, 'Echo::echo', context, bridge.tracedHandler(
+    let perProcessOptions = {};
+    serverThriftChannel.register(server, 'Echo::echo', perProcessOptions, bridge.tracedHandler(
         (context, req, head, body, callback) => {
-            // context will contain an 'openTracingSpan' field that stores the tracing span for the inbound request.
-            // Your handler code goes here.
+            // Context must implement 'setSpan', and 'getSpan' which are responsible,
```
yurishkuro: the real context is in `req.context`

src/tchannel_bridge.js
```diff
@@ -61,7 +62,8 @@ export default class TChannelBridge {
      * a the handler's context with a span.
      **/
     tracedHandler(handlerFunc: any, options: startSpanArgs = {}): Function {
-        return (context, request, headers, body, callback) => {
+        return (perProcessOptions, request, headers, body, callback) => {
+            let context: DefaultContext = new DefaultContext();
```
yurishkuro: if you do this, you cannot allow the user to substitute this for RPCContext. My recommendation is to accept a factory method `makeContext = func() { return new DefaultContext(); }` in the constructor and use it to instantiate a context when we need it. It's the same default behavior, but users can override that function with `return new RPCContext()`

src/tchannel_bridge.js
```diff
@@ -63,7 +65,7 @@ export default class TChannelBridge {
      **/
     tracedHandler(handlerFunc: any, options: startSpanArgs = {}): Function {
         return (perProcessOptions, request, headers, body, callback) => {
-            let context: DefaultContext = new DefaultContext();
+            let context: Context = this._contextFactory();
```
yurishkuro: `: Context` - 👍

README.md
```diff
@@ -18,12 +18,20 @@ Because [tchannel-node](https://github.com/uber/tchannel-node) does not have ins
 for OpenTracing, Jaeger-Client exposes methods wrapping tchannel handlers, and encoded channels.
 An encoded channel is a channel wrapped in either a thrift encoder `TChannelAsThrift`,
 or json encoder `TChannelAsJson`.  To wrap a server handler for thrift one can initialize
-a tchannel bridge, and wrap there encoded handler function with `tracedHandler` decorator.
+a tchannel bridge, and wrap the encoded handler function with a `tracedHandler` decorator.
+The tchannel bridge takes an OpenTracing  tracer, and a context factory.  The context factory
+must be a function that returns a context with the methods 'getSpan', and 'setSpan' which retrieve 
+and assign the span to the context respectively.
```
yurishkuro: fwiw, you can even relax the requirement for getSpan/setSpan methods by taking respective functions as constructor arguments. It would allow people to use a context object that simply has a field for a span, rather than methods. E.g.
```
function getSpan(context: Context): opentracing.Span
function setSpan(context: Context, span: opentracing.Span): void
```

src/tchannel_bridge.js
```diff
@@ -19,27 +19,29 @@
...
 
 let TCHANNEL_TRACING_PREFIX = '$tracing$';
 
 export default class TChannelBridge {
     _tracer: Tracer;
     _codec: TextMapCodec;
+    _contextFactory: Function;
 
-    constructor(tracer: Tracer) {
+    constructor(tracer: Tracer, contextFactory: Function) {
```
yurishkuro: I would suggest going with `options` 2nd parameter and reading `contextFactoryFn` from it. It makes this bridge easier to extend in the future.

yurishkuro: btw, this file is missing `@flow`. This should've been `?Function`.

src/tchannel_bridge.js
```diff
@@ -83,24 +84,24 @@ export default class TChannelBridge {
...
+            handlerFunc(perProcessOptions, request, headers, body, wrappingCallback);
         };
     }
 
-    _wrapTChannelSend(wrappedSend, channel, req, endpoint, headers, body, callback) {
+    _wrapTChannelSend(wrappedSend: Function, channel: any, req: any, endpoint: string, headers: any, body: any, callback: Function) {
         headers = headers || {};
         let context: Context = req.context || this._contextFactory();
-        let childOf = context.getSpan();
+        let childOf: SpanContext = context.getSpan();
```
yurishkuro: shouldn't this be `Span`?

src/tchannel_bridge.js
```diff
@@ -60,14 +64,15 @@ export default class TChannelBridge {
...
+    tracedHandler(handlerFunc: any, options: any = {}): Function {
+        return (perProcessOptions, request, headers, body, callback) => {
+            let context: Context = this._contextFactory();
+            let operationName: string = options.operationName || request.arg1;
+            let span: Span = this._extractSpan(operationName, headers);
 
             // set tags
             span.setTag(opentracing.Tags.PEER_SERVICE, request.callerName);
-            let hostPort = request.remoteAddr.split(':');
+            let hostPort: string = request.remoteAddr.split(':');
```
yurishkuro: split returns array, not a string

src/tchannel_bridge.js
```diff
@@ -76,27 +81,27 @@ export default class TChannelBridge {
                 span.setTag('as', request.headers.as);
             }
 
-            // In theory may overwrite tchannel span, but thats what we want anyway.
-            context.openTracingSpan = span;
+            context.setSpan(span);
 
             // remove headers prefixed with $tracing$
-            let headerKeys = Object.keys(headers);
+            let headerKeys: any = Object.keys(headers);
```
yurishkuro: Array<string>

test/tchannel_bridge.js
```diff
@@ -88,17 +91,17 @@ describe ('test tchannel span bridge', () => {
                 entryPoint: path.join(__dirname, 'thrift', 'echo.thrift') // ignored in json case
             });
 
-            // register the server request function.
-            let context = {};
-            encodedChannel.register(server, 'Echo::echo', context, bridge.tracedHandler(handleServerReq));
+            // We pass 'null' as the context because the handler will provide a context
+            // that gets created on every request.
```
yurishkuro: incorrect comment. I would recommend `let options: any = {}` and pass `options`, to match the semantics of the method signature.

PR #84, opened 2016-12-13
1.3.0

oibe:


black-adder:
Is this something that is blocking teams? Because it contains the adaptive sampling changes which *should* work but we haven't tested it that thoroughly yet.

yurishkuro:
@black-adder it should be backwards compatible though

Issue #80, opened 2016-11-29
Redo per-operation sampling if span.operationName is modified


yurishkuro:
Sometimes the span is started by instrumentation in a place were we do not have access to the route, and the concrete operation name is assigned later. In case of per-operation sampling we need to redo the sampling decision when operation name is changed.
  * It probably only needs to apply to root span
  * It means we should not be ignoring tags created during span construction when the span is not initially sampled

yurishkuro:
Fixed by #85

PR #85, opened 2017-01-05
Re-sample the span when operationName is set after span creation

oibe:
This preserves adaptive sampling by allowing a proper operationName
to be correlated with the correct sampling rate (eventhough the
sampling decision has already happened.)

### DIFF
--- a/src/span.js
+++ b/src/span.js
@@ -138,13 +138,15 @@ export default class Span {
 
     /**
      * Checks whether or not a span can be written to.
+     * A span can be written to if it:
+     *  - 'finish' has not been called on the span.
+     *  - the span is not finalized.
+     *  - the span has been finalized, but also is sampled.
+     *  - the span has not been serialized using injectors.
      *
      * @return {boolean} - The decision about whether this span can be written to.
      **/
     _isWriteable(): boolean {
-        // A span can NOT be written to if it:
-        // 1.) is finished.
-        // 2.) has been finalized, and is not sampled
         return  this._duration === undefined &&
                 (!this._spanContext.samplingFinalized ||
                 this._spanContext.isSampled());
@@ -157,31 +159,24 @@ export default class Span {
      * @return {Span} - returns this span.
      **/
     setOperationName(operationName: string): Span {
-        if (!this._isWriteable()) {
+        // TODO:oibe is it wrong to allow this to be set even if we are finished?
+        this._operationName = operationName;
+        // We re-sample the span if all of the following conditions hold:
+        // 1.)  The span is not marked as a 'debug' span.
+        // 2.)  The span is not the first in the process.
+        // 3.)  '_isWriteable' returns true.
+        if (this._spanContext.isDebug()  || !this._isWriteable()) {
             return this;
         }
 
-        this._operationName = operationName;
-        // If the span has debug set, then the question of resampling
-        // after setting the operation name is irrelevant.  The span is always
-        // sampled.
-
-        // Do not re-sample the span if:
-        // 1.)  Debug has been set on the span.  Sampling is irrelevant in this case
-        //      the span will be reported no matter what.
-        // 2.)  The span is the first in process.  If the span is the first in process
-        //      then the sampling decision has already been made.
-        // 3.)  The span is not writable.  This can happen if the span's context was used
-        //      to create another span, or if the span is finished.
-        if (!this._spanContext.isDebug() || !this._firstInProcess) {
-            let sampler = this.tracer()._sampler;
-            let tags = {};
-            if (sampler.isSampled(operationName, tags)) {
-                this._spanContext.flags |= constants.SAMPLED_MASK;
-                this.addTags(tags);
-            }
+        let sampler = this.tracer()._sampler;
+        let tags = {};
+        if (!this._spanContext._samplingFinalized && sampler.isSampled(operationName, tags)) {
+            this._spanContext.flags |= constants.SAMPLED_MASK;
+            this.addTags(tags);
         }
-        this._spanContext.samplingFinalized = true;
+        this._spanContext.finalizeSampling();
+
         return this;
     }
 
--- a/src/span.js
+++ b/src/span.js
@@ -22,7 +22,7 @@
 import * as constants from './constants.js';
 import NullLogger from './logger.js';
 import SpanContext from './span_context.js';
-import {Tags as opentracing_tags} from 'opentracing';
+import * as opentracing from 'opentracing';
 import Utils from './util.js';
 
 export default class Span {
@@ -138,18 +138,18 @@ export default class Span {
 
     /**
      * Checks whether or not a span can be written to.
-     * A span can be written to if it:
-     *  - 'finish' has not been called on the span.
-     *  - the span is not finalized.
-     *  - the span has been finalized, but also is sampled.
-     *  - the span has not been serialized using injectors.
+     * A span is finalized when either of these happens:
+     * - it has inherited the sampling decision from its parent
+     * - its debug flag is set via the sampling.priority tag
+     * - it is finish()-ed
+     * - setOperationName is called
+     * - it is used as a parent for another span
+     * - its context is serialized using injectors
      *
      * @return {boolean} - The decision about whether this span can be written to.
      **/
     _isWriteable(): boolean {
-        return  this._duration === undefined &&
-                (!this._spanContext.samplingFinalized ||
-                this._spanContext.isSampled());
+        return !this._spanContext.samplingFinalized || this._spanContext.isSampled();
     }
 
     /**
@@ -161,11 +161,8 @@ export default class Span {
     setOperationName(operationName: string): Span {
         // TODO:oibe is it wrong to allow this to be set even if we are finished?
         this._operationName = operationName;
-        // We re-sample the span if all of the following conditions hold:
-        // 1.)  The span is not marked as a 'debug' span.
-        // 2.)  The span is not the first in the process.
-        // 3.)  '_isWriteable' returns true.
-        if (this._spanContext.isDebug()  || !this._isWriteable()) {
+        // We re-sample the span if it has not been finalized.
+        if (this._spanContext.samplingFinalized) {
             return this;
         }
 
@@ -193,6 +190,7 @@ export default class Span {
             return;
         }
 
+        this._spanContext.finalizeSampling();
         if (this._spanContext.isSampled()) {
             let endTime = finishTime || Utils.getTimestampMicros();
             this._duration = endTime - this._startTime;
@@ -208,9 +206,9 @@ export default class Span {
      * @return {Span} - returns this span.
      **/
     addTags(keyValuePairs: any): Span {
-        if (opentracing_tags.SAMPLING_PRIORITY in keyValuePairs) {
-            this._setSamplingPriority(keyValuePairs[opentracing_tags.SAMPLING_PRIORITY]);
-            delete keyValuePairs[opentracing_tags.SAMPLING_PRIORITY];
+        if (opentracing.Tags.SAMPLING_PRIORITY in keyValuePairs) {
+            this._setSamplingPriority(keyValuePairs[opentracing.Tags.SAMPLING_PRIORITY]);
+            delete keyValuePairs[opentracing.Tags.SAMPLING_PRIORITY];
         }
 
         if (this._isWriteable()) {
@@ -233,7 +231,7 @@ export default class Span {
      * @return {Span} - returns this span.
      * */
     setTag(key: string, value: any): Span {
-        if (key === opentracing_tags.SAMPLING_PRIORITY) {
+        if (key === opentracing.Tags.SAMPLING_PRIORITY) {
             this._setSamplingPriority(value);
             return this;
         }
@@ -276,6 +274,7 @@ export default class Span {
     _setSamplingPriority(priority: number): void {
         if (priority > 0) {
             this._spanContext.flags = this._spanContext.flags | constants.SAMPLED_MASK | constants.DEBUG_MASK;
+            this._spanContext.finalizeSampling();
         } else {
             this._spanContext.flags = this._spanContext.flags & (~constants.SAMPLED_MASK);
         }
--- a/src/span.js
+++ b/src/span.js
@@ -137,24 +137,51 @@ export default class Span {
     }
 
     /**
+     * Checks whether or not a span can be written to.
+     *
+     * @return {boolean} - The decision about whether this span can be written to.
+     **/
+    _isWriteable(): boolean {
+        // A span can NOT be written to if it:
+        // 1.) is finished.
+        // 2.) has been finalized, and is not sampled
+        return  this._duration === undefined &&
+                (!this._spanContext.samplingFinalized ||
+                this._spanContext.isSampled());
+    }
+
+    /**
      * Sets the operation name on this given span.
      *
      * @param {string} name - The name to use for setting a span's operation name.
      * @return {Span} - returns this span.
      **/
     setOperationName(operationName: string): Span {
+        if (!this._isWriteable()) {
+            return this;
+        }
+
         this._operationName = operationName;
         // If the span has debug set, then the question of resampling
         // after setting the operation name is irrelevant.  The span is always
         // sampled.
-        if (!this._spanContext.isDebug()) {
+
+        // Do not re-sample the span if:
+        // 1.)  Debug has been set on the span.  Sampling is irrelevant in this case
+        //      the span will be reported no matter what.
+        // 2.)  The span is the first in process.  If the span is the first in process
+        //      then the sampling decision has already been made.
+        // 3.)  The span is not writable.  This can happen if the span's context was used
+        //      to create another span, or if the span is finished.
+        if (!this._spanContext.isDebug() || !this._firstInProcess) {
             let sampler = this.tracer()._sampler;
-            var tags = {};  // TODO:oibe shouldn't this be initialized if not passed?
+            let tags = {};
             if (sampler.isSampled(operationName, tags)) {
                 this._spanContext.flags |= constants.SAMPLED_MASK;
+                this.addTags(tags);
             }
         }
-        this._spanContext.samplingFinished = true;
+        this._spanContext.samplingFinalized = true;
         return this;
     }
 
@@ -179,19 +206,6 @@ export default class Span {
     }
 
     /**
-     * Checks whether a span's sampling decision is finalized by calling setOperationName 
-     * This is used to hold onto tags, and logs, so that we can re-sample the span after
-     * an operation name is set on the span.  This is necessary because setting a sampling
-     * rate on the span before we know the operation name means adaptive-sampling
-     * does not know how to correlate the operation with sampling decisions.
-     *
-     * @return {boolean} - The decision about whether this span is done being sampled.
-     **/
-    isSampledAndFinished(): boolean {
-        return this._spanContext.isSampled() || !this._spanContext.samplingFinished;
-    }
-
-    /**
      * Adds a set of tags to a span.
      *
      * @param {Object} keyValuePairs - An object with key value pairs
@@ -204,7 +218,7 @@ export default class Span {
             delete keyValuePairs[opentracing_tags.SAMPLING_PRIORITY];
         }
 
-        if (this.isSampledAndFinished()) {
+        if (this._isWriteable()) {
             for (let key in keyValuePairs) {
                 if (keyValuePairs.hasOwnProperty(key)) {
                     let value = keyValuePairs[key];
@@ -229,7 +243,7 @@ export default class Span {
             return this;
         }
 
-        if (this.isSampledAndFinished()) {
+        if (this._isWriteable()) {
             this._tags.push({'key': key, 'value': value});
         }
         return this;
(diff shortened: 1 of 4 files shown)


src/span.js
```diff
@@ -144,6 +144,17 @@ export default class Span {
      **/
     setOperationName(operationName: string): Span {
         this._operationName = operationName;
+        // If the span has debug set, then the question of resampling
+        // after setting the operation name is irrelevant.  The span is always
+        // sampled.
+        if (!this._spanContext.isDebug()) {
+            let sampler = this.tracer()._sampler;
+            var tags = {};  // TODO:oibe shouldn't this be initialized if not passed?
```
fastfrwrd: dumb lexical thing, but use `let` not `var` here.

src/span.js
```diff
@@ -168,6 +179,19 @@ export default class Span {
...
+     * Checks whether a span's sampling decision is finalized by calling setOperationName 
+     * This is used to hold onto tags, and logs, so that we can re-sample the span after
+     * an operation name is set on the span.  This is necessary because setting a sampling
+     * rate on the span before we know the operation name means adaptive-sampling
+     * does not know how to correlate the operation with sampling decisions.
+     *
+     * @return {boolean} - The decision about whether this span is done being sampled.
+     **/
+    isSampledAndFinished(): boolean {
+        return this._spanContext.isSampled() || !this._spanContext.samplingFinished;
```
black-adder: this logic is more like: is SampledOrFinished. even then !this._spanContext.samplingFinished is true when samplingFinished is false no?

src/span.js
```diff
@@ -168,6 +179,19 @@ export default class Span {
     }
 
     /**
+     * Checks whether a span's sampling decision is finalized by calling setOperationName
```
black-adder: some of this comment should be repurposed for the setOperationName function me thinks

test/span.js
```diff
@@ -182,6 +183,39 @@ describe('span should', () => {
         assert.isOk(unnormalizedKey in Span._getBaggageHeaderCache());
     });
 
+    describe('post trace sampling tests for span', () => {
```
black-adder: might want to rename because it's not really post trace sampling and when we finally do implement post trace sampling, this test name might get confusing

src/span.js
oibe: I'm confused by what your trying to say.

src/span.js
```diff
@@ -180,7 +204,7 @@ export default class Span {
             delete keyValuePairs[opentracing_tags.SAMPLING_PRIORITY];
         }
 
-        if (this._spanContext.isSampled()) {
+        if (this.isSampledAndFinished()) {
```
yurishkuro: 1. I suggest not using the word "finished" because it clashes with finishing the span. I suggest "final" as in "sampling decision is final / finalized'.
1. assuming "finished" here means the same as "final", then the logic is wrong. The right way would be "sampled OR finalized" (with a better naming).
1. for the purpose of these update functions (set()/log()), it is better to abstract away the question of sampling. For example, if you have a method `isWriteable()`, it becomes very single-purpose and obvious how to use here. The actual implementation of the method may be checking several conditions without having a weird method name. The span is writeable is if is either (1) sampled for sure (final), (2) maybe sampled, (3) not finish()'d.

src/span_context.js
```diff
@@ -33,6 +33,7 @@ export default class SpanContext {
     _flags: number;
     _baggage: any;
     _debugId: ?string;
+    _samplingFinished: boolean;
```
yurishkuro: * I think this field deserves a big comment explaining its semantics and various scenarios. 
* per previous comment, s/finished/finalized/
* the sampling decision must become final as soon as the span context is injected into something, or used as a reference for another span (maybe not for all references, but certainly for childOf and followsFrom).

yurishkuro:
I've made several comments that raise several scenarios this change currently does not account for. All those scenarios should have dedicated unit tests.

src/span.js
```diff
@@ -144,6 +144,17 @@ export default class Span {
      **/
     setOperationName(operationName: string): Span {
         this._operationName = operationName;
+        // If the span has debug set, then the question of resampling
+        // after setting the operation name is irrelevant.  The span is always
+        // sampled.
+        if (!this._spanContext.isDebug()) {
+            let sampler = this.tracer()._sampler;
+            var tags = {};  // TODO:oibe shouldn't this be initialized if not passed?
+            if (sampler.isSampled(operationName, tags)) {
```
yurishkuro: Here's a scenario for unit test - initial sampling set the tags based on default sampler, and re-sampling produces different sampling tags from a different type of sampler.

src/span.js
oibe: Its actually the name that is off.

src/span.js
oibe: actually changed my mind about this.

src/span.js
```diff
@@ -144,6 +144,17 @@ export default class Span {
      **/
     setOperationName(operationName: string): Span {
         this._operationName = operationName;
+        // If the span has debug set, then the question of resampling
+        // after setting the operation name is irrelevant.  The span is always
+        // sampled.
```
oibe: I would like to consider it user error to call setOperationName more than once.  What is a logical reason for why someone would need to call it more than once?

yurishkuro: I don't see any reason, but it doesn't mean we should not have a strategy for handling it. What does it mean to say "user error"? We're not going to throw an exception on it.

oibe: Whenever we call finish twice, we log a message.  We can do that here.  I think this case is the same as finish because even though some implementations can support the feature there is no logical reason to.

src/span_context.js
```diff
@@ -33,7 +33,20 @@ export default class SpanContext {
...
+     * correlated operation name -> sampling rate mapping, and when it cannot.
+     * Adaptive sampling uses the operation name of a span to correlate it with
+     * a sampling rate.  If an operation name is set on a span after the span's creation
+     * then adaptive sampling can no long connect the name to the proper sampling rate.
+     * In order to correct this we allow a span to be written to, so that we can re-sample
+     * it in the case that an operation name is set after span creation. Situations 
+     * where a span context's sampling decision is finalized include:
+     * 1.)  Finishing the span.
+     * 2.)  Using the span context as a parent when beginning a new span.
+     * 3.)  Setting the operation name on the span.
```
yurishkuro: missing use case: 4) span context is serialized using injectors

src/span_context.js
```diff
@@ -142,8 +155,8 @@ export default class SpanContext {
         this._debugId = debugId;
     }
 
-    set samplingFinished(finished: boolean) {
-        this._samplingFinished = finished;
+    set samplingFinalized(finished: boolean) {
```
yurishkuro: I do not believe there is a valid use case to set samplingFinalized to false. Until we have such use case, I suggest `finalizeSampling(): void` method instead of a property setter.

src/span.js
```diff
@@ -137,24 +137,51 @@ export default class Span {
...
+            return this;
+        }
+
         this._operationName = operationName;
         // If the span has debug set, then the question of resampling
         // after setting the operation name is irrelevant.  The span is always
         // sampled.
-        if (!this._spanContext.isDebug()) {
+
+        // Do not re-sample the span if:
```
yurishkuro: Let's re-write the comment without double negative: when do we DO re-sample?

src/span.js
```diff
@@ -137,24 +137,51 @@ export default class Span {
...
 
     /**
+     * Checks whether or not a span can be written to.
+     *
+     * @return {boolean} - The decision about whether this span can be written to.
+     **/
+    _isWriteable(): boolean {
+        // A span can NOT be written to if it:
+        // 1.) is finished.
+        // 2.) has been finalized, and is not sampled
```
yurishkuro: why not move this to the method comment? It could be part of the public docs

src/span.js
```diff
@@ -137,24 +137,51 @@ export default class Span {
...
-        if (!this._spanContext.isDebug()) {
+
+        // Do not re-sample the span if:
+        // 1.)  Debug has been set on the span.  Sampling is irrelevant in this case
+        //      the span will be reported no matter what.
+        // 2.)  The span is the first in process.  If the span is the first in process
+        //      then the sampling decision has already been made.
+        // 3.)  The span is not writable.  This can happen if the span's context was used
+        //      to create another span, or if the span is finished.
+        if (!this._spanContext.isDebug() || !this._firstInProcess) {
```
yurishkuro: I m not convinced about this condition. First let's rewrite the comment as suggested above. Then, why does "first in process" need to matter here? IIRC "first in process" was used in other clients to avoid sending process-level tags in the span. The new data model sends process-level tags at the batch level, so I am not sure fi we even need "first in process". Either way though, it is very different from the sampling - what we want for sampling is an indicator that the sampling decision has been inherited from the parent span. The easiest way to do that is probably to set `finalized` to true when creating such span.

src/span.js
```diff
@@ -137,24 +137,51 @@ export default class Span {
...
+    }
+
+    /**
      * Sets the operation name on this given span.
      *
      * @param {string} name - The name to use for setting a span's operation name.
      * @return {Span} - returns this span.
      **/
     setOperationName(operationName: string): Span {
+        if (!this._isWriteable()) {
```
yurishkuro: I am not sure about that. Say the span is not sampled but finalized because the sampling was inherited from the parent. Why shouldn't we still be able to override the operation name?

src/span.js
```diff
@@ -138,13 +138,15 @@ export default class Span {
 
     /**
      * Checks whether or not a span can be written to.
+     * A span can be written to if it:
+     *  - 'finish' has not been called on the span.
+     *  - the span is not finalized.
+     *  - the span has been finalized, but also is sampled.
+     *  - the span has not been serialized using injectors.
```
yurishkuro: unnecessary complex. Break into two parts:
  * span is finalized when either of these happens:
    * it had inherited sampling decision from its parent
    * it is finish()-ed
    * setOperationName is called
    * it it used as a parent for another span
    * its context is serialized using injectors

Then isWritable is simply `!finalized || sampled`

src/span.js
```diff
@@ -157,31 +159,24 @@ export default class Span {
      * @return {Span} - returns this span.
      **/
     setOperationName(operationName: string): Span {
-        if (!this._isWriteable()) {
+        // TODO:oibe is it wrong to allow this to be set even if we are finished?
+        this._operationName = operationName;
+        // We re-sample the span if all of the following conditions hold:
+        // 1.)  The span is not marked as a 'debug' span.
+        // 2.)  The span is not the first in the process.
+        // 3.)  '_isWriteable' returns true.
```
yurishkuro: I do not follow the logic of 2 and 3.

src/span.js
yurishkuro: (updated)

src/span.js
yurishkuro: shouldn't it be simply "re re-sample when it is not finalized"?

src/tracer.js
```diff
@@ -272,7 +271,14 @@ export default class Tracer {
             throw new Error(`Unsupported format: ${format}`);
         }
 
-        spanContext.samplingFinalized = true;
+        // Here because opentracing api compatibility tests can pass
+        // either span or span context.  Calling finalizeSampling below
+        // only works on a span context.
```
yurishkuro: irrelevant comment. OT API allows span instance to be passed into inject(), so this takes care of that. Has nothing to do with compatibility tests or finalizeSampling

src/tracer.js
```diff
@@ -292,7 +298,13 @@ export default class Tracer {
             throw new Error(`Unsupported format: ${format}`);
         }
 
-        return extractor.extract(carrier);
+        let spanContext = extractor.extract(carrier);
+
+        // spanContext can be null if using a differnt type of extractor. i.e. BinaryCodec
```
yurishkuro: irrelevant comment. Span context can be null purely because there was no context in the carrier.

test/span.js
```diff
@@ -217,6 +218,18 @@ describe('span should', () => {
...
+            let headers = {};
+            tracer.inject(span.context(), opentracing.FORMAT_HTTP_HEADERS, headers);
+
+            assert.isOk(span.context().samplingFinalized);
+        });
+
+        it ('a child span is finalized', () => {
+            let childSpan = tracer.startSpan('finalized', { childOf: span.context() });
+            assert.isOk(childSpan.context().samplingFinalized);
+        });
```
yurishkuro: not all use cases covered.

src/span_context.js
```diff
@@ -33,6 +33,21 @@ export default class SpanContext {
...
+     * correlated operation name -> sampling rate mapping, and when it cannot.
+     * Adaptive sampling uses the operation name of a span to correlate it with
+     * a sampling rate.  If an operation name is set on a span after the span's creation
+     * then adaptive sampling can no long connect the name to the proper sampling rate.
+     * In order to correct this we allow a span to be written to, so that we can re-sample
+     * it in the case that an operation name is set after span creation. Situations 
+     * where a span context's sampling decision is finalized include:
+     * 1.)  Finishing the span.
+     * 2.)  Extracting the span context from the wire format.
+     * 3.)  A span that was created wth a parent span context.
```
black-adder: s/wth/with

test/span.js
```diff
@@ -182,6 +186,121 @@ describe('span should', () => {
...
+                span.finish();
+
+                assert.isOk(deepEqual(span._tags[0], {key: 'tagKeyOne', value: 'tagValueOne'}));
+                assert.isOk(deepEqual(span._tags[1], {key: 'tagKeyTwo', value: 'tagValueTwo'}));
+                assert.isOk(deepEqual(span._logs[0].fields[0], {key: 'logkeyOne', value: 'logValueOne'}));
+                assert.equal(reporter.spans.length, o.reportedSpans);
+            });
+        });
+
+        describe('Finalize spans that', () => {
```
black-adder: Is there a way to test whether a span is finalized after it is used as a parent for another span?

src/span.js
```diff
@@ -244,6 +274,7 @@ export default class Span {
     _setSamplingPriority(priority: number): void {
         if (priority > 0) {
             this._spanContext.flags = this._spanContext.flags | constants.SAMPLED_MASK | constants.DEBUG_MASK;
+            this._spanContext.finalizeSampling();
```
yurishkuro: applies to both then/else parts.

src/span.js
```diff
@@ -137,13 +137,43 @@ export default class Span {
...
 
     /**
+     * Checks whether or not a span can be written to.
+     * A span is finalized when any of these happens:
+     * - it has inherited the sampling decision from its parent
+     * - its debug flag is set via the sampling.priority tag
+     * - it is finish()-ed
+     * - setOperationName is called
+     * - it is used as a parent for another span
+     * - its context is serialized using injectors
```
yurishkuro: the definition of what "finalized" means does not belong to this method's comment. It should be moved to the field definition in span context (where it is currently incomplete).

src/tracer.js
```diff
@@ -286,7 +298,12 @@ export default class Tracer {
             throw new Error(`Unsupported format: ${format}`);
         }
 
-        return extractor.extract(carrier);
+        let spanContext = extractor.extract(carrier);
+
+        if (spanContext) {
+            spanContext.finalizeSampling();
+        }
```
yurishkuro: a bit odd having this here, but ok.

test/span.js
```diff
@@ -182,6 +186,121 @@ describe('span should', () => {
...
+            it ('inherit sampling decision is finalized', () => {
```
yurishkuro: The naming of these tests is very arbitrary. describe/it pattern is supposed to read like natural language, i.e `describe` refers to the subject and its condition under test, and `it` refers to the expectation "it should ...".

yurishkuro: For example, this test should be split and read like this:
  * describe span used as a parent
    * it should be finalized
  * describe span that is a child
    * it should be finalized

yurishkuro: In case this creates a lot of single-entry groups, `it()` can be omitted

`describe("span used as a parent should be finalized") {}`

This reads naturally both in the source code and in the test output

test/span.js
```diff
@@ -182,6 +186,121 @@ describe('span should', () => {
...
+            tracer = new Tracer(
+                'test-service-name',
+                new InMemoryReporter(),
+                new ConstSampler(false),
+                { logger: new MockLogger() }
+            );
+            let unFinalizedSpan = tracer.startSpan('unFinalizedSpan');
+
+            assert.isOk(unFinalizedSpan._isWriteable());
+            assert.isOk(span._isWriteable());
```
yurishkuro: hard to follow - at least add a comment explaining that setup creates `span` as sampled, which is what you're relying on here.

yurishkuro: Actually even that is not a true proof - you need to finalize the `span` first to ensure that part of the check is not firing.

test/span.js
```diff
@@ -182,6 +186,121 @@ describe('span should', () => {
...
+            span.setOperationName('re-sampled-span');
+
+            assert.equal(span.operationName, 're-sampled-span');
+            assert.isOk(JaegerTestUtils.hasTags(span, {
+                'sampler.type': 'probabilistic',
+                'sampler.param': 1
+            }));
+        });
+
+        it ('setOperationName should not change the sampling tags, but should change the operationName', () => {
```
yurishkuro: you mean 2nd setOperationName

test/span.js
oibe: They are more less the same thing.  Technically you can have an "invalid" parent span and it won't inherit the sampling decision.  In that case the parent does not get finalized.

test/span.js
oibe: I strongly disagree with this.  To me the purpose of "describe" and "it" isn't to read like a literal sentence, but rather to "describe" exists to group related functionality.  Instead of repeating "it should be finalized" multiple times it makes sense to "describe" the types of spans that should be finalized.  Above if you read the tests like "describe span used as a parent it should be finalized"  that already reads awkwardly.  Furthermore you say that the "it() can be omitted", but when you emit the it then the code in the describe blocks no long takes advantage of "beforeEach", and "afterEach" code, which makes things a lot more redundant.

yurishkuro: > To me the purpose of "describe" and "it" isn't to read like a literal sentence, 

the frameworks built to support BDT have certain conventions about how the tests should be written. This describe/it convention is nearly uniform across many languages. Your argument is essentially to abuse the framework by describing tests in a completely different way, which as a result reads completely weird in the source code. Why use a BDT notation in that case?

> "describe span used as a parent it should be finalized"

yes, because it's not meant to be read as a single line, but as a group and items:
  * describe "span used as a parent"
     * it "should be finalized"
     * it "should do something else"

yurishkuro: another way you can adopt to the framework is by changing what the subject is, e.g. 
  * describe(span sampling finalizer)
     * it(should trigger on span used as parent)

test/span.js
```diff
@@ -182,6 +186,127 @@ describe('span should', () => {
...
+            );
+            let unFinalizedSpan = tracer.startSpan('unFinalizedSpan');
+            assert.isOk(unFinalizedSpan._isWriteable());
+
+            tracer._sampler = new ConstSampler(true);
+            let sampledSpan = tracer.startSpan('sampled-span');
+            assert.isOk(sampledSpan._isWriteable());
+        });
+
+        it ('2nd setOperationName should add sampler tags to span, and change operationName', () => {
```
oibe: I meant 2nd because the I want to make sure on the 2nd set operationName that the tags change, and the operation name changes.

src/tracer.js
```diff
@@ -267,7 +272,14 @@ export default class Tracer {
             throw new Error(`Unsupported format: ${format}`);
         }
 
-        injector.inject(spanContext, carrier);
+        if (spanContext.context) {
+            spanContext = spanContext.context();
+        }
+
+        if (spanContext) {
```
oibe: That might be true, but opentracing compatibility passes null into this function, so we would have to change that first.

test/span.js
```diff
@@ -182,6 +186,147 @@ describe('span should', () => {
...
+                assert.isOk(deepEqual(span._tags[0], {key: 'tagKeyOne', value: 'tagValueOne'}));
+                assert.isOk(deepEqual(span._tags[1], {key: 'tagKeyTwo', value: 'tagValueTwo'}));
+                assert.isOk(deepEqual(span._logs[0].fields[0], {key: 'logkeyOne', value: 'logValueOne'}));
+                assert.equal(reporter.spans.length, o.reportedSpans);
+            });
+        });
+
+        describe('span sampling finalizer', () => {
+            it ('should trigger when it inherits a sampling decision', () => {
+                // Span created in before each is not finalized.
```
yurishkuro: nits: 
  * usually methods like `assert.equal` take a message as the 3rd argument, so the text of the comment is better placed there, to serve both as documentation and a better error should the test fail.
  * s/before each/beforeEach()/

src/tracer.js
```diff
@@ -267,7 +270,14 @@ export default class Tracer {
             throw new Error(`Unsupported format: ${format}`);
         }
 
-        injector.inject(spanContext, carrier);
+        if (spanContext && spanContext.context) {
```
yurishkuro: > That might be true, but opentracing compatibility passes null into this function, so we would have to change that first.

In this case we can do the check at the beginning of the method and throw exception, rather than keep checking throughout the method.

yurishkuro:
Fixes #80

PR #86, opened 2017-01-11
node test client for jaeger integration tests

oibe:


black-adder:
Let's try to get all the clients in sync: https://github.com/uber/jaeger-client-go/pull/71

No init endpoint (sorrrrry) and lazy loading of tracer when "/create_spans" is called for the first time to avoid the issue when the test_driver address can't be resolved. Also, new endpoint in the existing crossdock http server rather than a new independent server.

PR #89, opened 2017-01-27
Add ip tag to client

black-adder:


src/test_util.js
```diff
@@ -69,19 +69,18 @@ export default class TestUtils {
         return span._duration === duration;
     }
 
-    static hasTags(span: Span, tags: any): boolean {
+    static hasTags(span: Span, expectedTags: any): boolean {
         // TODO(oibe) make this work for duplicate tags
-        let expectedTags = {};
+        let actualTags = {};
```
yurishkuro: thank you!

test/tchannel_bridge.js
```diff
@@ -122,7 +122,6 @@ describe ('test tchannel span bridge', () => {
                     let serverSpanTags = {};
                     serverSpanTags[opentracing.Tags.PEER_SERVICE] = 'echo';
                     serverSpanTags[opentracing.Tags.SPAN_KIND] = opentracing.Tags.SPAN_KIND_RPC_SERVER;
-                    serverSpanTags[opentracing.Tags.PEER_HOST_IPV4] = ((127 << 24) | 1);
```
yurishkuro: why do we remove this?

test/tracer.js
```diff
@@ -104,10 +104,11 @@ describe('tracer should', () => {
...
         assert.equal(actualTags[1].key, 'jaeger.version');
-        assert.equal(actualTags[2].key, 'sampler.param');
-        assert.equal(actualTags[3].key, 'sampler.type');
-        assert.equal(actualTags[2].value, true);
-        assert.equal(actualTags[3].value, 'const');
+        assert.equal(actualTags[2].key, 'peer.ipv4');
+        assert.equal(actualTags[3].key, 'sampler.param');
+        assert.equal(actualTags[4].key, 'sampler.type');
+        assert.equal(actualTags[3].value, true);
+        assert.equal(actualTags[4].value, 'const');
```
yurishkuro: why don't we use that hasTags util function here?

test/tchannel_bridge.js
black-adder: This check will essentially become Utils.ip() == Utils.ip(). I'm not sure how the host ipv4 was hardcoded before...

black-adder: Actually, what happened is that the tracer added the peer host tag after the tchannel bridge added it. The tchannel bridge ip was set to 127.0.0.1 but the tracer ip is dynamic.

test/tracer.js
black-adder: The values for these are dynamic and the helper function checks the value, meh I guess I can write a helper function that doesn't care if the values don't match but uhhhh, too lazy.

src/tchannel_bridge.js
```diff
@@ -87,7 +87,6 @@ export default class TChannelBridge {
             span.setTag(opentracing.Tags.PEER_SERVICE, request.callerName);
             let hostPort: Array<string> = request.remoteAddr.split(':');
             if (hostPort.length == 2) {
-                span.setTag(opentracing.Tags.PEER_HOST_IPV4, Utils.ipToInt(hostPort[0]));
```
black-adder: This guy is overwritten at a later point by the tracer

PR #98, opened 2017-02-08
Make sure initTracer passes options to the tracer

yurishkuro:


test/init_tracer.js
```diff
@@ -117,9 +117,23 @@ describe('initTracer', () => {
...
+        let tracer = initTracer({
+            serviceName: 'test-service'
+        }, {
+            logger: 'some logger',
+            metrics: 'some metrics',
+            tags: {
+                'x': 'y'
+            }
+        });
+        assert.equal(tracer._logger, 'some logger');
```
black-adder: dynamic typing ftw

yurishkuro: yeah, it has its moments

PR #103, opened 2017-03-04
Remove dependency on 'request' module, use 'http' instead

yurishkuro:


src/samplers/remote_sampler.js
```diff
@@ -101,38 +101,54 @@ export default class RemoteControlledSampler {
...
+        http.get({
+            'host': this._host,
+            'port': this._port,
+            'path': `/sampling?service=${serviceName}`
+        }, (res) => {
+            // explicitly treat incoming data as utf8 (avoids issues with multi-byte chars)
+            res.setEncoding('utf8');
+
+            // incrementally capture the incoming response body
+            let body = '';
```
black-adder: does node have a string buffer that we can use?

yurishkuro: Considering that there's even a custom module for that (https://github.com/raypulver/string-buffer), I am guessing there isn't a std string buffer. There's Buffer, but it works on bytes.

Given the frequency of polling for the sampling rate I don't think the gc pressure matters here.

PR #104, opened 2017-03-06
Fix leaky bucket bug

black-adder:
Fix bug where if the creditsPerSecond < itemCost, the leaky bucket never fills up. Added a maxBalance to set the upperbound of the balance.

### DIFF
--- a/src/rate_limiter.js
+++ b/src/rate_limiter.js
@@ -27,7 +27,7 @@ export default class RateLimiter {
 
     constructor(creditsPerSecond: number, maxBalance: number) {
         this._creditsPerSecond = creditsPerSecond;
-        this._balance = creditsPerSecond;
+        this._balance = maxBalance;
         this._maxBalance = maxBalance;
         this._lastTick = new Date().getTime();
     }
--- a/src/rate_limiter.js
+++ b/src/rate_limiter.js
@@ -22,11 +22,13 @@
 export default class RateLimiter {
     _creditsPerSecond: number;
     _balance: number;
+    _maxBalance: number;
     _lastTick: number;
 
-    constructor(creditsPerSecond: number) {
+    constructor(creditsPerSecond: number, maxBalance: number) {
         this._creditsPerSecond = creditsPerSecond;
         this._balance = creditsPerSecond;
+        this._maxBalance = maxBalance;
         this._lastTick = new Date().getTime();
     }
 
@@ -36,10 +38,9 @@ export default class RateLimiter {
         this._lastTick = currentTime;
 
         this._balance += elapsedTime * this._creditsPerSecond;
-        if (this._balance > this._creditsPerSecond) {
-            this._balance = this._creditsPerSecond;
+        if (this._balance > this._maxBalance) {
+            this._balance = this._maxBalance;
         }
-
         if (this._balance >= itemCost) {
             this._balance -= itemCost;
             return true;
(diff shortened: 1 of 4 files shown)


src/samplers/ratelimiting_sampler.js
```diff
@@ -32,7 +32,7 @@ export default class RateLimitingSampler {
         }
 
         this._maxTracesPerSecond = maxTracesPerSecond;
-        this._rateLimiter = new RateLimiter(maxTracesPerSecond);
+        this._rateLimiter = new RateLimiter(maxTracesPerSecond, 2.0);
```
yurishkuro: why `2.0`? If the `RateLimitingSampler` itself does not take this as argument, then `maxBalance` should probably be somehow calculated from `maxTracesPerSecond`

black-adder: The max balance should equal the checkCredit value (1.0) in this case. But I set it to 2 so that all existing tests can work with all the edge cases which unfortunately 1.0 doesn't hit. The only difference in practice is that we buffer up to 2 traces which isn't toooooo bad.

yurishkuro: but the previous behavior was `this._balance > this._creditsPerSecond`, so wouldn't it be backwards compatible to use `new RateLimiter(maxTracesPerSecond, maxTracesPerSecond)`? Otherwise if maxTracesPerSecond=10, you can never actually achieve that limit since max balance is just 2

black-adder: This is supposed to not be backwards compatible. Given the new rate limit is 0.001666 (once every 10 mins), if we set maxbalance to maxtracespersecond, we'll never surpass the checkCredit value 1.0. Given that we decide what maxTracesPerSecond is, I don't think this is an issue.

src/rate_limiter.js
```diff
@@ -22,11 +22,13 @@
 export default class RateLimiter {
     _creditsPerSecond: number;
     _balance: number;
+    _maxBalance: number;
     _lastTick: number;
 
-    constructor(creditsPerSecond: number) {
+    constructor(creditsPerSecond: number, maxBalance: number) {
         this._creditsPerSecond = creditsPerSecond;
         this._balance = creditsPerSecond;
```
yurishkuro: should this be set to maxBalance instead?

src/samplers/ratelimiting_sampler.js
yurishkuro: I am not convinced. Currently [this test](https://github.com/uber/jaeger-client-node/blob/master/test/samplers/ratelimiting_sampler_test.js#L28) is incorrect because (a) initially it succeeds due to constructor bug(?) I pointed out, and after timer is advanced it succeeds because we're only doing a single check. If we change it to 9-10 checks, it will fail with your change.

Remember that RateLimitingSampler was originally designed for N > 1 traces per second. If you want it to work for N << 1, then let's test both scenarios.

test/rate_limiter.js
```diff
@@ -43,7 +43,7 @@ describe ('leaky bucket ratelimiter should', () => {
         let clock = sinon.useFakeTimers(initialDate);
```
yurishkuro: I was referring to L37 when I said the test is misleading. Let's do a loop to 10 there, like in L30. Also, L31 should probably be an assert.

PR #105, opened 2017-03-28
Protect from exceptions in decodeURIComponent; do not url-encode span context

yurishkuro:
decodeURIComponent can throw `URIError: URI malformed` when unable to parse the string.

The string representation of the context is #:#:#:# (# is hex digits), so it is "safe" to be used as is in the http header value without URL encoding.

### DIFF
--- a/src/propagators/text_map_codec.js
+++ b/src/propagators/text_map_codec.js
@@ -40,7 +40,7 @@ export default class TextMapCodec {
         this._metrics = options.metrics || new Metrics(new NoopMetricFactory());
     }
 
-    _encodedValue(value: string): string {
+    _encodeValue(value: string): string {
         if (this._urlEncoding) {
             return encodeURIComponent(value);
         }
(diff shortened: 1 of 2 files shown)


test/propagators.js
```diff
@@ -0,0 +1,59 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import {assert} from 'chai';
+import TextMapCodec from '../src/propagators/text_map_codec';
+import SpanContext from '../src/span_context';
+
+describe ('TextMapCodec', () => {
+    it('should not URL-decode value that has no % meta-characters', () => {
```
vprithvi: This test is insufficient because `decodeURIComponent(abc)` returns `abc`. We should explicitly verify that `decodeURIComponent('abc')` has never been called.

yurishkuro: good point, but I am not sure how to do that. @saminzadeh is there a way to mock standard functions?

src/propagators/text_map_codec.js
```diff
@@ -49,8 +49,14 @@ export default class TextMapCodec {
     }
 
     _decodedValue(value: string): string {
-        if (this._urlEncoding) {
-            return decodeURIComponent(value);
+        // only use url-decoding if there are meta-characters '%'
+        if (this._urlEncoding && value.indexOf('%') > -1) {
+            // unfortunately, decodeURIComponent() can throw 'URIError: URI malformed'
+            try {
+                return decodeURIComponent(value);
```
yurishkuro: @saminzadeh do you know if there's an alternative std function that never throws exceptions?

PR #106, opened 2017-03-29
Add support for deferred sampling

vprithvi:
- When a downstream service doesn't have enough information to
  make a sampling decision, it sets the span context flag 4.

- When a span with the flag 4 is encountered, a sampling decision
  for any child spans is made based on the current operation name
  instead of depending on the sampling decision on the parent if
  the span context doesn't contain the debug flag(2). Debug spans
  are always sampled.

- set the tag deferredsampling.status to indicate that the sampling
  decision was taken at this span


src/tracer.js
```diff
@@ -231,13 +231,24 @@ export default class Tracer {
...
             ctx.baggage = parent.baggage;
 
-            parent.finalizeSampling();
-            ctx.finalizeSampling();
+            // Parent exists without a sampling decision
+            // An example of where this happens is in mobile clients where sampler parameters
+            // are not propagated
+            if ((parent.flags & constants.DEFERRED_SAMPLING_MASK)
+                === constants.DEFERRED_SAMPLING_MASK) {
+                ctx.flags = parent.flags ^ constants.DEFERRED_SAMPLING_MASK
```
black-adder: I think we need to remove the deferred_sampling_mask in the parent.flag too and also set the sampled_mask on it too if it should be sampled. This is because the same parent context can be used to create multiple spans. We don't want to run this logic multiple times, only once on the first span using this parent span. Future spans will just respect the parent context which has been modified correctly.

src/tracer.js
```diff
@@ -231,13 +231,24 @@ export default class Tracer {
...
+                if (this._sampler.isSampled(operationName, internalTags)) {
+                    ctx.flags = parent.flags |= constants.SAMPLED_MASK
```
black-adder: I prefer if we don't use parent.flags here because all its pertinent information has already been copied into ctx.flags.

src/tracer.js
```diff
@@ -231,13 +231,24 @@ export default class Tracer {
...
+                    internalTags[constants.DEFERRED_SAMPLING_STATUS] = true
+                }
+            } else {
+                ctx.flags = parent.flags
+                parent.finalizeSampling();
```
black-adder: This also needs to be called in the if case above.

test/span.js
```diff
@@ -336,3 +336,64 @@ describe('span should', () => {
     // TODO(oibe) need tests for standard tags, and handlers
 });
 
+describe('span with deferred sampling flag', () => {
```
black-adder: add a test where two spans are created from one deferred parent span.

src/tracer.js
```diff
@@ -231,13 +231,24 @@ export default class Tracer {
...
-
             // reuse parent's baggage as we'll never change it
             ctx.baggage = parent.baggage;
 
-            parent.finalizeSampling();
-            ctx.finalizeSampling();
+            // Parent exists without a sampling decision
+            // An example of where this happens is in mobile clients where sampler parameters
+            // are not propagated
+            if ((parent.flags & constants.DEFERRED_SAMPLING_MASK)
```
black-adder: nit: can we simply do
`if (parent.flags & constants.DEFERRED_SAMPLING_MASK)`
because all we need to do is confirm flag & mask is not 0.

src/constants.js
```diff
@@ -24,12 +24,18 @@ export const SAMPLED_MASK = 0x1;
...
+export const DEFERRED_SAMPLING_MASK = 0x4;
+
 // JAEGER_CLIENT_VERSION_TAG_KEY is the name of the tag used to report client version.
 export const JAEGER_CLIENT_VERSION_TAG_KEY = 'jaeger.version';
 
 // TRACER_HOSTNAME_TAG_KEY used to report host name of the process.
 export const TRACER_HOSTNAME_TAG_KEY = 'jaeger.hostname';
 
+// DEFERRED_SAMPLING_STATUS indicated that deferred sampling was performed at this span.
+export const DEFERRED_SAMPLING_STATUS = 'deferredsampling.status'
```
vprithvi: How about `sampler.deferred`, which makes it consistent with the other sampler tags. Essentially this indicates that sampling has been deferred until now

src/tracer.js
vprithvi: Done

src/constants.js
```diff
@@ -24,12 +24,18 @@ export const SAMPLED_MASK = 0x1;
 // DEBUG_MASK is the bit mask indicationg that a span has been marked for debug.
 export const DEBUG_MASK = 0x2;
 
+// DEFERRED_SAMPLING_MASK is the bit mask indicating that a span has been marked for deferred sampling.
```
yurishkuro: better to elaborate what deferred sampling means

src/constants.js
yurishkuro: first, if we want to use this as a tag, then the const naming pattern is XXX_TAG_KEY (see  JAEGER_CLIENT_VERSION_TAG_KEY, and its description).

But I don't see much need for this tag in the first place. What is the intended use case? The "deferred sampling was performed at this span" sounds somewhat strange to me, because "deferring" the sampling decision actually happened on the mobile side.

vprithvi: I was thinking that adaptive sampling might have to account for deferred sampling traffic separately because the sampling decision is taking place on a non-root span. 
Essentially, I wanted to indicate that the deferred sampling decision was converted to a concrete sampling decision at this span.

src/tracer.js
vprithvi: Done

src/constants.js
black-adder: I think the plan is to use the sampler.type / sampler.param flags instead to determine whether this span was sampled rather than using root span. This way we don't need this extra tag (I'm basing this off of the discussion we had in the mobile RFC).

However, I do like having this extra tag even if it's not useful for adaptive sampling. For audit purposes.

test/span.js
vprithvi: Done

src/constants.js
vprithvi: Done

src/tracer.js
vprithvi: Agree about running the same thing multiple times, but as Yuri said, I think that we shouldn't be modifying the parent span because it was created elsewhere.

src/tracer.js
vprithvi: Why is that? How do we know that `setOperation` isn't going to be called on the parent?
I think makes sense in the else, because a sampling decision is already made.

src/tracer.js
```diff
@@ -231,13 +231,23 @@ export default class Tracer {
...
             // reuse parent's baggage as we'll never change it
             ctx.baggage = parent.baggage;
+            ctx.flags = parent.flags;
 
-            parent.finalizeSampling();
-            ctx.finalizeSampling();
+            // Parent exists without a sampling decision
+            // An example of where this happens is in mobile clients where sampler parameters
+            // are not propagated
+            if (parent.flags & constants.DEFERRED_SAMPLING_MASK) {
```
yurishkuro: ctx.flags ?

src/tracer.js
```diff
@@ -231,13 +231,23 @@ export default class Tracer {
...
+                ctx.flags ^=  constants.DEFERRED_SAMPLING_MASK;
+                if (this._sampler.isSampled(operationName, internalTags)) {
+                    ctx.flags |= constants.SAMPLED_MASK;
+                    internalTags[constants.DEFERRED_SAMPLING_TAG_KEY] = true;
+                }
+            } else {
+                parent.finalizeSampling();
+                ctx.finalizeSampling();
```
yurishkuro: I would like to consider whether "finalized" and "deferred" are actually the same thing.

src/tracer.js
black-adder: This still doesn't solve what happens if the parent context creates more than 1 child span. The tracer might be using a ratelimiting sampler which means that sampler.isSampled might return different results for 2 spans in the same trace which is not what we want.

vprithvi: Done

src/tracer.js
vprithvi: I don't think they are the same thing. 

I was also looking at @black-adder's comment about this scheme not working with the rate limiting sampler (or in the more general case, any sampler that isn't consistent). The previous suggestion of modifying the parent span solves this problem. 

The `finalizeSampling` mechanism resets the sampling decision by calling `IsSampled` the first time `setOperation` is called.  Any subsequent `setOperation` calls are ignored.

Further, the `finalizeSampling` mechanism is only invoked for new root spans. (Anything extracted/injected is finalized in the extractor/injector)

I feel that @black-adder's suggestion might be the cleanest way of sampling non root spans in the middle of a trace.

yurishkuro: You are referring to the _current implementation_ of finalized sampling. I want to step back and consider what the actual difference is between deferred and not finalized - even from their name they appear to be very similar. The only difference I currently see is deferred works across services, while not-finalized within a single span. But I do think we can generalize this to a single concept and simplify the code.

Ultimately they both defer the sampling decision until some event later in the execution.

src/constants.js
yurishkuro: the first sentence is a bit of a tautology. 

DEFERRED_SAMPLING_MASK is the bit mask indicating that the upstream service that generated the span did not make a definitive sampling decision, but deferred it to be made by some downstream service.

No need to mention mobile, imo, especially since the claim that we cannot have a feedback loop with them is not exactly true.

src/tracer.js
```diff
@@ -228,6 +228,15 @@ export default class Tracer {
             ctx.parentId = null;
             ctx.flags = flags;
         } else {
+            if (!parent.samplingFinalized) {
+                if (this._sampler.isSampled(operationName, internalTags)) {
+                   parent._flags |= constants.SAMPLED_MASK
+                } else {
+                    parent._flags &= ~constants.SAMPLED_MASK
+                }
+                parent.finalizeSampling()
```
yurishkuro: don't like this. Parent context should be treated as immutable. Often it will be deserialized from the wire, so trying to modify it is misleading at best.

src/tracer.js
```diff
@@ -209,7 +209,7 @@ export default class Tracer {
         let internalTags: any = {};
         if (!parent || !parent.isValid) {
```
yurishkuro: I think this could be refactored to avoid repetition. For example, the invalid parent behavior can be done upfront:

```
if (parent && !parent.isValid()) {
   // ...
   parent = null;
}
```

then create new Context and update it depending on parent == null

then check if context is deferred and make sampling decision

src/tracer.js
```diff
@@ -297,12 +305,7 @@ export default class Tracer {
             throw new Error(`Unsupported format: ${format}`);
         }
 
-        let spanContext = extractor.extract(carrier);
-
-        if (spanContext) {
-            spanContext.finalizeSampling();
-        }
-        return spanContext;
+        return extractor.extract(carrier);
```
yurishkuro: hm, not sure why it was doing that before. Has any test broke because of this?

vprithvi: No, none of the tests break because of this. In this implementation,  finalizing a span removes the deferred sampling flag, so finalizing a span before we apply a sampling decision is problematic.

src/tracer.js
vprithvi: We could have another boolean in the parent span to store the state of whether all child spans must be sampled and update that instead of the parent span context. Would you prefer the implementation to do that?

src/constants.js
vprithvi: I'll update the first sentence and remove the example about mobile. 
However, I haven't claimed that such a feedback loop cannot exist.

yurishkuro: > because adaptive sampling parameters cannot be reliably pushed to mobile devices

src/tracer.js
yurishkuro: we do not have access to parent span when creating a child span, we only have access to its context

src/constants.js
vprithvi: Well, that statement is akin to saying that the network isn't reliable, and thus the feedback loop might not be reliable as well. It does not claim that we cannot have a feedback loop.

src/tracer.js
vprithvi: It becomes a bit messy with handling the debug spans as well. I'd prefer this refactor be a part of a separate commit.

src/tracer.js
vprithvi: Indeed, in which case modifying the parent seems reasonable.

src/span_context.js
```diff
@@ -126,8 +108,25 @@ export default class SpanContext {
...
+     * - it has inherited the sampling decision from its parent
+     * - its debug flag is set via the sampling.priority tag
+     * - it is finish()-ed
+     * - setOperationName is called
+     * - it is used as a parent for another span
+     * - its context is serialized using injectors
+     * */
     get samplingFinalized(): boolean {
-        return this._samplingFinalized;
+        return ((this.flags & constants.DEFERRED_SAMPLING_MASK) !== constants.DEFERRED_SAMPLING_MASK)
```
yurishkuro: - outer parentheses not needed
- missing `;`

src/span_context.js
```diff
@@ -162,7 +161,7 @@ export default class SpanContext {
     }
 
     finalizeSampling(): void {
-        this._samplingFinalized = true;
+        this.flags &= ~constants.DEFERRED_SAMPLING_MASK
```
yurishkuro: add `;`

src/span.js
```diff
@@ -118,12 +117,21 @@ export default class Span {
...
     context(): SpanContext {
-        return this._spanContext;
+        let ctx = this._spanContext;
+        if (!ctx.samplingFinalized) {
+            if (ctx.isDebug() || this._tracer._sampler.isSampled(this.operationName, this._tags)) {
+                ctx._flags |= constants.SAMPLED_MASK
+            } else {
+                ctx._flags &= ~constants.SAMPLED_MASK
+            }
+            ctx.finalizeSampling()
```
yurishkuro: add `;` everywhere

src/tracer.js
```diff
@@ -228,6 +228,7 @@ export default class Tracer {
             ctx.parentId = null;
             ctx.flags = flags;
         } else {
+            this.processDeferredSampling(parent, operationName, internalTags);
```
yurishkuro: I think this should be called after L235 and operate on `ctx`, not on `parent`.

vprithvi: The problem with the approach you suggest is that we can no longer guarantee the consistency of the sampling decision for all children (because each child has an independent sampling decision). We need to store the sampling decision for all future spans in the parent span context object or something associated with it to ensure all child spans have the same sampling decision applied to them.

I don't think that this is a easy problem to solve neatly while most of our stack assumes that sampling decisions happen when the span context is initially created.

yurishkuro: Since we narrowed the scope, would't the `parent` in this case always be something we just serialized from the wire? The _only_ child is going to be the server-side span.

And as a more general principle, if the parent wants to ensure that its sampling decision is fixed and should affect all children, then it's he parent who should make the change, not as a side effect of creating a child.

vprithvi: >Since we narrowed the scope, would't the parent in this case always be something we just serialized from the wire? The only child is going to be the server-side span.
You are right, I'll have an updated diff out shortly

PR #110, opened 2017-04-18
Fix strategyType

black-adder:
The thrift0.9.2 sampling endpoint returns strategyType as an integer but thrift0.9.3 sampling endpoint returns strategyType as a string.

This fixes the remote sampler to use the string strategyType.


black-adder:
We should be converting the JSON received from the agent to an internal thrift model like we do in go rather than doing string/integer comparisons like this. (I'll have to investigate if it's doable in node)

test/samplers/remote_sampler.js
```diff
@@ -114,7 +114,7 @@ describe('RemoteSampler', () => {
             remoteSampler._refreshSamplingStrategy();
         };
         server.addStrategy('service1', {
-            strategyType: 0,
+            strategyType: 'PROBABILISTIC',
```
oibe: why not import these from remote_sampler.js?

test/samplers/remote_sampler.js
```diff
@@ -139,7 +139,7 @@ describe('RemoteSampler', () => {
 
     it('should set per-operation sampler', (done) => {
         server.addStrategy('service1', {
-            strategyType: 0,
+            strategyType: 'PROBABILISTIC',
```
oibe: Same as comment above.

test/samplers/remote_sampler.js
yurishkuro: Because it makes the test less strict

src/configuration.js
```diff
@@ -155,16 +152,18 @@ export default class Configuration {
      *        as process-level tags on the Tracer itself.
      */
     static initTracer(config, options = {}) {
-        let reporters = [];
         let reporter;
         let sampler;
+        if (options.metrics) {
+            options.metrics = new Metrics(options.metrics);
```
yurishkuro: Isn't it possible that Metrics type is already passed in?

black-adder: hypothetically they can pass in whatever they want given it's javascript. We do have a comment above that says "metrics factory"

PR #108, opened 2017-04-06
Release 3.6.0

vprithvi:

PR #111, opened 2017-04-19
Release v3.6.0

black-adder:


yurishkuro:
This is now a duplicate of #108, and I have the same concern, we need to test deferred sampling before we release.

black-adder:
lol, should've checked before I went on this escapade. OK, let's close that one and wait until it's fully tested before landing this one.

PR #113, opened 2017-04-24
Revert "Add support for deferred sampling (#107)"

black-adder:
This reverts commit caac859468d29db6932a4923f5bb6ca2cd5513f0.

Reverting deferred sampling to release a bug fix.


yurishkuro:
Can we instead add a boolean param to the tracer that will enable the deferred sampling behavior? The code seems to be fairly isolated, so we can just exit from `processDeferredSampling` if the flag is not set, then there will be no impact, yet all the code will remain in place and we'll just need to take out the config param in the future (or make it default to enabled).

black-adder:
Yes, we can do it but in another diff. I just wanna get my fix out for now.

PR #119, opened 2017-05-08
Add short circuiting

black-adder:


test/samplers/guaranteed_throughput_sampler.js
```diff
@@ -69,53 +69,76 @@ describe('GuaranteedThroughput sampler', () => {
...
+    let assertValues = function assertValues(sampler, rate, lb, up) {
         assert.equal(rate, sampler._probabilisticSampler.samplingRate);
+        assert.equal(lb, sampler._lowerBoundSampler.maxTracesPerSecond);
+        assert.equal(up, sampler._upperBoundRateLimiter.creditsPerSecond);
     };
 
     it('should not change when update() called with the same values', () => {
-        let sampler = new GuaranteedThroughputSampler(2, 1.0);
-        assertValues(sampler, 2, 1.0);
+        let sampler = new GuaranteedThroughputSampler(1.0, 2, 3);
```
oibe: When I look at a lot of this code i'm not sure what the parameters mean for GuaranteedThroughputSampler.  Numbers almost always be declared as variables, to convey understanding.  Even better would be defining a GuranteedThroughputSampler in a beforeEach block, since most of them have the same parameters.  If some of the samplers have unique parameters then override them for that test.

test/samplers/guaranteed_throughput_sampler.js
```diff
@@ -69,53 +69,76 @@ describe('GuaranteedThroughput sampler', () => {
...
         let p2 = sampler._lowerBoundSampler;
-        let isUpdated: boolean = sampler.update(2, 1.0);
+        let p3 = sampler._upperBoundRateLimiter;
+        let isUpdated: boolean = sampler.update(1.0, 2, 3);
         assert.isFalse(isUpdated);
         assert.strictEqual(sampler._probabilisticSampler, p1);
         assert.strictEqual(sampler._lowerBoundSampler, p2);
-        assertValues(sampler, 2, 1.0);
+        assert.strictEqual(sampler._upperBoundRateLimiter, p3);
+        assertValues(sampler, 1.0, 2, 3);
```
oibe: same concept as above, since these assertValues match the sampler params then use the same variable.

PR #128, opened 2017-06-21
Consider sampling flag as a recommendation

vprithvi:
Add the ability to upsample child spans.
When a service recieves a rpc that has a span context, but with flags set to 0,
the service may decide to resample the span.


src/configuration.js
```diff
@@ -150,6 +151,7 @@ export default class Configuration {
      * @param {Object} [options.logger] - a logger (see ./_flow/logger.js)
      * @param {Object} [options.tags] - set of key-value pairs which will be set
      *        as process-level tags on the Tracer itself.
+     * @param {Object} [options.upsampleChildSpans] - if set, the tracer will upsample child spans
```
yurishkuro: I think we should have a top-level "upsampling" flag that refers to first-in-process span, in addition to `upsampleChildSpans`. It could be a hierarchical struct like
```javascript
'upsampling': {
  'enabled': true,
  'allowOnInnerSpans': false,
}
```

Because without it we're changing the existing behavior of the client where it respects the upstream sampling decision.

src/tracer.js
```diff
@@ -237,7 +240,9 @@ export default class Tracer {
             ctx.baggage = parent.baggage;
 
             parent.finalizeSampling();
-            this.processDeferredSampling(ctx, operationName, internalTags);
+            if (this._upsampleChildSpans) {
+                this.upsampleChild(parent, ctx, operationName, internalTags);
+            }
             ctx.finalizeSampling();
```
yurishkuro: not sure this is correct anymore. When upsampling a child we still need to account for the possibility of operationName changing and us having to re-sample it

src/tracer.js
```diff
@@ -254,23 +259,20 @@ export default class Tracer {
...
      * @param tags tags to be applied by the sampler
      */
-    processDeferredSampling(ctx: SpanContext, operationName: string, tags: any) {
-        if (ctx.isDeferredSampling) {
+    upsampleChild(parentCtx: SpanContext, ctx: SpanContext, operationName: string, tags: any) {
+        if (!parentCtx.isSampled() && !parentCtx.isDebug()) {
             if (this._sampler.isSampled(operationName, tags)) {
                 ctx._flags |= constants.SAMPLED_MASK;
             } else {
                 ctx._flags &= ~constants.SAMPLED_MASK;
```
yurishkuro: per the commend above, how do we capture the non-final nature of upsampling decision?

test/init_tracer.js
```diff
@@ -156,4 +155,26 @@ describe('initTracer', () => {
...
+        beforeEach(() => {
+            config = {
+                serviceName: 'test-service'
+            };
+        });
+
+        it('should default to true', () => {
+            let tracer = initTracer(config);
+
+            assert.isOk(tracer._upsampleChildSpans)
```
yurishkuro: I don't think it should

test/init_tracer.js
```diff
@@ -156,4 +155,26 @@ describe('initTracer', () => {
         assert.equal(tracer._metrics._factory, metrics);
         assert.equal(tracer._tags['x'], 'y');
     });
+
+    describe("upsampleChildSpans", () => {
```
yurishkuro: we are getting to a situation where the possible state machine is growing due to various dimensions, so we need to have a data driven test. Dimensions include (could be more, some may not be fully orthogonal):
  * root span or not
  * first in process span or not
  * child span or not
  * upstream span was sampled or not
  * upsampling enabled or not
    * upsampling on inner spans enabled or not
  * finalized or not
  * operation name changed or not
  * TODO: there's PR in OT to allow changing service name via "service" tag, which has similar effect to changing op name

src/tracer.js
vprithvi: I propose moving sampling from when a span is created to when the span is injected/finished/used as a parent and remove the notion of finalizing a sampling decision.

This change has the following advantages
- Simplifies the code by removing the redundant sampling logic in startSpan
- Upsampling logic isn't separated from the standard sampler. 
- Fixes metrics about spans sampled being incorrect because they are emitted based on the initial sampling decision
- Fixes the bug where a sampling decision is finalized the first time setOperation is called. (We can fix this independently, but I feel it's due to repeated code) 
- Simplifies integration when service names are tags in spans.

Apart from the refactoring required, I don't may big drawbacks. What are your thoughts on this?

yurishkuro: > from when a span is created to when the span is injected/finished/used as a parent

doesn't it mean instead of doing it in one place we're going to do it in 3 places?

vprithvi:
Yes, we already have resampling logic in those three places. I don't think
that there is a easy way of reducing that complexity

src/tracer.js
```diff
@@ -54,11 +55,13 @@ export default class Tracer {
...
      * @param {Object} [options.metrics] - instance of the Metrics class from ./metrics/metrics.js.
      * @param {Object} [options.logger] - a logger matching NullLogger API from ./logger.js.
+     * @param {Boolean} [upsampling] - determines whether to resample child spans of unsampled parents.
      */
     constructor(serviceName: string,
             reporter: Reporter = new NoopReporter(),
             sampler: Sampler = new ConstSampler(false),
-            options: any = {}) {
+            options: any = {},
+            upsampling: any = {}) {
```
yurishkuro: upsampling should be passed as part of the `options`: `options.upsampling.enabled`

yurishkuro: because new argument creates an unnecessary API change (many tests are affected for no reason)

src/configuration.js
```diff
@@ -179,15 +184,23 @@ export default class Configuration {
             );
         }
 
+        let upsamplingEnabled = false;
+        if (config.upsampling && config.upsampling.enabled !== null && config.upsampling.enabled) {
```
black-adder: do we need the null check? according to https://developer.mozilla.org/en-US/docs/Glossary/Falsy any uninitialized value should make this false

src/tracer.js
vprithvi: > because new argument creates an unnecessary API change (many tests are affected for no reason)

The test changes were leftover from a previous change, they are unnecessary. 

I'm not very clear on why something goes into `options` vs using function parameters as my js experience is limited. Could you elucidate?

src/tracer.js
```diff
@@ -254,23 +257,26 @@ export default class Tracer {
     }
 
     /**
-     *  Makes a concrete sampling decision for the ctx span context based on information
-     *  available to it's child span.
-     *
-     *  The ctx context's deferred mask is unset after this decision is made
+     * Reapplies sampling if the parent span isn't sampled
      *
+     * @param config the upsampling config
```
black-adder: does the api need to have config passed in? given config is available to `this`? is this just for tests?

CHANGELOG.md
```diff
@@ -2,10 +2,10 @@ sh --registry=https://registry.npmjs.org
 Changes by Version
 ==================
 
-3.5.4 (unreleased)
+3.6.0 (unreleased)
 ------------------
 
-- Nothing yet.
+- Add support for upsampling
```
black-adder: add link to this PR

src/tracer.js
oibe: ES6 has a strange implementation of optional parameters.  Its not like python were I can specify which optional parameter I mean with a keyword.  So if optional parameter 1 is missing  you can't specify optional parameter 2.  Basically it lets nullable arguments not be passed, however they are not true optional parameters.

src/tracer.js
```diff
@@ -254,23 +257,26 @@ export default class Tracer {
...
+     * @param parentCtx the parent span context
      * @param ctx the span context
      * @param operationName the operation name for the child
      * @param tags tags to be applied by the sampler
      */
-    processDeferredSampling(ctx: SpanContext, operationName: string, tags: any) {
-        if (ctx.isDeferredSampling) {
+    handleUpsampling(config: any, parentCtx: SpanContext, ctx: SpanContext,
+                     operationName: string, tags: any) {
+        if (!config.enabled) {
```
oibe: if config is null you can get an exception often in ES6 people initialize a default dictionary, or write config = config || {}

test/span.js
```diff
@@ -43,7 +42,8 @@ describe('span should', () => {
             'test-service-name',
             reporter,
             new ConstSampler(true),
-            { logger: new MockLogger() }
+            { logger: new MockLogger() },
+            {},
```
oibe: Its not clear to me what this parameter is.  Can we do something like "let upsamplingConfig = {}", also its been a while since I coded ES6, but this empty dict may refer to injectors, since that is the next parameter.

src/configuration.js
vprithvi: Looks like we don't!

src/tracer.js
vprithvi: No, this isn't required.

test/span.js
vprithvi: Removed this parameter

src/tracer.js
vprithvi: Removed the config

src/configuration.js
```diff
@@ -150,6 +155,7 @@ export default class Configuration {
      * @param {Object} [options.logger] - a logger (see ./_flow/logger.js)
      * @param {Object} [options.tags] - set of key-value pairs which will be set
      *        as process-level tags on the Tracer itself.
+     * @param {Object} [options.upsampling] - overrides upsampling set on config
```
yurishkuro: this should be passed via config. `options` here is typically for things that cannot be initialized from a config file, like live dependencies.

src/tracer.js
```diff
@@ -254,23 +256,25 @@ export default class Tracer {
...
+    handleUpsampling(parentCtx: SpanContext, ctx: SpanContext,
+                     operationName: string, tags: any) {
+        if (!this._upsampling.enabled) {
+            return
+        }
+
+        if (!parentCtx.isSampled() && !parentCtx.isDebug()) {
             if (this._sampler.isSampled(operationName, tags)) {
                 ctx._flags |= constants.SAMPLED_MASK;
             } else {
```
yurishkuro: this is doing downsampling, we don't want that

yurishkuro: this is doing downsampling, we don't want that

src/tracer.js
```diff
@@ -237,8 +240,7 @@ export default class Tracer {
             ctx.baggage = parent.baggage;
 
             parent.finalizeSampling();
-            this.processDeferredSampling(ctx, operationName, internalTags);
-            ctx.finalizeSampling();
+            this.handleUpsampling(parent, ctx, operationName, internalTags);
```
yurishkuro: I'm not convinced if simply removing finalizedSampling has the same effect. If the span is started with opName="GET" and upsampling decides to sample, this is going to be final regardless of changes to the operationName later. Previously `finalized` logic was only making it just before starting a new span (or serializing)

test/span.js
```diff
@@ -185,31 +184,41 @@ describe('span should', () => {
         assert.isOk(unnormalizedKey in Span._getBaggageHeaderCache());
     });
 
-    describe('with deferred sampling', () => {
```
yurishkuro: should not be removing this test (it could've been written without explicitly mentioning deferred tag).

src/configuration.js
vprithvi: Done

src/tracer.js
vprithvi: My understanding of downsampling is dropping spans with a sampled decision. 
Could you elaborate on how this code performs downsampling?

test/span.js
vprithvi: Most test cases that this test covered are covered in the `with upsampling` test.
The `should not pass deferred sampling flag to child spans` no longer applies to upsampling because we don't have a special flag.

yurishkuro: your new test does not cover the use case of changing the operation name, which was the main reason for deferred samping logic

src/tracer.js
yurishkuro: `ctx._flags &= ~constants.SAMPLED_MASK` can reset sampled flag to 0, it is downsampling

vprithvi: >ctx._flags &= ~constants.SAMPLED_MASK can reset sampled flag to 0, it is downsampling

We will never reach that line when we encounter a sampled or debug parent because of `!parentCtx.isSampled() && !parentCtx.isDebug()`.

src/tracer.js
vprithvi: `handleUpsampling` does not call `ctx.finalizeSampling`; why would the first operationName change be disregarded?
(This case seems to be missing a test, let me add it)

src/tracer.js
yurishkuro: then remove it

src/tracer.js
```diff
@@ -237,8 +240,11 @@ export default class Tracer {
             ctx.baggage = parent.baggage;
 
             parent.finalizeSampling();
-            this.processDeferredSampling(ctx, operationName, internalTags);
-            ctx.finalizeSampling();
+            if(parent.isSampled() || parent.isDebug()) {
+                ctx.finalizeSampling();
+            } else {
+                this.handleUpsampling(ctx, operationName, internalTags);
```
yurishkuro: nit: s/handleUpsampling/tryUpsampling/ ("handle" is too generic)

src/constants.js
```diff
@@ -21,14 +21,9 @@
...
 export const SAMPLED_MASK = 0x1;
 
-// DEBUG_MASK is the bit mask indicationg that a span has been marked for debug.
+// DEBUG_MASK is the bit mask indicating that a span has been marked for debug.
 export const DEBUG_MASK = 0x2;
 
-// DEFERRED_SAMPLING_MASK is the bit mask indicating that the upstream service that generated the
-// span did not make a definitive sampling decision, but deferred it to be made by some
-// downstream service.
-export const DEFERRED_SAMPLING_MASK = 0x4;
```
yurishkuro: to keep a cleaner git history, can you revert deferred sampling diff first in master, and rebase this? I would rather see diffs vs. "finalized" logic, not the deferred logic that we're throwing out.

src/tracer.js
```diff
@@ -254,23 +260,19 @@ export default class Tracer {
...
      */
-    processDeferredSampling(ctx: SpanContext, operationName: string, tags: any) {
-        if (ctx.isDeferredSampling) {
-            if (this._sampler.isSampled(operationName, tags)) {
-                ctx._flags |= constants.SAMPLED_MASK;
-            } else {
-                ctx._flags &= ~constants.SAMPLED_MASK;
-            }
-            ctx.unsetDeferredSampling();
+    handleUpsampling(ctx: SpanContext, operationName: string, tags: any) {
```
yurishkuro: this does not take into account `allowOnInnerSpans` flag - https://github.com/uber/jaeger-client-node/pull/128#discussion_r123358878

test/span.js
```diff
@@ -185,31 +184,80 @@ describe('span should', () => {
...
+                },
+                new: {
+                    operationName: newOperation,
+                    samplingDecision: newSamplingDecision,
+                },
+                finalSamplingDecision: finalSamplingDecision
+            }
+        }
+        const testCases = [
+            makeTestCase(0, true, true, "idli", null, null, true),
```
yurishkuro: this is quite difficult to read. When we do similar tests in Go we always try to use struct initialization with named fields. Here you don't need the helper method, just define the objects inline with explicit field names. And you can drop false / null assignments, only define fields relevant to the test. 

You also don't have to define an array, you can write the tests like this:

```
describe('upsampling', () =>{
    beforeEach...
    doit('first test description', {
        flags: ...,
        callSampler: ...,
        ...,
    });
    doit(... another test ...)
    // then the actual body
    function doit(name, testCase) {
        it(name, () => {
           // test logic
        });
    }
});
```

The benefit of this structure is that you can easily breakpoint on specific tests and step through, which is much harder when you drive them from an array of data.

test/span.js
```diff
@@ -185,31 +184,80 @@ describe('span should', () => {
...
+            makeTestCase(constants.DEBUG_MASK, false, true, "avial", false, "puri", true),
+            makeTestCase(constants.DEBUG_MASK, false, false, "avial", true, "puri", true),
+            makeTestCase(constants.DEBUG_MASK, false, false, "avial", false, "puri", true),
+            
+            makeTestCase(constants.SAMPLED_MASK | constants.DEBUG_MASK, false, true, "avial", null, null, true),
+            makeTestCase(constants.SAMPLED_MASK | constants.DEBUG_MASK, false, false, "avial", null, null, true),
+            makeTestCase(constants.SAMPLED_MASK | constants.DEBUG_MASK, false, true, "avial", true, "puri", true),
+            makeTestCase(constants.SAMPLED_MASK | constants.DEBUG_MASK, false, true, "avial", false, "puri", true),
+            makeTestCase(constants.SAMPLED_MASK | constants.DEBUG_MASK, false, false, "avial", true, "puri", true),
+            makeTestCase(constants.SAMPLED_MASK | constants.DEBUG_MASK, false, false, "avial", false, "puri", true),
```
yurishkuro: given the amount of tests, I would suggest moving this to another file `tests/upsampling.js`

test/span.js
```diff
@@ -185,31 +184,80 @@ describe('span should', () => {
...
+
+                let mockSampler = sinon.mock(tracer._sampler);
+                if (testCase.callSampler) {
+                    mockSampler.expects('isSampled').withExactArgs(testCase.original.operationName, {})
+                        .returns(testCase.original.samplingDecision);
+                }
+
+                let child = tracer.startSpan(testCase.original.operationName, {childOf: span.context()});
+
+                if (testCase.callSampler && testCase.new.operationName) {
```
yurishkuro: shouldn't `callSampler` be a property of each of `original` and `new`? Are there scenarios when it's called the first time but not the second?

test/span.js
```diff
@@ -185,31 +184,80 @@ describe('span should', () => {
...
+                },
+                finalSamplingDecision: finalSamplingDecision
+            }
+        }
+        const testCases = [
+            makeTestCase(0, true, true, "idli", null, null, true),
+            makeTestCase(0, true, false, "idli", null, null, false),
+            makeTestCase(0, true, true, "poriyal", true, "murungakkai", true),
+            // We ignore the following case because setOperationName can only go from a sampling decision of 0->1
+            // makeTestCase(0, true, true, "poriyal", false, "murungakkai", false),
```
yurishkuro: if it works, why ignore it?

test/span.js
```diff
@@ -185,31 +184,80 @@ describe('span should', () => {
...
-            let child = tracer.startSpan('goodOperation', {childOf: span.context()});
-            mockSampler.verify();
-            assert.isOk(child.context().isSampled());
-        });
+        function makeTestCase(flags, callSampler, originalSamplingDecision, originalOperation, newSamplingDecision, newOperation, finalSamplingDecision) {
+            return {
+                flags: flags,
+                callSampler: callSampler,
+                name: `Should ${callSampler?"":"not "}call sampler for flag ${flags} ${newOperation?"with modified operationName and samplingDecision":""}`,
+                original: {
```
yurishkuro: you could define these as arrays:
```
   operationNames: [
      { name: 'foo', callSampler: true, samplingDecision: true },
      { name: 'bar', callSampler: true, samplingDecision: true }
   ]
```

test/span.js
vprithvi: It could be, but there are no cases where the sampler is called the first time but not the second when there is an `operationName` change.

test/span.js
vprithvi: This testcase will fail because it is an instance of downsampling

yurishkuro: even more reason to have a test - we certainly want to test the scenario where downsampling "would have happened" and make sure it doesn't

src/tracer.js
vprithvi: Yes, I didn't want to add complexity when we don't have a need for it yet. Do you think it is important to support `allowOnInnerSpans` right now?

src/constants.js
vprithvi: +1, #132

test/span.js
vprithvi: I really like your suggestion, let me update the tests.

src/tracer.js
yurishkuro: I think it is, otherwise you are going to keep re-sampling on every inner span. By default we only want to re-sample on the first in process span

src/tracer.js
```diff
@@ -73,6 +75,20 @@ export default class Tracer {
...
+            this._upsampling = {
+                allowOnInnerSpans: options.upsampling.allowOnInnerSpans
+                    ? options.upsampling.allowOnInnerSpans : false,
+                enabled: options.upsampling.enabled ? options.upsampling.enabled : false
+            };
+        } else {
+            this._upsampling = {
+                enabled: false,
+                allowOnInnerSpans: false
+            }
```
black-adder: missing semicolon

src/tracer.js
```diff
@@ -237,7 +253,11 @@ export default class Tracer {
             ctx.baggage = parent.baggage;
 
             parent.finalizeSampling();
-            ctx.finalizeSampling();
+            if(parent.isSampled() || parent.isDebug()) {
```
black-adder: nit: space between if and (

src/tracer.js
```diff
@@ -253,6 +273,38 @@ export default class Tracer {
...
 
     /**
+     * Reapplies sampling if the parent span isn't sampled
+     *
+     * @param parent the parent span context
+     * @param ctx the span context
+     * @param operationName the operation name for the child
+     * @param tags tags to be applied by the sampler
+     */
+    handleUpsampling(parent: SpanContext, ctx: SpanContext, operationName: string, tags: any) {
```
black-adder: i has some questions about this function, i'll come bug you later

black-adder: If we were to upsample this span but later change the operation name, we will attempt to sample again. If  we decide to not sample, the upsamplingDecision on the parent is still set to true and the next child span of the parent will be sampled. Is that ok?

src/tracer.js
```diff
@@ -253,6 +273,38 @@ export default class Tracer {
...
+        // Run upsampler on every span
+        if (this._upsampling.allowOnInnerSpans) {
+            if (this._sampler.isSampled(operationName, tags)) {
+                ctx.samplingDecision = true;
+            }
+            return
+        }
+
+        // Reuse upsampling decision, if one exists in parent
+        if (parent.upsamplingDecision) {
```
black-adder: This check can go before L289? If the upsampling decision is already true on the parent, you can skip L290.

src/tracer.js
```diff
@@ -253,6 +273,38 @@ export default class Tracer {
...
+     * @param tags tags to be applied by the sampler
+     */
+    handleUpsampling(parent: SpanContext, ctx: SpanContext, operationName: string, tags: any) {
+        if (!this._upsampling.enabled) {
+            return
+        }
+
+        // Run upsampler on every span
+        if (this._upsampling.allowOnInnerSpans) {
+            if (this._sampler.isSampled(operationName, tags)) {
```
black-adder: if you're the first span in the process and you decide to upsample here, don't you have to set the upsampling decision on the parent?

test/span_context.js
```diff
@@ -60,6 +60,19 @@ describe ('SpanContext should', () => {
...
+        context.samplingDecision = 1;
+        assert.equal(context._flags, constants.SAMPLED_MASK);
+    });
+
+    it ('should set sampled to 0 when passed 0', () => {
+        let context = SpanContext.fromString('100:7f:0:1');
+        context.samplingDecision = 0;
+        assert.equal(context._flags, 0);
+    });
+    //TODO: Add tests with debug flag
```
black-adder: Why can't this TODO be done in this PR?

src/tracer.js
vprithvi: This is not ok, I'll add a test and fix this behavior.

src/tracer.js
vprithvi: Not really, each sampling decision can be made independently here.

src/tracer.js
vprithvi: I spoke too soon, I'd forgotten that `setOperationName` can only go from a sampling decision of 0->1. The situation that you are talking about can never arise.

src/tracer.js
vprithvi: No, it cannot because if `allowOnInnerSpans` is set, we would use need a new sampling decision on every new span. We shouldn't reuse sampling decisions.

test/span_context.js
vprithvi: No reason, I'll add it

src/tracer.js
black-adder: @ys is this really what we want? should allowOnInnerSpans allow you to have different sampling decisions for each span in the process? I don't see the use case for this, if the parent span in the process is upsampled, all child spans should respect it and be upsampled. Don't know why we'd want to flip the coin for each span.

yurishkuro: upsampling on inner spans is an opt-in behavior. TBH I don't see much use for it either. But if the parent is already upsampled, there should be no coin flipping, we never downsample.

vprithvi: @yurishkuro I'm confused on why you asked for this configuration option in the first place. If it doesn't have a good use case, I feel we shouldn't provide this. Do you have any objections on removing this behavior?

src/span_context.js
```diff
@@ -70,6 +75,7 @@ export default class SpanContext {
         this._baggage = baggage;
         this._debugId = debugId;
         this._samplingFinalized = samplingFinalized;
+        this.upsamplingDecision = null;
```
black-adder: should this be _upsamplingDecision?

src/tracer.js
yurishkuro: I did not ask for it. Your earlier version was applying upsampling to _all spans_, and I said if you want to do that, it has to be opt-in, with default being only first-in-process span.

vprithvi: I apologize, I understood https://github.com/uber/jaeger-client-node/pull/128#discussion_r125911128 to mean that you wanted me to add support for the configuration option. 
I would have understood you better if it was simply stated that there isn't a use case for the configuration and that we should always apply the decision on the root span to the inner spans.

src/span_context.js
```diff
@@ -49,6 +49,11 @@ export default class SpanContext {
      * - its context is serialized using injectors
```
yurishkuro: why is the IDL changing, is it related to this diff?

src/tracer.js
```diff
@@ -73,6 +75,17 @@ export default class Tracer {
...
 
+        if (options.upsampling) {
+            this._upsampling = {
+                enabled: options.upsampling.enabled ? options.upsampling.enabled : false
+            };
+        } else {
+            this._upsampling = {
+                enabled: false
+            };
+        }
```
yurishkuro: this can be done without the if
```
enabled: (options.upsampling && options.upsampling.enabled) ? true : false
```

src/tracer.js
```diff
@@ -253,6 +270,32 @@ export default class Tracer {
...
+            ctx.samplingDecision = parent.upsamplingDecision.valueOf();
+            ctx.upsamplingDecision = parent.upsamplingDecision;
+            return
+        }
+
+        // Store upsampling decision in parent
+        parent.upsamplingDecision = new Boolean(this._sampler.isSampled(operationName, tags));
+        ctx.samplingDecision = parent.upsamplingDecision.valueOf();
+        ctx.upsamplingDecision = parent.upsamplingDecision;
+    }
```
yurishkuro: this doesn't look right.
  * I thought upsampling and finalized will be the same now. When do they need to be different?
  * the propagation of upsampling into the parent in L293 looks just wrong, why do we want that?
  * I am not seeing the conditional of only allowing upsampling on first in process span.

src/span_context.js
vprithvi: I assumed that was because of the merge from master

src/tracer.js
vprithvi: Yes, I missed it because we removed allowOnInnerSpans, will update

src/tracer.js
vprithvi: >I thought upsampling and finalized will be the same now. When do they need to be different?

vprithvi: >I thought upsampling and finalized will be the same now. When do they need to be different?

They are different for sampled and debug spans. 
They are also different for unsampled spans where `setOperation` hasn't been called.
Why did you expect them to be the same?

>the propagation of upsampling into the parent in L293 looks just wrong, why do we want that?
We use that to store the first sampling decision so that that other in process spans can reuse it. Why do you think that it is wrong?

>I am not seeing the conditional of only allowing upsampling on first in process span.
We don't have a conditional because it is the only behavior supported.

yurishkuro: `debug` flag makes all other flags irrelevant. It implies finalized=true.

`sampled` flag does not imply finalized, but if sampled=true, it can never become false, and it will eventually become finalized.

Upsampling can only happen
  * on the *first span in process*
  * which is *not* a root span (otherwise regular sampling happens)
  * if that span has `sampled=false`

But it can happen at most twice:
  * in `startSpan` where currently we simply inherit parent `flags` and finalize both parent and new span. Instead, we can make a sampling decision here (subject to conditions above). If that decision is Yes then upsampling use case is done, it will never come up again. If that decision is No (i.e. we didn't change sampled=false inherited from parent), then it can be retried in setOperationName. The other difference here is that previously we were always finalizing the child span. Now we can skip finalizing if we match the upsampling criteria.
  * in setOperationName the difference now is that we could reach it for a child span that's not finalized (was impossible before). We'd still have sampled=false as otherwise it doesn't matter. I don't think we need any changes there, it would "just work" and potentially upsample the span due to it not being finalized (but it would only apply to first-in-process span per above conditions.

vprithvi: I didn't realize that we could determine the _first span in process_ easily using tags. The flag was only used to maintain that information, so I was able to remove it.

src/tracer.js
```diff
@@ -236,8 +243,13 @@ export default class Tracer {
             // reuse parent's baggage as we'll never change it
             ctx.baggage = parent.baggage;
 
-            parent.finalizeSampling();
```
yurishkuro: parent.finalized should behave same as before, I think, why do you remove it?

src/tracer.js
```diff
@@ -253,6 +265,28 @@ export default class Tracer {
...
+     */
+    handleUpsampling(parent: SpanContext, ctx: SpanContext, operationName: string, tags: any) {
+        if (!this._upsampling.enabled) {
+            return
+        }
+
+        let samplingDecision = this._sampler.isSampled(operationName, tags);
+
+        parent.samplingDecision = samplingDecision;
+        parent.finalizeSampling();
```
yurishkuro: why are we touching parent here? If this span is first in process, then parent is irrelevant, but not first in process, then we will inherit parent's decision without upsampling

src/tracer.js
vprithvi: Restored.

src/tracer.js
vprithvi: Agree, updated.

test/init_tracer.js
```diff
@@ -156,4 +155,23 @@ describe('initTracer', () => {
...
+        };
+
+        it('should default to false', () => {
+            let tracer = initTracer(config);
+
+            assert.isNotOk(tracer._upsampling.enabled)
+        });
+
+        it('should pass true if set to true', () => {
+            config.upsampling = {"enabled" : true};
```
tiffon: Typically what I would recommend here is instead of mutating the shared config create a new config with the desired change:
```
const config = { ...config, upsampling: {"enabled" : true}}
```

(This is more relevant when the config is shared between more test cases.)

test/init_tracer.js
```diff
@@ -156,4 +155,23 @@ describe('initTracer', () => {
         assert.equal(tracer._metrics._factory, metrics);
         assert.equal(tracer._tags['x'], 'y');
     });
+
+    describe("upsampling", () => {
+        let config = {
```
tiffon: Why not use `const`? Side note: Eslint can be config'd to detect when declarations should use `const`.

test/init_tracer.js
```diff
@@ -156,4 +155,23 @@ describe('initTracer', () => {
...
+            serviceName: 'test-service'
+        };
+
+        it('should default to false', () => {
+            let tracer = initTracer(config);
```
tiffon: Why not use `const`? This is present in a number of other places, but will omit, hereafter.

PR #130, opened 2017-06-29
Bump jaeger idl

black-adder:
Bumping jaeger-idl broke some things. Before `jaeger.idl` had the emitBatch function but it was moved to `agent.idl`; however, `agent.idl` is not thriftrw compatible.

I added a new `agent.idl` that's thriftrw compatible, but it has to be manually updated to keep up to date. The thriftrw idl could be moved to jaeger-idl repo however it's only nodejs specific.

### DIFF
--- a/crossdock/test/endtoend_handler.js
+++ b/crossdock/test/endtoend_handler.js
@@ -40,8 +40,9 @@ describe('Endtoend Handler should', () => {
         server = dgram.createSocket('udp4');
         server.bind(PORT, HOST);
         thrift = new Thrift({
-            source: fs.readFileSync(path.join(__dirname, '../../src/jaeger-idl/thrift/jaeger.thrift'), 'ascii'),
-            allowOptionalArguments: true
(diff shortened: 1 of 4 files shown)


yurishkuro:
> however, agent.idl is not thriftrw compatible.

what is not compatible?

black-adder:
2 Things:
1) For thriftrw to work, include paths must be prefixed with ./ or ../ (This is doable)
2) zipkincore.thrift is not compatible with thriftrw and I don't want to rewrite it to be compatible

yurishkuro:
Can we generate the compatible agent.thift in the Makefile instead of creating it manually? As far as I understand the only thing you need is to remove zipkin import and method.

Can we not change the import to `./` in the main idl repo? Will it break regular thrift?

black-adder:
I have two choices:
1) hard code it like I've done
2) write some bash code that will remove zipkin, format imports, generate a new .thrift file

I don't see the value in the second option. If we were to change agent.thrift in the future, then the script might break given how fragile the logic is so both methods aren't that future proof but option 1 is easier.

PR #131, opened 2017-07-05
Add baggage restriction manager

black-adder:


src/baggage/remote_baggage_restriction_manager.js
```diff
@@ -0,0 +1,161 @@
...
+
+    _initialized: boolean;
+
+    /**
+     * Creates a BaggageRestrictionManager that fetches baggage restrictions remotely from jaeger-agent.
+     *
+     * @param {string} [serviceName] - name of the current service / application, same as given to Tracer
+     * @param {object} [options] - optional settings
+     * @param {object} [options.logger] - optional logger, see _flow/logger.js
+     * @param {object} [options.failClosed] - this determines the startup failure mode of RemoteBaggageRestrictionManager.
```
vprithvi: It isn't easy to grasp what`failClosed` does simply by looking at the name.
I suggest renaming it to something like the following:
```
startupFailurePolicy:  allow | deny
```

black-adder: I thought so too, but according to Yuri: "it's a very common engineering term and you should be ashamed for not knowing it"

src/baggage/remote_baggage_restriction_manager.js
```diff
@@ -0,0 +1,161 @@
...
+const DEFAULT_BAGGAGE_PORT = 5778;
+const DEFAULT_FAIL_CLOSED = false;
+
+export default class RemoteBaggageRestrictionManager {
+
+    _serviceName: string;
+    _logger: Logger;
+    _metrics: Metrics;
+
+    _refreshInterval: number;
```
vprithvi: Could you append the unit to the key, something like `refreshIntervalMs` or so

src/baggage/remote_baggage_restriction_manager.js
```diff
@@ -0,0 +1,161 @@
...
+     * If failClosed is true, RemoteBaggageRestrictionManager will not allow any baggage to be written until baggage restrictions
+     * have been retrieved from agent. If failClosed is false (ie FailOpen), RemoteBaggageRestrictionManager will allow any baggage
+     * to be written until baggage restrictions have been retrieved from agent.
+     * @param {object} [options.metrics] - instance of Metrics object
+     * @param {number} [options.refreshInterval] - interval in milliseconds before baggage restrictions refreshes (0 to not refresh)
+     * @param {string} [options.host] - host for jaeger-agent, defaults to 'localhost'
+     * @param {number} [options.port] - port for jaeger-agent for BaggageRestrictionManager endpoint
+     * @param {function} [options.onRestrictionsUpdate]
+     */
+    constructor(serviceName: string, options: any = {}) {
```
vprithvi: How does the caller know that startup of the `RemoteBaggageRestrictionManager` is successful?

src/span.js
```diff
@@ -106,6 +122,9 @@ export default class Span {
             if (this.getBaggageItem(normalizedKey)) {
                 logs['override'] = 'true';
             }
+            if (truncated) {
+                logs['truncated'] = 'true';
```
vprithvi: Is there a better way of capturing this data? 
With multiple baggage items, it might be difficult to figure out which one was truncated

src/baggage/remote_baggage_restriction_manager.js
```diff
@@ -0,0 +1,161 @@
...
+        this._metrics = options.metrics || new Metrics(new NoopMetricFactory());
+        this._refreshInterval = options.refreshInterval || DEFAULT_REFRESH_INTERVAL;
+        this._host = options.host || DEFAULT_BAGGAGE_HOST;
+        this._port = options.port || DEFAULT_BAGGAGE_PORT;
+        this._failClosed = options.failClosed || DEFAULT_FAIL_CLOSED;
+        this._restrictionsMap = {};
+        this._initialized = false;
+        this._onRestrictionsUpdate = options.onRestrictionsUpdate;
+
+        this._refreshBaggageRestrictions();
```
vprithvi: Why does this not respect the initial delay?

src/baggage/remote_baggage_restriction_manager.js
```diff
@@ -0,0 +1,161 @@
...
+            this._metrics.baggageRestrictionsUpdateFailure.increment(1);
+            return;
+        }
+        if (this._onRestrictionsUpdate) {
+            this._onRestrictionsUpdate(this);
+        }
+    }
+
+    _updateRestrictions(restrictions: Array<BaggageRestriction>) {
+        let restrictionsMap: { [key: string]: number } = {};
```
vprithvi: Does it make more sense to have the agent return a map?

src/baggage/remote_baggage_restriction_manager.js
yurishkuro: https://www.google.com/search?q=fail+open+fail+closed. Widely used in security.

But I agree that it's not the best name to use for the options, I was simply referring to the principle. A property with `allow | deny` values sgtm. Is it only on startup? I'd be more specific with the name.

src/baggage/remote_baggage_restriction_manager.js
black-adder: The user can decide up front what to do incase of initialization failure via the failclosed variable. I dont think we need a callback.

src/baggage/remote_baggage_restriction_manager.js
black-adder: the baggage restrictions cache should be initialized (or at least attempted) before this constructor completes. This way when the tracer is initialized, users can start adding baggage right away rather than having to wait for at most 60 seconds for the initial delay.

src/baggage/remote_baggage_restriction_manager.js
black-adder: sheeeeet, yeah this is a good idea... My only gripe with this approach is that in the idl, I defined it as: https://github.com/uber/jaeger-idl/blob/master/thrift/baggage.thrift which I prefer over map<string, i32> because it's harder to tell what the variables mean without having a comment describing it.

Now the debate becomes lets leave the thrift idl as is but have the agent convert the array into a map before returning it. This does make sense but again you have that issue where as the consumer of the json, you dont know what the variables in the map symbolizes. By returning it as json array instead of map, each variable is keyed correctly so you know what they symbolize. wdyt?

src/span.js
black-adder: Each baggage item gets its own set of logs. If we were to dump everything into the same logs, then there would be trouble. I think this current approach is fine.

src/baggage/remote_baggage_restriction_manager.js
vprithvi: Yes, but what if I want custom behavior. For e.g., I don't want the service to startup if this condition is not met.

src/baggage/remote_baggage_restriction_manager.js
vprithvi: I'm not a fan of having blocking network operations in the constructor by default. 
While this would work well when the jaeger-agent is healthy, it could cause an unnecessary increase in service startup time when the http connection is accepted by jaeger-agent, but jaeger-agent takes a long time to respond. In the worst case, it could wait until the http timeout.

vprithvi: Why though? 
Without making this a blocking call, there is no guarantee that the request was completed.

src/baggage/remote_baggage_restriction_manager.js
vprithvi: re: thrift idl: 
Thrift supports typedefs, so you can have a map<BaggageKey, BaggageValue>

yurishkuro: I prefer wire format to be lists, not maps. But thrift format should be hidden behind an interface, it's fine to return a map from the interface.

vprithvi: @yurishkuro Why do you prefer the wire format to be a list even when holding map data? What advantages does it have to simply having it as a map?

yurishkuro: extensibility - https://github.com/uber/jaeger-idl/pull/30

PR #137, opened 2017-07-19
Uncaught exception from finish()

### LINKED ISSUE #136: Uncaught exception from finish()
Moved from https://github.com/opentracing/opentracing-javascript/issues/83

@frankgreco is getting an uncaught exception

```
{ Error: getaddrinfo ENOTFOUND jaeger-agent.kube-system.svc.cluster.local
    at errnoException (dns.js:28:10)
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:76:26)
  code: 'ENOTFOUND',
  errno: 'ENOTFOUND',
  syscall: 'getaddrinfo',
  hostname: 'jaeger-agent.kube-system.svc.cluster.local' }
```

We need to wrap the sender in a try/catch and log the error if the logger is provider.

frankgreco:
Resolves #136

### DIFF
--- a/src/tracer.js
+++ b/src/tracer.js
@@ -142,7 +142,7 @@ export default class Tracer {
         try {
           this._reporter.report(span);
         } catch (err) {
-          this._logger.error(`could not report span ${span.operationName()}: ${err.message}`)
+          this._logger.error(`could not report span ${span.operationName}: ${err.message}`)
         }
     }
 
@@ -186,7 +186,7 @@ export default class Tracer {
         let userTags = options.tags || {};
         let startTime = options.startTime || this.now();
 
-        // This flag is used to ensure that CHILD_OF reference is preferred 
+        // This flag is used to ensure that CHILD_OF reference is preferred
         // as a parent even if it comes after FOLLOWS_FROM reference.
         let followsFromIsParent = false;
         let parent: ?SpanContext = options.childOf instanceof Span ? options.childOf.context() : options.childOf;
--- a/src/tracer.js
+++ b/src/tracer.js
@@ -139,7 +139,11 @@ export default class Tracer {
 
     _report(span: Span): void {
         this._metrics.spansFinished.increment(1);
-        this._reporter.report(span);
+        try {
+          this._reporter.report(span);
+        } catch (err) {
+          this._logger.error(`could not report span ${span.operationName()}: ${err.message}`)
+        }
     }
 
     registerInjector(format: string, injector: Injector): void {
@@ -182,7 +186,7 @@ export default class Tracer {
         let userTags = options.tags || {};
         let startTime = options.startTime || this.now();
 
-        // This flag is used to ensure that CHILD_OF reference is preferred 
+        // This flag is used to ensure that CHILD_OF reference is preferred
         // as a parent even if it comes after FOLLOWS_FROM reference.
         let followsFromIsParent = false;
         let parent: ?SpanContext = options.childOf instanceof Span ? options.childOf.context() : options.childOf;
--- a/src/tracer.js
+++ b/src/tracer.js
@@ -182,7 +182,7 @@ export default class Tracer {
         let userTags = options.tags || {};
         let startTime = options.startTime || this.now();
 
-        // This flag is used to ensure that CHILD_OF reference is preferred
+        // This flag is used to ensure that CHILD_OF reference is preferred 
         // as a parent even if it comes after FOLLOWS_FROM reference.
         let followsFromIsParent = false;
         let parent: ?SpanContext = options.childOf instanceof Span ? options.childOf.context() : options.childOf;
--- a/src/tracer.js
+++ b/src/tracer.js
@@ -186,7 +186,7 @@ export default class Tracer {
         let userTags = options.tags || {};
         let startTime = options.startTime || this.now();
 
-        // This flag is used to ensure that CHILD_OF reference is preferred
+        // This flag is used to ensure that CHILD_OF reference is preferred 
         // as a parent even if it comes after FOLLOWS_FROM reference.
         let followsFromIsParent = false;
         let parent: ?SpanContext = options.childOf instanceof Span ? options.childOf.context() : options.childOf;
--- a/src/tracer.js
+++ b/src/tracer.js
@@ -139,11 +139,7 @@ export default class Tracer {
 
     _report(span: Span): void {
         this._metrics.spansFinished.increment(1);
-        try {
-          this._reporter.report(span);
-        } catch (err) {
-          this._logger.error(`could not report span ${span.operationName}: ${err.message}`)
-        }
+        this._reporter.report(span);
     }
 
     registerInjector(format: string, injector: Injector): void {
--- a/test/udp_sender.js
+++ b/test/udp_sender.js
@@ -249,4 +249,13 @@ describe('udp sender should', () => {
         assert.equal(response.err, false);
         assert.equal(response.numSpans, 0);
     });
+
+    it ('flush gracefully handles errors emitted by socket.send', done => {
+        sender._client.on('error', err => {
+            done();
+        });
+        sender._host = 'foo.bar.com';
+        sender._port = 1234;
+        sender.flush();
+    });
 });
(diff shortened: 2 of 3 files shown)


src/tracer.js
```diff
@@ -139,7 +139,11 @@ export default class Tracer {
 
     _report(span: Span): void {
         this._metrics.spansFinished.increment(1);
-        this._reporter.report(span);
+        try {
+          this._reporter.report(span);
+        } catch (err) {
+          this._logger.error(`could not report span ${span.operationName}: ${err.message}`)
+        }
```
yurishkuro: while it is tempting to protect all tracer methods with try-catch, doing so at the top level methods leads to Node engine marking them as [non-optimizable](https://community.risingstack.com/how-to-find-node-js-performance-optimization-killers/). The `report` method is called a lot more often than the actual place that was causing your exception, which was around the networking code, so I would rather move try-catch down the stack to the place that _actually can throw an exception_.

yurishkuro:
Also, let's add a unit test for the exception.

frankgreco:
@yurishkuro alright working on it now :)

frankgreco:
@yurishkuro concerning code coverage, is it required that the coverage not drop even if the committed code is fully covered? (see image below):

This committed only contains changes in `upd_sender.js` which is 100% covered.

screen shot 2017-07-19 at 5 32 28 pm

test/udp_sender.js
```diff
@@ -249,4 +250,20 @@ describe('udp sender should', () => {
         assert.equal(response.err, false);
         assert.equal(response.numSpans, 0);
     });
+
+    it ('flush gracefully handles errors emitted by socket.send', () => {
+        sender._client.on('error', err => {
+            assert.equal(err.message, 'getaddrinfo ENOTFOUND foo.bar.com')
```
yurishkuro: Nit: this isn't really testing the new error handler you're adding. The test should simply try to send a span when the sender has the wrong agent address, and without the new handler it should throw uncaught error, but with the handler just log it to console and succeed.

Also, I am not sure you would actually get uncaught error here since the default reporter buffers the spans.

The ideal test would be to validate that `console.log` was called with expected error string.

frankgreco: > Also, I am not sure you would actually get uncaught error here since the default reporter buffers the spans.

I can confirm that my fix indeed solves the problem. Here is a simple reproducible example:

```js
var http = require('http')
var tracer = require('jaeger-client').initTracer({
    'serviceName': 'myService',
    'reporter': {
      logSpans: false,
      agentHost: 'foo.bar.com',
      flushIntervalMs: 10
    },
    'sampler': {
      host: 'foo.bar.com',
      param: 1,
      type: 'const'
    }
  })

http.createServer((req, res) => {
  span = tracer.startSpan('mySpan')
  res.end('response string')
  span.finish()
}).listen('8080')
```

```json
{
  "name": "example",
  "version": "1.0.0",
  "dependencies": {
    "jaeger-client": "3.5.3"
  }
}
```
```sh
$ npm install
$ node index.js
events.js:160
      throw er; // Unhandled 'error' event
      ^

Error: getaddrinfo ENOTFOUND foo.bar.com
    at errnoException (dns.js:28:10)
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:76:26)
```

Now to see the fix in action, navigate to:
```
node_modules/jaeger-client/dist/src/reporters/udp_sender.js
```
and add my fix to line 141:
```js
this._client.on('error', err => {
    console.log(`error sending span: ${err}`)
})
```
now restart the node server and hit `localhost:8080` and you will see the error is handled gracefully.

frankgreco: > Nit: this isn't really testing the new error handler you're adding. The test should simply try to send a span when the sender has the wrong agent address, and without the new handler it should throw uncaught error, but with the handler just log it to console and succeed.

```js
sender._host = 'foo.bar.com';
sender._port = 1234;
let reporter = new RemoteReporter(sender);
tracer = new Tracer(
    'test-service-name',
     reporter,
     new ConstSampler(true)
);
```
the above code (in my test case) is creating as sender that will attempt to send the span to a host that will throw an error on DNS lookup. now if I commented out the event handler in that test case, the test would fail and mirror the error that the issue stemmed from. How the same event emitter that is in the code is in my test and we expect that event emitter to be called which it is. 

The above reasoning is why I believe it is a valid test case: Here are some resources to back up my method for testing event emitters:
https://stackoverflow.com/questions/37232574/node-js-testing-if-event-was-emitted-using-sinon-js-without-the-use-of-timeouts

yurishkuro: I am  not saying that the event is not emitted, rather that the test is not testing the actual event handler. Can you use sinon to mock console.log? If so, you don't have to add another event in the test, and you will validate that the event handler in the sender is actually invoked, by validating that console.log was invoked.

test/udp_sender.js
```diff
@@ -249,4 +251,20 @@ describe('udp sender should', () => {
...
+    it ('flush gracefully handles errors emitted by socket.send', (done) => {
+        let spy = sinon.spy(console, 'log');
+        sender._host = 'foo.bar.com';
+        new Tracer(
+            'test-service-name',
+            new RemoteReporter(sender),
+            new ConstSampler(true)
+        ).startSpan('testSpan').finish();
+        sender.flush();
+        setTimeout(() => {
```
frankgreco: *NOTE*: `setTimeout` is needed due to Node's asynchronous nature. Because I am spying on `console.log`, if `setTimeout` is not used, the error message we expect will be printed to the console but not before mocha prints other messages to the console.

yurishkuro: the 1sec wait is unfortunate. What if you patch console.log directly?

```javascript
var oldLog = console.log;
console.log = function (message) {
    assert(message == ...);
    console.log = oldLog;
    done();
};
// rest of the test
```

test/udp_sender.js
```diff
@@ -262,10 +262,12 @@ describe('udp sender should', () => {
             new ConstSampler(true)
         ).startSpan('testSpan').finish();
         sender.flush();
-        setTimeout(() => {
-            assert(spy.calledWith('error sending span: Error: getaddrinfo ENOTFOUND foo.bar.com', 'error sending span: Error: getaddrinfo ENOTFOUND'));
+        let oldLog = console.log;
+        console.log = message =>  {
```
frankgreco: @yurishkuro how's that

test/udp_sender.js
```diff
@@ -249,4 +251,23 @@ describe('udp sender should', () => {
...
+        sender._host = 'foo.bar.com';
+        sender._port = 1234;
+        new Tracer(
+            'test-service-name',
+            new RemoteReporter(sender),
+            new ConstSampler(true)
+        ).startSpan('testSpan').finish();
+        sender.flush();
+        let oldLog = console.log;
+        console.log = message =>  {
```
yurishkuro: shouldn't this be done _before_ `sender.flush()`?

also, since you're resetting the function back to oldLog, I don't think you need sinon anymore.

test/udp_sender.js
```diff
@@ -249,4 +250,21 @@ describe('udp sender should', () => {
...
+        sender._host = 'foo.bar.com';
+        sender._port = 1234;
+        new Tracer(
+            'test-service-name',
+            new RemoteReporter(sender),
+            new ConstSampler(true)
+        ).startSpan('testSpan').finish();
+        let oldLog = console.log;
+        sender.flush();
+        console.log = message =>  {
```
yurishkuro: still a race condition, is there not? The flush might call the log function before you have a chance to replace it.

frankgreco: No I don't believe so. In the context of replacing `console.log`, the code in question is:

```js
this._client.on('error', err => {
    console.log(`error sending span: ${err}`)
})
```

Now, there is now way this code will be called before `socket.send` has a chance to execute. Now, according to the Node dgram docs where it says, *Note that DNS lookups delay the time to send for at least one tick of the Node.js event loop.* During that time, the next line of code, the code that replaces the logger to our mock one, will have a chance to execute.

frankgreco: regardless, moving `sender.flush()` after the `console.log` assignment on line 264 should remove any doubt.

frankgreco:
@yurishkuro what is Uber's process with cutting new releases. Does a new release get cut every time master is updated?

yurishkuro:
No, the release process is manual. I assume this fix is not too critical to go out?  It will be a new minor release because of some new features since the last release.

frankgreco:
okay thanks for the explanation. No, it's not critical.

PR #138, opened 2017-07-20
Reduce the use of URL encoding

### LINKED ISSUE #41: Reduce the use of url encoding
URL encoding of the tracing headers showed up as a large part of the flame graph. We need to minimize its use
- we do not need to invoke url encoding for encoding trace IDs, because x:y:z:w format is safe for http headers
- when decoding trace IDs, we should first try `split(":")`, and only if that didn't work we can try urlDecode
- for baggage values we should always use url encoding, in each direction

frankgreco:
Resolves #41


tiffon:
@frankgreco This looks good to me, can you resolve the conflict?

frankgreco:
@tiffon Merge conflicts have been resolved. Should be ready to merge.

src/propagators/text_map_codec.js
```diff
@@ -68,12 +68,17 @@ export default class TextMapCodec {
       if (carrier.hasOwnProperty(key)) {
         let lowerKey = key.toLowerCase();
         if (lowerKey === this._contextKey) {
-          let decodedContext = SpanContext.fromString(this._decodeValue(carrier[key]));
+          let decodedContext = SpanContext.fromString(carrier[key]);
+          if (decodedContext !== null) { // trying to create context from split(":") worked ... no need to decode
```
yurishkuro: this can be simpler (one less `continue`) with
```
if (decodedContext == null) {
  decodedContext = SpanContext.fromString(this._decodeValue(carrier[key]));
}
```

PR #142, opened 2017-07-28
Add BaggageRestrictionManager and BaggageSetter

black-adder:
subset of: https://github.com/uber/jaeger-client-node/pull/131


src/baggage/baggage_setter.js
```diff
@@ -0,0 +1,95 @@
...
+     *
+     * @param {Span} span - The span to set the baggage on.
+     * @param {string} key - The baggage key to set.
+     * @param {string} value - The baggage value to set.
+     * @return {SpanContext} - The SpanContext with the baggage set.
+     */
+    setBaggage(span: Span, key: string, value: string): SpanContext {
+        let truncated = false;
+        let prevItem = span.getBaggageItem(key);
+        if (!this._valid) {
```
oibe: I find this use of this._valid confusing.  Where is this._valid set? what makes baggage invalid?

src/baggage/default_baggage_restriction_manager.js
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Creates a BaggageRestrictionManager that allows any baggage key.
+ */
+export default class DefaultBaggageRestrictionManager {
+    _baggageSetter: BaggageSetter;
+
+    constructor(metrics: Metrics, maxValueLength: ?number) {
+        maxValueLength = maxValueLength || DEFAULT_MAX_VALUE_LENGTH;
+        this._baggageSetter = new BaggageSetter(true, maxValueLength, metrics);
```
oibe: nit: Me and you go back and forth on this a lot, but personally I would prefer.
let valid = true;
this._baggageSetter = new BaggageSetter(valid, maxValueLength, metrics);

As of now when I read this code I have to jump to BaggageSetter to know what true means, whereas everything else has a variable name.  Anyway really a nit.

oibe:
Some small nits, and also coverage dropped a bit, but otherwise looks good to me.

src/baggage/baggage_setter.js
black-adder: The collectors will have a list of valid baggage keys and their maximum size. The clients will retrieve these restrictions once a minute. Anytime the client writes baggage, it'll be validated against these rules. If invalid, the baggage is logged in the span but is not propagated.

The remote fetching of baggage rules will come in a future PR. Right now, the DefaultBaggageRestrictionManager will return a BaggageSetter that allows all baggage. ie all baggage is valid.

src/baggage/default_baggage_restriction_manager.js
black-adder: valid

src/baggage/baggage_setter.js
yurishkuro: it's because a better way would be to have different implementations for the Setter

src/tracer.js
```diff
@@ -70,6 +75,11 @@ export default class Tracer {
         this._reporter = reporter;
         this._sampler = sampler;
         this._logger = options.logger || new NullLogger();
+        if (options.baggageRestrictionManager) {
```
oibe: this could be simpler written as:
this._baggageSetter = new BaggageSetter(
    options.baggageRestrictionManager || new DefaultBaggageRestrictionManager(),
    this._metrics
);

test/baggage/baggage_setter.js
```diff
@@ -0,0 +1,128 @@
...
+        span = tracer.startSpan('op-name');
+    });
+
+    afterEach(() => {
+        tracer.close();
+    });
+
+    it ('fail for invalid baggage key', (done) => {
+        let mgr = new DefaultBaggageRestrictionManager();
+        let stub = sinon.stub(mgr, 'getRestriction', function(key) {
```
oibe: I don't think you need "let stub", since the variable is never used.

test/baggage/baggage_setter.js
```diff
@@ -0,0 +1,128 @@
...
+            return new Restriction(false, 0);
+        });
+        let setter = new BaggageSetter(mgr, metrics);
+        let key = "key";
+        let value = "value";
+        let ctx = setter.setBaggage(span, key, value);
```
oibe: nit: s/ctx/spanContext/

test/baggage/default_baggage_restriction_manager.js
```diff
@@ -0,0 +1,32 @@
...
+
+import {assert} from 'chai';
+import DefaultBaggageRestrictionManager from "../../src/baggage/default_baggage_restriction_manager";
+import Restriction from "../../src/baggage/restriction";
+
+describe('DefaultBaggageRestrictionManager should', () => {
+
+    it ('return a Restriction', (done) => {
+        let mgr = new DefaultBaggageRestrictionManager();
+        assert.deepEqual(mgr.getRestriction("key"), new Restriction(true, 2048));
```
oibe: If the default restrictions change this test will change.
Perhaps: assert.deepEqual(mgr.getRestriction("key"), mgr._restriction);

test/span.js
```diff
@@ -149,7 +139,7 @@ describe('span should', () => {
         assert.equal(span._logs[0].fields[0].value, event);
     });
 
-    it('add logs with paylaod', () => {
+    it('add logs with paylod', () => {
```
oibe: s/paylod/payload/

oibe:
After you address the comments I above I would say "LGTM"

test/baggage/default_baggage_restriction_manager.js
black-adder: I believe for things like this we would want it to break because what you're suggesting is tantamount to this.var == this.getVar.

I remember there being a informative Google Testing Blog entry about this but can't find it.

src/span.js
```diff
@@ -95,26 +97,15 @@ export default class Span {
...
+
         // We create a new instance of the context here instead of just adding
         // another entry to the baggage dictionary. By doing so we keep the
         // baggage immutable so that it can be passed to children spans as is.
         // If it was mutable, we would have to make a copy of the dictionary
-        // for every child span, which on average we expect to occur more 
+        // for every child span, which on average we expect to occur more
         // frequently than items being added to the baggage.
-        this._spanContext = this._spanContext.withBaggageItem(normalizedKey, value);
+        this._spanContext = this._baggageSetter.setBaggage(this, normalizedKey, value);
```
tiffon: It might make sense to externally expose a facility to get the normalized key to support custom restriction managers since they're going to be passed in normalized keys and therefore need to sensitive to the transformation.

src/span.js
```diff
@@ -36,6 +36,7 @@ export default class Span {
     _tags: Array<Tag>;
     static _baggageHeaderCache: any;
```
tiffon: A relatively minor thing, but you might consider using a `Map` instead of an object for the baggage header cache. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#Objects_and_maps_compared

test/baggage/baggage_setter.js
```diff
@@ -0,0 +1,128 @@
...
+    let metrics: Metrics;
+    let reporter = new InMemoryReporter();
+    let tracer, span;
+
+    let assertBaggageLogs = function(log, key, value, truncated, override, invalid) {
+        let fields: { [key: string]: string } = {};
+        log.fields.forEach((kv) => {
+           fields[kv.key] = kv.value;
+        });
+        assert.equal(fields['event'], 'baggage');
```
tiffon: Just curious, why not use `fields.event`?

test/baggage/baggage_setter.js
```diff
@@ -0,0 +1,128 @@
...
+        );
+
+        span = tracer.startSpan('op-name');
+    });
+
+    afterEach(() => {
+        tracer.close();
+    });
+
+    it ('fail for invalid baggage key', (done) => {
```
tiffon: Re `done`, is this an [async test](https://mochajs.org/#asynchronous-code)?

test/baggage/baggage_setter.js
```diff
@@ -0,0 +1,128 @@
...
+    beforeEach(() => {
+        metrics = new Metrics(new LocalMetricFactory());
+
+        tracer = new Tracer(
+            'test-service-name',
+            reporter,
+            new ConstSampler(true),
+            {
+                metrics: metrics,
+            }
```
tiffon: With the babel transformations, does `{ metrics }` work?

test/baggage/baggage_setter.js
```diff
@@ -0,0 +1,128 @@
...
+        let key = "key";
+        let value = "value";
+        let spanContext = setter.setBaggage(span, key, value);
+        assert.isUndefined(spanContext._baggage[key]);
+        assertBaggageLogs(span._logs[0], key, value, false, false, true);
+        assert.equal(LocalBackend.counterValue(metrics.baggageUpdateFailure), 1);
+        done();
+    });
+
+    it ('succeed for valid baggage key', (done) => {
```
tiffon: Test case description doesn't mention truncation is also being verified.

test/baggage/default_baggage_restriction_manager.js
```diff
@@ -0,0 +1,32 @@
...
+        done();
```
tiffon: See question above about async test case.

src/_flow/baggage_restriction_manager.js
```diff
@@ -0,0 +1,32 @@
...
+import Restriction from '../baggage/restriction.js'
+
+/**
+ * BaggageRestrictionManager is an interface for a class that manages baggage
+ * restrictions for baggage keys. The manager will return a Restriction
+ * for a specific baggage key which will determine whether the baggage key is
+ * allowed and any other applicable restrictions on the baggage value.
+ */
+declare interface BaggageRestrictionManager {
+    getRestriction(key: string): Restriction;
```
yurishkuro: do you want to add the service name here, similar to Go client?

src/baggage/baggage_setter.js
```diff
@@ -0,0 +1,88 @@
...
+     * @param {string} key - The baggage key to set.
+     * @param {string} value - The baggage value to set.
+     * @return {SpanContext} - The SpanContext with the baggage set if applicable.
+     */
+    setBaggage(span: Span, key: string, value: string): SpanContext {
+        let truncated = false;
+        let prevItem = '';
+        let restriction = this._restrictionManager.getRestriction(key);
+        if (!restriction.keyAllowed) {
+            this._logFields(span, key, value, prevItem, truncated, restriction.keyAllowed);
```
yurishkuro: I don't think sensitive info is a concern - we cannot protect against it in case the user calls the general span.log() method, so why worry about protecting here.

src/baggage/baggage_setter.js
```diff
@@ -0,0 +1,88 @@
...
+            this._metrics.baggageUpdateFailure.increment(1);
+            return span.context();
+        }
+        if (value.length > restriction.maxValueLength) {
+            truncated = true;
+            value = value.substring(0, restriction.maxValueLength);
```
yurishkuro: seems like a stale comment, since `value` is a local var now

src/span.js
yurishkuro: +1, I did't even know JS had a Map. But I'd leave it out of this diff, it sounds like a global refactoring

src/span.js
yurishkuro: we need to get rid of normalizing baggage keys altogether, but it's a larger change across all clients (https://github.com/uber/jaeger/issues/346)

src/tracer.js
```diff
@@ -70,6 +75,8 @@ export default class Tracer {
         this._reporter = reporter;
         this._sampler = sampler;
         this._logger = options.logger || new NullLogger();
+        this._baggageSetter  = new BaggageSetter(options.baggageRestrictionManager || new DefaultBaggageRestrictionManager(),
```
yurishkuro: maybe add `\n` after `new BaggageSetter(` for better readability

test/baggage/baggage_setter.js
yurishkuro: perhaps, but certainly makes the code more cryptic

PR #146, opened 2017-08-08
Fix linter to work with flow

black-adder:
eslint doesn't work with flow, had to install babel-eslint which is compatible with flow. This PR also sorts all imports.

We probably want to land everything everything else before landing this because:
✖ 277 problems (275 errors, 2 warnings)


yurishkuro:
Can you elaborate what exactly this is trying to achieve?

PR #151, opened 2017-08-27
Fix buffer size counting bug and add logging to UDP sender

### LINKED ISSUE #150: Spans may be lost under heavy load
this code is from https://github.com/uber/jaeger-client-node/blob/master/src/reporters/udp_sender.js
when high concurrency, spans may be dropped frequently:

```javascript
    append(span: any): SenderResponse {
        let spanSize: number = this._calcSpanSize(span); 
        if (spanSize > this._maxSpanBytes) {
            return { err: true, numSpans: 1 };
        }

        this._byteBufferSize += spanSize;          // comment 1
        if (this._byteBufferSize <= this._maxSpanBytes) {
            this._batch.spans.push(span);
            if (this._byteBufferSize < this._maxSpanBytes) {
                return {err: false, numSpans: 0};
            }
            return this.flush();
        }

        let flushResponse: SenderResponse = this.flush();  // comment 2
        this._batch.spans.push(span);                                  //comment 3
        this._byteBufferSize = spanSize;
        return flushResponse;
    }
```

here is a example:

this._maxSpanBytes:  64696
a span lenth: 639
current _byteBufferSize: 64671
now run append(span)

at `comment 1`: this._byteBufferSize  will be `64671+639=65310`, it's bigger than `this._maxSpanBytes`
so it will run `comment 2`: `this.flush()`

but in `this.flush()`:

```javascript
flush(): SenderResponse {
        let numSpans: number = this._batch.spans.length;
        if (numSpans == 0) {
            return {err: false, numSpans: 0}
        }

        let bufferLen = this._byteBufferSize + this._emitSpanBatchOverhead;  //comment 4
        let thriftBuffer = new Buffer(bufferLen);   
```
`comment 4` use `this._byteBufferSize`, which now is 65310, which will cause "EMSGSIZE" error, if you add a send callback:

```javascript
this._client.send(thriftBuffer, 0, thriftBuffer.length, this._port, this._host, function(e, sent) {
  console.log(e)
});
```

and the 64671 byte span is dropped!

fix : 

add `this._byteBufferSize -= spanSize;`  before `comment 2`.
or maybe reconstruct it

`this._byteBufferSize += spanSize;` and `this._batch.spans.push(span);   ` should  be atomic, but the code separated them.

yurishkuro:
Resolves #150

### DIFF
--- a/src/_flow/socket.js
+++ b/src/_flow/socket.js
@@ -1,0 +1,33 @@
+// @flow
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+declare class dgram$Socket {
+    on(event: string, listener: Function): this;
+    close(): void;
+    send(
+      msg: Buffer,
+      offset: number,
+      length: number,
+      port: number,
+      address: string,
+      callback?: (err: ?any, bytes: any) => mixed,
+    ): void;
+  };
--- a/src/_flow/socket.js
+++ b/src/_flow/socket.js
@@ -1,5 +1,5 @@
 // @flow
-// Copyright (c) 2016 Uber Technologies, Inc.
+// Copyright (c) 2017 Uber Technologies, Inc.
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
--- a/src/reporters/udp_sender.js
+++ b/src/reporters/udp_sender.js
@@ -23,6 +23,7 @@ import dgram from 'dgram';
 import fs from 'fs';
 import path from 'path';
 import {Thrift} from 'thriftrw';
+import NullLogger from '../logger.js';
 
 const HOST = 'localhost';
 const PORT =  6832;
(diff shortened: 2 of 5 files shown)


src/_flow/socket.js
```diff
@@ -0,0 +1,33 @@
+// @flow
+// Copyright (c) 2016 Uber Technologies, Inc.
```
black-adder: we should update the license.sh to say 2017

src/_flow/socket.js
```diff
@@ -0,0 +1,33 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+declare class dgram$Socket {
```
oibe: I've never seen this naming convention.  Is this part of ES6's style?

src/reporters/udp_sender.js
```diff
@@ -99,18 +105,19 @@ export default class UDPSender {
             return { err: true, numSpans: 1 };
         }
 
-        this._byteBufferSize += spanSize;
-        if (this._byteBufferSize <= this._maxSpanBytes) {
+        if (this._totalSpanBytes + spanSize <= this._maxSpanBytes) {
             this._batch.spans.push(span);
-            if (this._byteBufferSize < this._maxSpanBytes) {
+            this._totalSpanBytes += spanSize;
+            if (this._totalSpanBytes < this._maxSpanBytes) {
```
black-adder: why is this extra logic is needed? ie if totalspanbytes == maxspanbytes, the next time round it will be flushed on L118. Do we really need this extra check that has only a 1 byte tolerance?

src/reporters/udp_sender.js
```diff
@@ -120,23 +127,24 @@ export default class UDPSender {
...
         let thriftBuffer = new Buffer(bufferLen);
-        let bufferResult = this._agentThrift.Agent.emitBatch.argumentsMessageRW.writeInto(
+        let writeResult = this._agentThrift.Agent.emitBatch.argumentsMessageRW.writeInto(
             this._convertBatchToThriftMessage(this._batch), thriftBuffer, 0
         );
 
-        if (bufferResult.err) {
-            console.log('err', bufferResult.err);
+        if (writeResult.err) {
+            this._logger.error(`error writing Thrift object: ${writeResult.err}`);
```
oibe: Do we want to emit a metric here?

src/reporters/udp_sender.js
```diff
@@ -120,23 +127,24 @@ export default class UDPSender {
...
-        this._client.on('error', err => {
-            console.log(`error sending span: ${err}`)
-        })
-
-        this._client.send(thriftBuffer, 0, thriftBuffer.length, this._port, this._host);
+        // Having the error callback here does not prevent uncaught exception from being thrown,
+        // that's why in the constructor we also add a general on('error') handler.
+        this._client.send(thriftBuffer, 0, thriftBuffer.length, this._port, this._host, (err, sent) => {
+            if (err) {
+                this._logger.error(`error sending spans over UDP: ${err}, packet size: ${writeResult.offset}, bytes sent: ${sent}`);
```
oibe: same as above.

oibe:
For the most part looks good to me.  The only real concern is whether or not we should emit a metric on error logs.

src/reporters/udp_sender.js
yurishkuro: I think you're right. We can just check for strict `<` in L108 and return without flushing

yurishkuro: * if `total + spanSize` is less than max packet size, we don't flush (diff could be 1 byte or 50Kb, whichever)
  * if `total + spanSize` is exactly the max packet size, we can flush the buffer _including_ the current span, rather than flushing a smaller buffer and keeping the current span for the next flush

minor optimization, but seems worth it given that it doesn't complicate the code that much.

src/reporters/udp_sender.js
yurishkuro: the sender is not integrated with metrics currently. Also the two new errors are a sign of misconfiguration, not something that can keep happening repeatedly, thus I think the logs are a lot more useful.

Issue #123, opened 2017-06-02
Enabling logSpans flag causes exception


yurishkuro:
The logging reporter writes the span with `JSON.stringify`, which fails because of some recursive links in the span.

Instead, it should log the span in the same format as other clients, as trace context + operation name.

rajeshsegu:
Is there a reason why we are not using just `util.inspect` since this is a node only framework

```
util.inspect(span, {depth: 3});
```

yurishkuro:
I don't see the need for logging any details of the span itself.  For example, in Go client the log looks like this
```
2017/09/22 20:26:49 Reporting span 5642914c078ef2f0:5642914c078ef2f0:0:1
```

PR #153, opened 2017-09-27
Log only span context

yurishkuro:
Resolves #123

### DIFF
--- a/src/reporters/logging_reporter.js
+++ b/src/reporters/logging_reporter.js
@@ -30,7 +30,7 @@ export default class LoggingReporter {
     }
 
     report(span: Span): void {
-        this._logger.info(`Reporting span ${JSON.stringify(span)}`);
+        this._logger.info(`Reporting span ${span.context().toString()}`);
     }
 
     name(): string {
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "lodash": "^4.15.0",
     "minimist": "1.2.0",
     "mocha": "^3.0.1",
-    "request": "^2.74.0",
+    "request": "2.74.0",
     "rsvp": "^3.3.1",
     "semver": "^5",
     "sinon": "^1.17.5",
--- a/test/all_reporters.js
+++ b/test/all_reporters.js
@@ -78,14 +78,22 @@ describe('All Reporters should', () => {
     });
 
     describe('Logging reporter', () => {
-        it('report span logs span as a stringified object', () => {
+        it('logs span as context().toString()', () => {
             let logger = new MockLogger();
             let reporter = new LoggingReporter(logger);
-            let spanMock = { key: 'some-span' };
+            let spanMock = {
+                context: function context() {
+                    return {
+                        toString: function toString() {
+                            return "span-as-string";
+                        }
+                    };
+                }
+            };
 
             reporter.report(spanMock);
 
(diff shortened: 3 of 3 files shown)


yurishkuro:
@tiffon @saminzadeh could you please take a look at the build failures? All seem to be related to the error below, a transitive dependency of `request` module. There was a recent change there https://github.com/request/request/pull/2751. How can we fix it if we're using babel?

```
> mocha --compilers js:babel-register crossdock/test
/home/travis/build/uber/jaeger-client-node/node_modules/request/node_modules/hawk/node_modules/boom/lib/index.js:5
const Hoek = require('hoek');
^^^^^
SyntaxError: Use of const in strict mode.
    at Module._compile (module.js:439:25)
    at Module._extensions..js (module.js:474:10)
    at Object.require.extensions.(anonymous function) [as .js] (/home/travis/build/uber/jaeger-client-node/node_modules/babel-register/lib/node.js:152:7)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Module.require (module.js:364:17)
    at require (module.js:380:17)
```

tiffon:
@yurishkuro Looks related to running node `0.10.x` which doesn't support `const` without a flag. [Details.](https://stackoverflow.com/questions/22603078/syntaxerror-use-of-const-in-strict-mode)

I would say we can either file a bug with request because their upgrade is a breaking change or change our `package.json` to get rid of the `^` in `"request": "^2.74.0"`. Generally, I would lean toward the latter or both.

src/reporters/logging_reporter.js
```diff
@@ -30,7 +30,7 @@ export default class LoggingReporter {
     }
 
     report(span: Span): void {
-        this._logger.info(`Reporting span ${JSON.stringify(span)}`);
+        this._logger.info(`Reporting span ${span.context().toString()}`);
```
tiffon: Interpolated values are automatically converted to strings:

```javascript
`${{toString: () => 'abc'}}`
// -> "abc"
```

yurishkuro: I don't follow. IIRC Javascript does not have Java's convention of automatically calling `toString()` method when converting to strings.

yurishkuro:
> Looks related to running node `0.10.x` which doesn't support `const` without a flag.

The `request` module declares `"node": ">= 4"` in its package.json, so they are free to say they do not intend to support 0.10.

It seems like we're only using `request` in the crossdock tests, couldn't we use something else that is 0.10 compatible?

src/reporters/logging_reporter.js
tiffon: JS does call `toString()` when converting to string.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/toString

>Every object has a toString() method that is automatically called when the object is to be represented as a text value or when an object is referred to in a manner in which a string is expected.

saminzadeh:
@yurishkuro Can't we just build the src code with Node 4 (or above) and pipe it through babel so its compatible with Node 0.10?

yurishkuro:
@saminzadeh we can, but it makes the build a lot more complicated because we want the unit and integration tests run on the actual Node version from Travis matrix. Besides, this `request` module is a runtime dependency of the crossdock tests, so I don't believe what you propose would've helped with the `const` issue.

PR #161, opened 2017-10-23
Randomize rater limiter balance on initialization

black-adder:


black-adder:
yes, but that would be a breaking change :(

yurishkuro:
no, it would be an optional parameter, Javascript is like that.

black-adder:
hmmm, makes sense, ok will make the change.

Issue #32, opened 2016-10-12
Change SenderResponse.err to be a message string


yurishkuro:
We're using this to log, so it's better to log the actual underlying failure.

tiffon:
Is there a reason `err` should be a `string` instead of an `Error`?

PR #166, opened 2017-10-31
Fix SenderResponse.err to be a message string

MarckK:
Issue #32

### DIFF
--- a/src/reporters/udp_sender.js
+++ b/src/reporters/udp_sender.js
@@ -126,8 +126,8 @@ export default class UDPSender {
         );
 
         if (writeResult.err) {
-            this._logger.error(`error writing Thrift object: ${writeResult.err}`);
-            return {err: true, numSpans: numSpans};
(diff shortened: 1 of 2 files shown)


src/reporters/udp_sender.js
```diff
@@ -126,8 +126,8 @@ export default class UDPSender {
         );
 
         if (writeResult.err) {
-            this._logger.error(`error writing Thrift object: ${writeResult.err}`);
-            return {err: true, numSpans: numSpans};
+            let errMessage = this._logger.error(`error writing Thrift object: ${writeResult.err}`);
```
yurishkuro: logger doesn't return a string afaik. Can you add a unit test that verifies that the error is returned as a string?

tiffon: At the very least, the `NullLogger` doesn't:

https://github.com/jaegertracing/jaeger-client-node/blob/8b783aaf26ac007d2f064ad72904d91fddcf020a/src/logger.js#L14-L16

yurishkuro:
superseded by #226

Issue #165, opened 2017-10-30
Support Zipkin HTTP B3 propagation format


dcarney:
Support for Zipkin-style `X-B3-*` headers has already been added to the [go client](https://github.com/jaegertracing/jaeger-client-go/tree/master/zipkin) and the [java client](https://github.com/jaegertracing/jaeger-client-java/tree/master/jaeger-zipkin).  It appears to be straightforward to add such support to the node client as well.

Per the [contribution guidelines](https://github.com/jaegertracing/jaeger-client-node/blob/master/CONTRIBUTING.md), I'm opening an issue before submitting a PR to add this functionality.

The bulk of the new "propagator" implementation for Zipkin-style headers would be similar to the current `TextMapCodec`.  I've got a working prototype (I called it a `ZipkinB3TextMapCodec`, after the [go implementation of the same name](https://github.com/jaegertracing/jaeger-client-go/tree/master/zipkin)) that I could clean up and submit in a PR.

yurishkuro:
+1

afoninsky:
> I've got a working prototype
@dcarney  could you please point to this prototype?

dcarney:
@afoninsky Thanks for the reminder!  I'm cleaning up my PR right now, and will submit it today.

dcarney:
Submitted PR #175

yurishkuro:
Thanks!

yurishkuro:
Released as 3.7.0

PR #175, opened 2017-11-13
Adds support for the Zipkin HTTP B3 propagation format

dcarney:
This PR addresses #165 by adding a new text map codec.  

The new codec (named `ZipkinB3TextMapCodec`) injects and extracts the relevant `X-B3-*` headers used by zipkin, according to [the zipkin spec](https://github.com/openzipkin/b3-propagation).

The new codec can be used by registering it with a tracer instance as both an injector and an extractor:

```
let codec = new ZipkinB3TextMapCodec({                                                            
    urlEncoding: true                                                                          
}); 

tracer.registerInjector(opentracing.FORMAT_HTTP_HEADERS, codec);                                  
tracer.registerExtractor(opentracing.FORMAT_HTTP_HEADERS, codec);
```

In order to accomodate this functionality, I added this new codec (as well as the existing `TextMapCodec` ) to the exported entrypoint of the lib.  If this is not a preferred change to the lib, please let me know.

Baggage handling was left unchanged from the way it's currently done in the `TextMapCodec`.

If a `README` is desired for this new Zipkin support (as is done in [the Jaeger Go client](https://github.com/jaegertracing/jaeger-client-go/tree/master/zipkin)), I'd be happy to add one.

Thanks in advance for feedback on the PR!

### DIFF
--- a/src/propagators/zipkin_b3_text_map_codec.js
+++ b/src/propagators/zipkin_b3_text_map_codec.js
@@ -21,7 +21,7 @@ const ZIPKIN_PARENTSPAN_HEADER = 'x-b3-parentspanid';
 const ZIPKIN_SPAN_HEADER = 'x-b3-spanid';
 const ZIPKIN_TRACE_HEADER = 'x-b3-traceid';
 const ZIPKIN_SAMPLED_HEADER = 'x-b3-sampled';
-const ZIPKIN_DEBUG_HEADER = 'x-b3-flags';
+const ZIPKIN_FLAGS_HEADER = 'x-b3-flags';
 
 export default class ZipkinB3TextMapCodec {
     _urlEncoding: boolean;
@@ -82,7 +82,7 @@ export default class ZipkinB3TextMapCodec {
                     case ZIPKIN_SAMPLED_HEADER:
                         spanContext.flags = spanContext.flags | constants.SAMPLED_MASK;
                         break;
-                    case ZIPKIN_DEBUG_HEADER:
+                    case ZIPKIN_FLAGS_HEADER:
                         // "debug implies sampled"
                         // https://github.com/openzipkin/b3-propagation
                         spanContext.flags = spanContext.flags | constants.SAMPLED_MASK | constants.DEBUG_MASK;
@@ -115,7 +115,7 @@ export default class ZipkinB3TextMapCodec {
         // https://github.com/openzipkin/b3-propagation
 
         if (spanContext.isDebug()) {
-           carrier[ZIPKIN_DEBUG_HEADER] = '1';
+           carrier[ZIPKIN_FLAGS_HEADER] = '1';
         } else {
             if (spanContext.isSampled()) {
                 carrier[ZIPKIN_SAMPLED_HEADER] = '1';
--- a/src/propagators/zipkin_b3_text_map_codec.js
+++ b/src/propagators/zipkin_b3_text_map_codec.js
@@ -1,0 +1,157 @@
+// @flow
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import * as constants from '../constants.js';
+import Metrics from '../metrics/metrics.js';
+import NoopMetricFactory from '../metrics/noop/metric_factory';
+import SpanContext from '../span_context.js';
+import Utils from '../util.js';
+
+const ZIPKIN_PARENTSPAN_HEADER = 'x-b3-parentspanid';
+const ZIPKIN_SPAN_HEADER = 'x-b3-spanid';
+const ZIPKIN_TRACE_HEADER = 'x-b3-traceid';
+const ZIPKIN_SAMPLED_HEADER = 'x-b3-sampled';
+const ZIPKIN_DEBUG_HEADER = 'x-b3-flags';
+
+export default class ZipkinB3TextMapCodec {
+    _urlEncoding: boolean;
+    _baggagePrefix: string;
+    _metrics: any;
+
+    constructor(options: any = {}) {
+        this._urlEncoding = !!options.urlEncoding;
+        this._baggagePrefix = options.baggagePrefix || constants.TRACER_BAGGAGE_HEADER_PREFIX;
+        this._baggagePrefix = this._baggagePrefix.toLowerCase();
+        this._metrics = options.metrics || new Metrics(new NoopMetricFactory());
+    }
+
+    _encodeValue(value: string): string {
+        if (this._urlEncoding) {
+            return encodeURIComponent(value);
+        }
+
+        return value;
+    }
+
+    _decodeValue(value: string): string {
+        // only use url-decoding if there are meta-characters '%'
+        if (this._urlEncoding && value.indexOf('%') > -1) {
+            return this._decodeURIValue(value);
+        }
+
+        return value;
+    }
+
+    _decodeURIValue(value: string): string {
+        // unfortunately, decodeURIComponent() can throw 'URIError: URI malformed' on bad strings
+        try {
+            return decodeURIComponent(value);
+        } catch (e) {
+            return value;
+        }
+    }
+
+    extract(carrier: any): ?SpanContext {
+        let spanContext = new SpanContext();
+        let baggage = {};
+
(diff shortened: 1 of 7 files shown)


src/propagators/zipkin_b3_text_map_codec.js
```diff
@@ -0,0 +1,157 @@
+// @flow
+// Copyright (c) 2016 Uber Technologies, Inc.
```
yurishkuro: 2017 The Jaeger Authors

src/propagators/zipkin_b3_text_map_codec.js
```diff
@@ -0,0 +1,157 @@
...
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
```
yurishkuro: this looks like MIT license, it should be Apache

src/propagators/zipkin_b3_text_map_codec.js
```diff
@@ -0,0 +1,157 @@
...
+import Metrics from '../metrics/metrics.js';
+import NoopMetricFactory from '../metrics/noop/metric_factory';
+import SpanContext from '../span_context.js';
+import Utils from '../util.js';
+
+const ZIPKIN_PARENTSPAN_HEADER = 'x-b3-parentspanid';
+const ZIPKIN_SPAN_HEADER = 'x-b3-spanid';
+const ZIPKIN_TRACE_HEADER = 'x-b3-traceid';
+const ZIPKIN_SAMPLED_HEADER = 'x-b3-sampled';
+const ZIPKIN_DEBUG_HEADER = 'x-b3-flags';
```
yurishkuro: s/DEBUG/FLAGS/ ?

src/propagators/zipkin_b3_text_map_codec.js
```diff
@@ -0,0 +1,157 @@
...
+                        break;
+                    case ZIPKIN_TRACE_HEADER:
+                        spanContext.traceId = this._decodeValue(carrier[ZIPKIN_TRACE_HEADER]);
+                        break;
+                    case ZIPKIN_SAMPLED_HEADER:
+                        spanContext.flags = spanContext.flags | constants.SAMPLED_MASK;
+                        break;
+                    case ZIPKIN_DEBUG_HEADER:
+                        // "debug implies sampled"
+                        // https://github.com/openzipkin/b3-propagation
```
yurishkuro: not clear - the header string is "flags", not debug

src/propagators/zipkin_b3_text_map_codec.js
```diff
@@ -0,0 +1,157 @@
...
+        carrier[ZIPKIN_TRACE_HEADER] = spanContext.traceIdStr;
+        carrier[ZIPKIN_PARENTSPAN_HEADER] = spanContext.parentIdStr;
+        carrier[ZIPKIN_SPAN_HEADER] = spanContext.spanIdStr;
+
+        // > Since Debug implies Sampled, so don't also send "X-B3-Sampled: 1"
+        // https://github.com/openzipkin/b3-propagation
+
+        if (spanContext.isDebug()) {
+           carrier[ZIPKIN_DEBUG_HEADER] = '1';
+        } else {
```
yurishkuro: why `else` rather than independently?

src/propagators/zipkin_b3_text_map_codec.js
```diff
@@ -0,0 +1,157 @@
...
+            }
+        }
+    }
+
+    // Parse comma separated key=value pair list and add to baggage.
+    // E.g. "key1=value1, key2=value2, key3 = value3"
+    // is converted to map[string]string { "key1" : "value1",
+    //                                     "key2" : "value2",
+    //                                     "key3" : "value3" }
+    _parseCommaSeparatedBaggage(baggage: any, values: string): void {
```
yurishkuro: maybe refactor it and share with the main jaeger propagator?

test/propagators.js
```diff
@@ -52,3 +54,91 @@ describe ('TextMapCodec', () => {
...
+            'x-b3-traceid': 'some-trace',
+            'x-b3-sampled': '1',
+            'x-b3-flags': '1',
+            'foo': 'bar'
+        };
+
+        let ctx = codec.extract(carrier);
+        assert.equal(ctx.parentId, 'some-parent');
+        assert.equal(ctx.spanId, 'some-span');
+        assert.equal(ctx.traceId, 'some-trace');
```
yurishkuro: I think the codec should be returning a null or empty context because the values are not valid Zipkin IDs. Otherwise the thrift serialization will fail later anyway.

src/propagators/zipkin_b3_text_map_codec.js
dcarney: Sure, sounds good. I wasn't sure how far to dive down the refactoring path just yet.

src/propagators/zipkin_b3_text_map_codec.js
dcarney: Based on [the zipkin propagation spec](https://github.com/openzipkin/b3-propagation#debug-flag):

> Debug is encoded as X-B3-Flags: 1. Since Debug implies Sampled, so don't also send "X-B3-Sampled: 1".

So, the `inject` call only sets the sampled header if debug is false.  I've modified the code to make this a little more clear.

test/propagators.js
dcarney: Codec now returns an empty context, in the same manner as the `TextMapCodec`.

src/propagators/zipkin_b3_text_map_codec.js
```diff
@@ -0,0 +1,168 @@
...
+        // using it to validate that the string is a valid hex number (which is
+        // faster than doing it manually).  We cannot use
+        // Int64(numberValue).toBuffer() because it throws exceptions on bad
+        // strings.
+        if (!value) {
+            return true
+        }
+
+        let v = parseInt(value, 16);
+        return !isNaN(v) && v.toString(16) === value;
```
yurishkuro: are you sure about the last `===` term? Given the comment about precision loss I don't think the toString transformation has to return the same original value. For example
```
v = parseInt('ffffffffffffffff', 16);
18446744073709552000
v
18446744073709552000
v.toString(16)
"10000000000000000"
```

test/zipkin_b3_text_map_codec.js
```diff
@@ -0,0 +1,159 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
yurishkuro: wrong copyright

src/propagators/zipkin_b3_text_map_codec.js
dcarney: Sure, I'll remove that.  

I noticed that in `TextMapCodec`, the "validation" of the trace/span IDs is done by just calling `parseInt(foo, 16)` and checking for `NaN`.  In addition to having the limitations mentioned in the code already, this allows values that aren't strictly composed of hex digits to be accepted as valid IDs, since `parseInt` has [the following behavior](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt):

> If parseInt encounters a character that is not a numeral in the specified radix, it ignores it and all succeeding characters and returns the integer value parsed up to that point. 

So, you get the following behavior from `parseInt`:
```
> parseInt('a', 16)
10
> parseInt('azzzzz', 16)
10
```

yurishkuro: +1. It's unfortunate that parseInt behaves this way, but in the worse case it will treat invalid string as a valid trace id (which is somewhat similar to creating a new random trace ID), whereas doing the reverse `===` comparison will discard many valid IDs.

dcarney: Yes, agreed. If you wanted to be super strict, you'd validate all the trace IDs with a regex for lower hex digits or something, but it's likely not worth the overhead.

yurishkuro:
> Do you have any desire for me to add README details to this PR for enabling the Zipkin codec with a tracer?

Yes, it would be good to add, since you have all that handy right now.  You may also add an entry to the changelog.

dcarney:
@yurishkuro Added the `README` section.  [Per your release guidelines](https://github.com/jaegertracing/jaeger-client-node/blob/master/RELEASE.md), I can open a new PR for releasing a new version that would include the CHANGELOG entry.  Seems like minor version bump to me.

src/propagators/zipkin_b3_text_map_codec.js
```diff
@@ -0,0 +1,176 @@
...
+        // Note: The Number type in JS cannot represent the full range of 64bit
+        // unsigned ints, so using parseInt() on strings representing 64bit hex
+        // numbers only returns an approximation of the actual value.
+        // Fortunately, we do not depend on the returned value, we are only
+        // using it to validate that the string is a valid hex number (which is
+        // faster than doing it manually).  We cannot use
+        // Int64(numberValue).toBuffer() because it throws exceptions on bad
+        // strings.
+        if (!value) {
+            return true
```
yurishkuro: return false?

yurishkuro:
@dcarney I have just one comment about the `return true`, otherwise ready to merge. Could you please rebase off master to remove the iojs build step, hopefully it will result in a green build.

src/propagators/zipkin_b3_text_map_codec.js
dcarney: I return true here since an empty string is technically a valid value to find in a zipkin header. (`ParentSpanId` can be absent [for instance](https://github.com/openzipkin/b3-propagation#parentspanid)).  I'm only using the return value of this method in the `extract`method, to determine if any invalid header values were found.

dcarney:
@yurishkuro I rebased the PR off the latest upstream master.  Build looks good now!

yurishkuro:
Thanks!

dcarney:
@yurishkuro Awesome, thanks!  I created #176 as well, following the instructions in the [release process](https://github.com/jaegertracing/jaeger-client-node/blob/master/RELEASE.md).

yurishkuro:
Resolves #165

PR #184, opened 2017-12-07
Log error when Thrift conversion fails

black-adder:

Kinda fixes https://github.com/jaegertracing/jaeger-client-node/issues/181 in that it elucidates why the spanSize is undefined.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,8 +18,7 @@
     "node-int64": "^0.4.0",
     "thriftrw": "^3.5.0",
     "xorshift": "^0.2.0",
-    "opentracing": "^0.13.0",
-    "bufrw":"^1.2.1"
+    "opentracing": "^0.13.0"
   },
   "devDependencies": {
     "babel-cli": "^6.11.4",
--- a/package.json
+++ b/package.json
@@ -18,7 +18,8 @@
     "node-int64": "^0.4.0",
     "thriftrw": "^3.5.0",
     "xorshift": "^0.2.0",
-    "opentracing": "^0.13.0"
+    "opentracing": "^0.13.0",
+    "bufrw":"^1.2.1"
   },
   "devDependencies": {
     "babel-cli": "^6.11.4",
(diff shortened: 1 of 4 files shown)


src/reporters/udp_sender.js
```diff
@@ -93,6 +93,10 @@ export default class UDPSender {
 
     append(span: any): SenderResponse {
         let spanSize: number = this._calcSpanSize(span);
+        if (spanSize == undefined) {
```
yurishkuro: nit: should this be `===`? I've seen people use that with undefined.

src/reporters/udp_sender.js
```diff
@@ -93,6 +93,10 @@ export default class UDPSender {
...
+            this._logger.error(`error converting span to Thrift: ${JSON.stringify(span, null, 4)}`);
```
yurishkuro: definitely better than not having that error, but still doesn't explain why the conversion failed. 

The actual logic is `return this._jaegerThrift.Span.rw.byteLength(new this._jaegerThrift.Span(span)).length;` - it seems it returns a struct that has `.length` property, but if the conversion failed it probably has a more descriptive error in some other property. Maybe change _calcSpanSize to return that object and check it for error instead of `.length == undefined`

src/reporters/udp_sender.js
```diff
@@ -15,6 +15,7 @@ import dgram from 'dgram';
 import fs from 'fs';
 import path from 'path';
 import {Thrift} from 'thriftrw';
+import {LengthResult} from 'bufrw';
```
yurishkuro: you are not using this class, no need to import it. You are only using its declaration, which you need to add to `_flow` dir.

package.json
```diff
@@ -18,7 +18,8 @@
     "node-int64": "^0.4.0",
     "thriftrw": "^3.5.0",
     "xorshift": "^0.2.0",
-    "opentracing": "^0.13.0"
+    "opentracing": "^0.13.0",
+    "bufrw":"^1.2.1"
```
yurishkuro: don't need this

PR #186, opened 2017-12-07
Remove test-ci script

MarckK:
`test-ci` script is possibly redundant. It runs `npm run test && npm run lint`; however, `npm run test` runs `make test`, which runs `npm run lint`. Thus, running `npm run lint` again is possibly redundant, hence this whole script is redundant, and one could just run `npm run test`.


yurishkuro:
it's called from .travis

PR #188, opened 2017-12-07
Remove test-ci script

MarckK:

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,6 @@
     "flow": "flow",
     "lint": "eslint $(ls src/ | grep '.js$') && echo '# linter passed' && npm run check-license",
     "test": "make test",
-    "test-ci": "npm run test && npm run lint",
     "test-all": "npm run test-core && npm run test-samplers && npm run test-crossdock && npm run test-baggage",
     "test-core": "mocha --compilers js:babel-core/register test",
     "test-samplers": "mocha --compilers js:babel-core/register test/samplers",
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,7 +44,7 @@ before_install:
 - if [ "$CROSSDOCK" = "1" ]; then make install_docker_ci ; fi
 
 script:
-- if [ "$CROSSDOCK" != "1" ]; then npm run test-ci ; fi
+- if [ "$CROSSDOCK" != "1" ]; then make test ; fi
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -60,7 +60,6 @@
     "flow": "flow",
     "lint": "eslint $(ls src/ | grep '.js$') && echo '# linter passed' && npm run check-license",
     "test": "make test",
-    "test-ci": "npm run test && npm run lint",
```
yurishkuro: it is used in .travis

MarckK: OK, I'm definitely learning a lot about Travis

MarckK: In the .travis.yml we could change `if [ "$CROSSDOCK" != "1" ]; then npm run test-ci ; fi` to 
`if [ "$CROSSDOCK" != "1" ]; then npm run test ; fi` or for that matter, `if [ "$CROSSDOCK" != "1" ]; then make test ; fi`.

This PR was mostly to check if CI would hang for a PR from a completely new fork. However, `test-ci` script is possibly redundant as it runs `npm run test && npm run lint`; however, `npm run test` runs `make test`, which runs `npm run lint`. Thus, running `npm run lint` again is possibly redundant, hence this whole script is redundant, and one could just run `npm run test`.

black-adder: Good point, there's some clear duplication of work in travis that can be removed. Do you mind repurposing this PR to remove some of the wasted work?

Ideally, for
```
node_js:
- '6'
- '5'
- '4'
- '0.12'
```
We should only run the tests. For
```
- node_js: '0.10'
    env: COVER=1
    sudo: required
    services:
    - docker
```
We should run tests + lint + cover. This should improve travis runtimes. One thing I noticed was that running coveralls actually runs the tests themselves so maybe we can remove the duplication of work there too, but doesn't have to be in this PR since I don't know the work involved.

yurishkuro:
@black-adder We still need to figure our why Kara's PRs are not triggering Travis builds.

MarckK:
I emailed Travis yesterday and they got back with: "Our system automatically flagged your account for possible abuse, but after further review, we were able to clear that up".  I really don't know what to say about that. However, I'll make some more commits--repurposing the PR re @black-adder's suggestion--and hopefully all will be fine :)

PR #192, opened 2017-12-12
Throw error if serviceName is not provided

black-adder:
This was the cause of https://github.com/jaegertracing/jaeger-client-node/issues/181

Since serviceName is not provided, the jaeger batch to thrift conversion fails.

### DIFF
--- a/src/configuration.js
+++ b/src/configuration.js
@@ -153,6 +153,8 @@ export default class Configuration {
         }
         if (config.disable) {
             return new opentracing.Tracer();
+        } else if (!config.serviceName) {
+            throw new Error(`config.serviceName must be provided`);
         } else {
             if (config.sampler) {
                 sampler = Configuration._getSampler(config);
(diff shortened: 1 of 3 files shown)


src/configuration.js
```diff
@@ -173,6 +173,10 @@ export default class Configuration {
             );
         }
 
+        if (config.serviceName === undefined || config.serviceName === null) {
```
black-adder: @tiffon is this the best way to check whether the serviceName was not set in node?

yurishkuro: * I think you want this at the very top of the function
 * can you not just check `if (!config.serviceName)`? An empty string as a service name is also unacceptable.

src/configuration.js
```diff
@@ -153,6 +153,8 @@ export default class Configuration {
         }
         if (config.disable) {
             return new opentracing.Tracer();
+        } else if (!config.serviceName) {
+            throw new Error(`config.serviceName must be provided`);
         } else {
```
yurishkuro: can we break apart the if statements?
```
if disabled
  return
if no serviceName
  throw
// continue
```

yurishkuro:
there seems to be a bug in .travis
```
after_failure:
- timeout 5 if [ "$CROSSDOCK" = "1" ]; then make crossdock-fresh ; fi
```

was that supposed to be `make crossdock-logs`? And it fails in any case

image

yurishkuro:
I think `timeout 5` should be after `then`

black-adder:
looking at the logs, it looks like all the services just about finish initializing. Not sure why the test_driver times out. Maybe increasing the test_driver wait time might make it less flaky.

PR #197, opened 2018-01-08
Use babel-preset-env instead of babel-preset-2015

### LINKED ISSUE #154: Compiler warning about babel-preset-es2015@6.24.1
happens during `npm install`

```
npm WARN deprecated babel-preset-es2015@6.24.1: We're super 😸  excited that you're trying to
use ES2015 syntax, but instead of continuing yearly presets 😭 , we recommend using 
babel-preset-env: npm install babel-preset-env. preset-env without options will compile ES2015+
down to ES5. And by targeting specific browsers, Babel can do less work and you can ship native 
ES2015+ to users 😎 ! Also, we are in the process of releasing v7, so give 
http://babeljs.io/blog/2017/09/12/planning-for-7.0 a read and test it! Thanks so much for using 
Babel 🙏 , please give us a follow @babeljs for updates, join slack.babeljs.io for discussion and help 
support at opencollective.com/babel
```

MarckK:
Resolves #154

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "check-license": "./scripts/check-license.sh",
     "check-ls":
       "npm ls --loglevel=http --parseable 1>/dev/null && echo '# npm is in a good state'",
+    "compile:es2015": "babel ./source --out-dir ./es2015 --presets es2015",
     "cover":
       "babel-node ./node_modules/.bin/babel-istanbul cover ./node_modules/.bin/_mocha -- test/ test/samplers/ test/baggage/",
     "coveralls": "npm run cover && cat ./coverage/lcov.info | coveralls",
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "babel-istanbul": "^0.11.0",
     "babel-plugin-transform-class-properties": "^6.11.5",
     "babel-plugin-transform-flow-strip-types": "^6.8.0",
-    "babel-preset-es2015": "^6.13.2",
+    "babel-preset-env": "^1.6.1",
     "babel-register": "^6.11.6",
     "beautify-benchmark": "^0.2.4",
     "benchmark": "^2.1.1",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -53,13 +53,17 @@
...
+      "babel-node ./node_modules/.bin/babel-istanbul cover ./node_modules/.bin/_mocha -- test/ test/samplers/ test/baggage/",
     "coveralls": "npm run cover && cat ./coverage/lcov.info | coveralls",
     "flow": "flow",
-    "lint": "eslint $(ls src/ | grep '.js$') && echo '# linter passed' && npm run check-license",
+    "lint":
+      "eslint $(ls src/ | grep '.js$') && echo '# linter passed' && npm run check-license",
     "test": "make test",
-    "test-all": "npm run test-core && npm run test-samplers && npm run test-crossdock && npm run test-baggage",
+    "test-all":
+      "npm run test-core && npm run test-samplers && npm run test-crossdock && npm run test-baggage",
```
yurishkuro: can we do without this pure formatting change?

yurishkuro:
@MarckK notice that these changes are breaking the build on Node 0.10 / 0.12. Having the pre-compilation tooling work with 0.10 might be a burden going forward. Maybe we should consider the same approach as in opentracing-javascript, where the tooling is using newer versions, but once the transpiling is done, the final tests run with different runtimes.

package.json
MarckK: It does seem superfluous. However, the formatting changes are being done automatically by [Prettier](https://github.com/prettier/prettier), which is fantastic and should arguably be adopted on all JS projects. Up to now, I've just been toggling the package to disabled and then re-enabled in my editor. But I've been thinking we should consider [using Prettier](https://prettier.io/) in this project. [A number of 'big name' projects use it,](https://github.com/prettier/prettier/issues/1351) including Babel, and projects at Uber :)

yurishkuro: I don't have objection to auto-formatting, I think we also use it in jaeger-ui.

But in that case we need to introduce it as a separate PR which reformats all existing source code without making any other changes. It's not a good idea to mix formatting changes with logic changes.

MarckK:
Interesting. As far as I can tell, opentracing-javascript is doing its [transpiling using entirely different tools](https://github.com/opentracing/opentracing-javascript/pull/68). That's very interesting, but very new to me.

package.json
MarckK: Agreed, completely. (It was a mistake.) I'll make a PR for prettier.

yurishkuro:
That's because it is transpiled from typescript. I didn't mean that, but the fact that after transpiling to JS the tests in Travis run with different versions of Node.

MarckK:
[Babel 7 no longer supports Node v0.10 and v0.12](https://github.com/babel/babel/issues/4315), and we've found this problem with [babel-preset-env](https://github.com/babel/babel-preset-env/issues/85) as well. This makes it more complicated for projects, such as ours, that have to support node versions that are unsupported by Babel.

MarckK:
It is worth debating if we should support node v0.10 and v0.13, considering they are EOL. Both [Babel](https://github.com/babel/babel/issues/4315) and [ESLint](https://github.com/eslint/eslint/issues/4483) have decided to drop support for node v0.10/0.12 and their discussions are illustrative, both of why they made that decision and also possible workarounds for us. It is also worth considering that running either of those in production is not recommended since any vulnerabilities will never be patched, and we shouldn't be 'enabling' bad practices. This might be an issue that should be brought up with opentracing-javascript?

yurishkuro:
0.10 is still actively used, including at Uber. These sorry of libraries cannot be on the cutting edge because we need to support many last long legacy apps. Similarly, OpenTracing Java still supports Java 1.6

yurishkuro:
My primary suggestion is separate language level used for tooling from the level of the final published library.
  * nvm use 6
  * run babel
  * nvm use 0.10
  * run unit tests

MarckK:
In the meantime, the easiest workaround to continuing to support node v0.10/0.12 is to not use babel-preset-env, but keep the status quo with babel-preset-2015. I had been trying to change the travis.yml so that when the node version running on Travis is not LTS, the LTS version of node would be used to set up the project and then it would be run on Travis on the intended (not LTS) node version.

MarckK:
OK, I can see how the opentracing-javascript solution is great; however, I [can't see where TRAVIS_NODE_VERSION is defined](https://github.com/opentracing/opentracing-javascript/search?q=TRAVIS_NODE_VERSION&type=Code&utf8=%E2%9C%93).

PR #200, opened 2018-01-12
Prettify .js, .json files

### LINKED ISSUE #198: Use Prettier
[What is Prettier?](https://prettier.io/docs/en/index.html)
[Why Prettier?](https://prettier.io/docs/en/why-prettier.html)
😍

MarckK:
Pin Prettier v1.10.2 as dev dependency
Close #198

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ test: build-node
 .PHONY: test-without-build
 test-without-build:
 	npm run flow
-	# npm run lint
+	npm run lint
 	npm run test-all
 	npm run check-license
 
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,6 @@ test: build-node
 .PHONY: test-without-build
 test-without-build:
 	npm run flow
-	#npm run lint
 	npm run test-all
 	npm run check-license
 
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ test: build-node
 .PHONY: test-without-build
 test-without-build:
 	npm run flow
-	npm run lint
+	#npm run lint
 	npm run test-all
 	npm run check-license
 
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ test: build-node
 .PHONY: test-without-build
 test-without-build:
 	npm run flow
-	npm run lint
+	# npm run lint
 	npm run test-all
 	npm run check-license
 
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,5 +1,5 @@
 {
-  "extends": "eslint-config-airbnb",
+  "extends": ["eslint-config-airbnb", "prettier"],
   "env": {
     "node": true
   },
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,6 +1,9 @@
 [include]
 ./node_modules/
 
+[ignore]
+.*/node_modules/eslint-plugin-jsx-a11y/.*
+
 [libs]
 ./src/_flow/
 ./test/_flow/
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,5 @@
+{
+  "printWidth": 110,
+  "singleQuote": true,
+  "trailingComma": "es5"
+}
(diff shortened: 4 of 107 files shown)


CHANGELOG.md
```diff
@@ -1,129 +1,112 @@
-Changes by Version
-==================
+# Changes by Version
```
black-adder: can we skip .md files for now? This file post pretty looks incorrectly formatted.

MarckK: Yes. Also, @black-adder, if you have any formatting preferences, please let me know. I've just used Prettier's default ones, as I don't have strong preferences for formatting, other than it be done automatically.

pavolloffay:
How did you format it? Is the format target run automatically during the build/commit? Is there a target to format it?

MarckK:
Thus far, I've used Prettier's default formatting. This is just a start to incorporate Prettier in the project. Next steps would be setting it up so that Prettier will run against all relevant files that are staged (pre-commit).

MarckK:
We could also use a script that runs Prettier as part of our CI checks, and that fails the build if relevant files differ from the Prettier output for that file.

pavolloffay:
If we are forcing some style I would prefer to actually fail the build if the code is not formatted properly. Rather that precommit I like more npm/make target which can be run anytime.

MarckK:
We can definitely have a format script, that is able to be run at anytime. The pre-commit hook is just another layer of reinforcement. Also, because this is opensource, the pre-commit hook makes it easier for people dropping in to make a PR (it removes the necessity of running the format script).

MarckK:
@Pavol, I'll make another commit excluding the .md files from being autoformatted. Prettier also has [options](https://prettier.io/docs/en/options.html) for formatting, so is configurable. If there are style choices you feel strongly about, I should add them in now as well. :)

yurishkuro:
@tiffon do you use Prettier in jaeger-ui?

tiffon:
@yurishkuro Yes, we use prettier in Jaeger UI; it's great. I recommend setting the line length to 110. 

FWIW, Jaeger UI uses a precommit hook and does not validate in CI.

yurishkuro:
I would also be in favor of pre-commit hook. Nothing is more annoying when the Travis build fails because `go fmt` wasn't run on a file.

@MarckK do you want to borrow the setup from jaeger-ui repo then?

MarckK:
Yes. Sounds good 👍 :)

MarckK:
Coverage is reduced as the number of lines has increased with the formatting done by Prettier (ratio of tests per line is reduced, so more untested lines).

pavolloffay:
How does it work?
* is code style checked in the build?
* is formatting automatic? If not how to format the code?

MarckK:
Hi @pavolloffay, `npm run format` will use prettier to format your local copy of the repo. 

Currently, this is using a pre-commit hook to prettify (to our specifications), so the formatting is done automatically before every commit. This makes it easier for others to contribute to the open source project (they don't have to remember to run format before committing).

There isn't a CI check for formatting, re Yuri's comment above: "I would also be in favor of pre-commit hook. Nothing is more annoying when the Travis build fails because go fmt wasn't run on a file."

Makefile
```diff
@@ -14,7 +14,7 @@ test: build-node
 .PHONY: test-without-build
 test-without-build:
 	npm run flow
-	npm run lint
+	#npm run lint
```
pavolloffay: this seems suspicious

MarckK: Yeah, I should take out the commented out line.  The reason `npm run lint` is being taken out is that we have reconfigured Travis to lint on LTS versions of node.

pavolloffay:
I have tested it locally with `v6.11.2` all worked well including precommit hook

@yurishkuro can we merge this?

black-adder:
@tiffon can you take a look at the new dependencies in package.json and the commit hook? This looks fine to land but want some expert eyes.

tiffon:
@black-adder, just seeing your comment, will look it over again tonight :+1:

tiffon:
@black-adder - This looks great.

Thanks for the awesome work @MarckK!

PR #204, opened 2018-01-22
Edit test script in Makefile

MarckK:


Makefile
```diff
@@ -9,7 +9,7 @@ publish: build-node
 
 .PHONY: test
 test: build-node
-	test-without-build
+	make test-without-build
```
yurishkuro: the correct way is 
```
test: build-node test-without-build
```

Issue #206, opened 2018-01-23
Spans being dropped by udp_sender


black-adder:
I wrote a load test that creates 100 traces at once. The udp_sender seemingly successfully sends the batches to agent with no errors. However,  I check the UI and I don't see all 100 traces. So I used tcpdump to see how many packets were arriving at the agent and indeed some packets were missing.

Given the fire and forget nature of udp, I guess this is valid? However, the dropped spans aren't observable since the dgram udp socket never returns an error even if the packet is dropped.

black-adder:
@tiffon you know anything we could do about this?

tiffon:
@black-adder How is this handled in the other clients? As long as the approach is feasible and practical in node, we should aim for the same approach, or similar to be consistent.

black-adder:
sigghhhhh, it was https://github.com/jaegertracing/jaeger-client-node/issues/124

Issue #157, opened 2017-10-08
Support callback in udpSender.flush


yurishkuro:
Right now there is no way to use Jaeger client in script-like apps (like the [OpenTracing tutorial for Node.js](https://github.com/yurishkuro/opentracing-tutorial/tree/master/nodejs)) because when the app exists the reporter may still have span buffered and they don't get flushed, even if the app calls `tracer.close()`.

The RemoteReporter and UDPSender are written in non-idiomatic style for JS/Node where UDPSender returns values from its append/flush method instead of using callbacks. It would be best to re-implement their interfaces to use callbacks for reporting the results. Then the reporter's own `close()` method make make sure that the UDP packet is actually flushed before returning.

MarckK:
A first attempt to wrap UDPSender flush method in a Promise has thrown up a number of issues: It compiles fine; however, I do get Flow errors, and some tests break, since I am returning a promise value and not a return value (for example: udp sender should flush with no spans returns false for error, and 0: AssertionError: expected undefined to equal false).

Additionally, when I compile the change in my local dev environment and use the edited tracer (for example, with the nodejs tutorial), it seems like `let numSpans: number = this._batch.spans.length;` can no longer access the `this._batch` property: I get the following error: UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): TypeError: Cannot read property '_batch' of undefined

Here's what I've done to the UDPSender flush method: 
```js
flush(): SenderResponse {
    return new Promise(function(resolve, reject) {
      let numSpans: number = this._batch.spans.length;
      if (numSpans == 0) {
        return resolve({ err: false, numSpans: 0 });
      }

      let bufferLen = this._totalSpanBytes + this._emitSpanBatchOverhead;
      let thriftBuffer = new Buffer(bufferLen);
      let writeResult = this._agentThrift.Agent.emitBatch.argumentsMessageRW.writeInto(
        this._convertBatchToThriftMessage(this._batch),
        thriftBuffer,
        0,
      );

      if (writeResult.err) {
        this._logger.error(`error writing Thrift object: ${writeResult.err}`);
        return reject({ err: true, numSpans: numSpans });
      }

      // Having the error callback here does not prevent uncaught exception from being thrown,
      // that's why in the constructor we also add a general on('error') handler.
      this._client.send(
        thriftBuffer,
        0,
        thriftBuffer.length,
        this._port,
        this._host,
        (err, sent) => {
          if (err) {
[...]

PR #207, opened 2018-01-23
Use callbacks in udpsender flush method

MarckK:

### DIFF
--- a/src/reporters/udp_sender.js
+++ b/src/reporters/udp_sender.js
@@ -155,6 +155,7 @@ export default class UDPSender {
 
     // Having the error callback here does not prevent uncaught exception from being thrown,
     // that's why in the constructor we also add a general on('error') handler.
+    this._reset();
     this._client.send(thriftBuffer, 0, thriftBuffer.length, this._port, this._host, (err, sent) => {
       if (err) {
         this._logger.error(
@@ -165,7 +166,6 @@ export default class UDPSender {
         callback({ err: err !== 0, numSpans: numSpans });
       }
     });
-    this._reset();
   }
 
   _convertBatchToThriftMessage() {
--- a/src/reporters/udp_sender.js
+++ b/src/reporters/udp_sender.js
@@ -143,7 +143,7 @@ export default class UDPSender {
 
         // Having the error callback here does not prevent uncaught exception from being thrown,
         // that's why in the constructor we also add a general on('error') handler.
-        this._client.send(thriftBuffer, 0, thriftBuffer.length, this._port, this._host, (err) => {
+        this._client.send(thriftBuffer, 0, thriftBuffer.length, this._port, this._host, (err, sent) => {
             if (err) {
                 this._logger.error(`error sending spans over UDP: ${err}, packet size: ${writeResult.offset}, bytes sent: ${sent}`);
             }
--- a/src/reporters/udp_sender.js
+++ b/src/reporters/udp_sender.js
@@ -118,9 +118,12 @@ export default class UDPSender {
         return flushResponse;
     }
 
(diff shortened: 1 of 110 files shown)


MarckK:
Relevant to issues #157 and #206.

src/reporters/udp_sender.js
```diff
@@ -132,19 +135,23 @@ export default class UDPSender {
...
+            }
+            return;
         }
 
         // Having the error callback here does not prevent uncaught exception from being thrown,
         // that's why in the constructor we also add a general on('error') handler.
-        this._client.send(thriftBuffer, 0, thriftBuffer.length, this._port, this._host, (err, sent) => {
+        this._client.send(thriftBuffer, 0, thriftBuffer.length, this._port, this._host, (err) => {
             if (err) {
                 this._logger.error(`error sending spans over UDP: ${err}, packet size: ${writeResult.offset}, bytes sent: ${sent}`);
```
yurishkuro: `sent` is used here

src/reporters/udp_sender.js
```diff
@@ -132,19 +135,23 @@ export default class UDPSender {
...
             }
+            if (callback) {
+              callback({err: err !== 0, numSpans: numSpans});
+            }
         });
         this._reset();
```
yurishkuro: this doesn't look right here.

yurishkuro:
do we need new tests to cover this extension?

src/reporters/udp_sender.js
MarckK: I had been thinking that the error that was passed into the optional callback (which is being passed in here) was a number: zero representing essentially no errors, other numbers denoting some type of error.  But this error that is being passed to the callback, is it in fact an object?
I've been looking for a definition of the error. [This test in node repo](https://github.com/nodejs/node/blob/98e54b0bd4854bdb3e2949d1b6b20d6777fb7cde/test/parallel/test-dgram-msgsize.js) implies that it is an error object.

yurishkuro: I was referring to the call to reset(). I think that logic is altogether broken since there are no guarantees that more spans won't be added to the buffer while the send is in progress.

Cc @black-adder

MarckK:
Question: for the moment we are failing on flow errors as UDPSender.flow no longer returns a SenderResponse object, but instead returns undefined. UDPSender.append currently expects flow method to return SenderResponse object so that it can then return that object -- I can make that happen, but only if I return the callback function in the flush method, which means flush method would then sometimes return out, sometimes return a callback.

MarckK:
Commit to move this._reset call in flush method of UDPSender closes #214.

src/reporters/udp_sender.js
```diff
@@ -91,37 +91,49 @@ export default class UDPSender {
...
-      return this.flush();
+      this.flush(callback);
+      return;
     }
-
-    let flushResponse: SenderResponse = this.flush();
-    this._batch.spans.push(span);
-    this._totalSpanBytes = spanSize;
-    return flushResponse;
+    this.flush(() => {
```
pavolloffay: missing callback?

MarckK: I'm using a callback here.

src/reporters/udp_sender.js
```diff
@@ -91,37 +91,49 @@ export default class UDPSender {
     this._maxSpanBytes = this._maxPacketSize - this._emitSpanBatchOverhead;
   }
 
-  append(span: any): SenderResponse {
+  append(span: any, callback: ?Function): void {
```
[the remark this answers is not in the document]
MarckK: Yes, because flush will only know if there is an error asynchronously.

PR #209, opened 2018-01-24
fix `this._batch` is undefined when start

nswbmw:
when start a Node.js application, wait a few seconds, it will throw:
```
       let numSpans: number = this._batch.spans.length;
                                           ^
TypeError: Cannot read property 'spans' of undefined
    at UDPSender.flush (/xxx/node_modules/jaeger-client/src/reporters/udp_sender.js:117:44)
```

### DIFF
--- a/src/reporters/udp_sender.js
+++ b/src/reporters/udp_sender.js
@@ -119,7 +119,7 @@ export default class UDPSender {
     }
 
     flush(): SenderResponse {
-        let numSpans: number = this._batch.spans.length;
+        let numSpans: number = this._batch ? this._batch.spans.length : 0;
         if (numSpans == 0) {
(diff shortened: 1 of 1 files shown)


black-adder:
Thanks for this, it's nice to have a sanity check there just in case. Do you mind posting how you initialized the jaeger client? because this should not be possible since the batch should've already been initialized by that time

nswbmw:
@black-adder https://github.com/nswbmw/koa-await-breakpoint-jaeger/blob/master/index.js#L18

black-adder:
can you try with the latest 3.8.0? This should've been fixed upstream before udp_sender gets called.

nswbmw:
@black-adder It works! Thanks

Issue #214, opened 2018-01-26
Possible race condition in UDPSender


yurishkuro:
I think the race condition is in this line

https://github.com/jaegertracing/jaeger-client-node/blob/f19b9e9b0af04aa96e0f81973ccafe8abff254cc/src/reporters/udp_sender.js#L145

The reset() should be called before `this._client.send`, because by then we have already taken the spans from the buffer and converted into thriftBuffer. The latter is local, so not an issue as I thought previously.

PR #216, opened 2018-02-07
Promisify RemoteReporter and UDPSender

MarckK:

### DIFF
--- a/src/_flow/reporter.js
+++ b/src/_flow/reporter.js
@@ -15,14 +15,14 @@ import Span from '../span.js';
 
 declare interface Reporter {
   report(span: Span): void;
-  close(callback: ?Function): void;
+  close(): Promise<void>;
   setProcess(serviceName: string, tags: Array<Tag>): void;
 }
 
 declare class Sender {
   append(span: Span): Promise<SenderResponse>;
   flush(): Promise<SenderResponse>;
-  close(): void;
+  close(): Promise<void>;
   setProcess(process: Process): void;
 }
 
--- a/src/_flow/reporter.js
+++ b/src/_flow/reporter.js
@@ -20,8 +20,8 @@ declare interface Reporter {
 }
 
 declare class Sender {
-  append(span: Span): SenderResponse;
-  flush(): SenderResponse;
+  append(span: Span): Promise<SenderResponse>;
+  flush(): Promise<SenderResponse>;
   close(): void;
   setProcess(process: Process): void;
 }
--- a/src/reporters/composite_reporter.js
+++ b/src/reporters/composite_reporter.js
@@ -11,6 +11,7 @@
 // or implied. See the License for the specific language governing permissions and limitations under
 // the License.
 
+import 'babel-polyfill';
 import Span from '../span.js';
 
 export default class CompositeReporter {
--- a/src/reporters/composite_reporter.js
+++ b/src/reporters/composite_reporter.js
@@ -30,23 +30,9 @@ export default class CompositeReporter {
     });
   }
 
-  compositeCallback(callback: ?Function): ?Function {
-    let count = 0;
-    return () => {
-      count++;
-      if (count >= this._reporters.length) {
-        if (callback) {
-          callback();
-        }
-      }
-    };
-  }
-
-  close(callback: ?Function): void {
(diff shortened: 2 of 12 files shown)


src/reporters/remote_reporter.js
```diff
@@ -45,29 +45,32 @@ export default class RemoteReporter {
...
   report(span: Span): void {
-    let response: SenderResponse = this._sender.append(ThriftUtils.spanToThrift(span));
-    if (response.err) {
-      this._logger.error('Failed to append spans in reporter.');
-      this._metrics.reporterDropped.increment(response.numSpans);
-    }
+    let responsePromise: Promise<SenderResponse> = this._sender.append(ThriftUtils.spanToThrift(span));
+
+    responsePromise.then(response => {
+      this._sender.append(ThriftUtils.spanToThrift(span));
```
yurishkuro: is this needed?

src/reporters/udp_sender.js
```diff
@@ -148,7 +148,7 @@ export default class UDPSender {
     });
     this._reset();
```
yurishkuro: I think reset needs to be called at L134, after the spans are converted to thrift buffer

src/reporters/udp_sender.js
```diff
@@ -148,7 +148,7 @@ export default class UDPSender {
...
 
-    return { err: false, numSpans: numSpans };
+    return Promise.resolve({ err: false, numSpans: numSpans });
```
yurishkuro: is there a Promise-based version of `this._client.send()`? If not, the promise needs to be created before calling `send()` and completed in the callback at L143

src/_flow/reporter.js
```diff
@@ -20,8 +20,8 @@ declare interface Reporter {
 }
 
 declare class Sender {
-  append(span: Span): SenderResponse;
-  flush(): SenderResponse;
+  append(span: Span): Promise<SenderResponse>;
+  flush(): Promise<SenderResponse>;
```
yurishkuro: how will the Promise class get translated to Node 0.10? Is there some backfill implementation that babel provides?

src/_flow/reporter.js
```diff
@@ -20,8 +20,8 @@ declare interface Reporter {
...
   close(): void;
```
yurishkuro: I think close() needs to be changed to use Promise as well

yurishkuro:
I think tracer.close() should also be changed to use promise from reporter.close()

Overall the change looks great.

src/_flow/reporter.js
MarckK: There is a babel polyfill.  I'm checking if Babel will implement it for us for Node 0.10 and 0.12. 🤞

MarckK: 😌 I have so much appreciation for Babel.

src/reporters/udp_sender.js
MarckK: Would this then close #214?

src/reporters/remote_reporter.js
```diff
@@ -45,29 +45,30 @@ export default class RemoteReporter {
   }
 
   report(span: Span): void {
-    let response: SenderResponse = this._sender.append(ThriftUtils.spanToThrift(span));
-    if (response.err) {
-      this._logger.error('Failed to append spans in reporter.');
-      this._metrics.reporterDropped.increment(response.numSpans);
-    }
+    this._sender.append(ThriftUtils.spanToThrift(span)).then(response => {
+      this._sender.append(ThriftUtils.spanToThrift(span));
```
yurishkuro: duplicate

src/reporters/composite_reporter.js
```diff
@@ -30,23 +30,9 @@ export default class CompositeReporter {
...
-  }
-
-  close(callback: ?Function): void {
-    let modifiedCallback: ?Function = this.compositeCallback(callback);
-    this._reporters.forEach(r => {
-      r.close(modifiedCallback);
-    });
+  close(): Promise<void> {
+    let promises = this._reporters.map(r => r.close());
+    return Promise.all(promises).then(() => {});
```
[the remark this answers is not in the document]
MarckK: Unfortunately, generates flow errors. Might be better to take out the option to pass callbacks into RemoteReporter close method.

Issue #139, opened 2017-07-20
Make tracing headers configurable


frankgreco:
Currently, `TextMapCodec ` allows dynamic configuration of `contextKey` and `baggagePrefix`. However, there seems to be no way for a user to configure these values as `TextMapCodec` is [instantiated](https://github.com/uber/jaeger-client-node/blob/6d7544087e148072fcb8bf6754772a4781d90929/src/tracer.js#L76-L79) as follows:

```js
let textCodec = new TextMapCodec({
    urlEncoding: false,
    metrics: this._metrics
});
```

This is also an open issue with the go client [here](https://github.com/uber/jaeger-client-go/issues/109).

astub:
I'm going to open a PR for this soon.

yurishkuro:
Fixed by #217

PR #217, opened 2018-02-07
Made tracing headers configurable. (#139)

astub:
Signed-off-by: WARREN, JACOB <jw748c@us.att.com>


test/tracer.js
```diff
@@ -37,6 +37,20 @@ describe('tracer should', () => {
...
+      contextKey: "test-trace-id"
+    });
+
+    let headers = {
+      'test-trace-id': "a:b:c:d"
+    };
+
+    let mycontext = mytracer.extract(opentracing.FORMAT_HTTP_HEADERS, headers);
+
+    assert.equal(mycontext.toString(), 'a:b:c:d');
```
yurishkuro: you may also want to test `inject()`

astub:
@yurishkuro 

Do you need changes to be a single commit? 

Is the build failure expected? I looks like its failing after passing creds into docker.

tiffon:
I think the problem is your changes were made on a branch named `master` so the travis build is trying to upload the docker image. Renaming your branch should do the trick.

PR #222, opened 2018-02-18
Exposing formats, to make it easy to get them to inject span_context

eldadcohen1:
Today we can not get formats from the library.
I expose the formats to make it easy import formats const string.

### DIFF
--- a/entrypoint.js
+++ b/entrypoint.js
@@ -17,6 +17,7 @@ module.exports = {
   SpanContext: require('./dist/src/span_context.js').default,
   Span: require('./dist/src/span.js').default,
   Tracer: require('./dist/src/tracer.js').default,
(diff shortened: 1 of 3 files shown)


tiffon:
The formats should be available as `opentracing` is already an API export:

https://github.com/jaegertracing/jaeger-client-node/blob/a78629002c39338d56e579095c9f3a35ed603f81/entrypoint.js#L37

Seems like that should suffice...?

eldadcohen1:
@tiffon Is not because you should search them when you coding from opentracing library.
What I've done is to make it easier to use.

yurishkuro:
@eldadcohen1 not sure I follow. When someone writes instrumentation, they should only be concerned with the opentracing API, including all the constants from opentracing. Jaeger module only needs to be imported by the main().

PR #223, opened 2018-02-18
Expose constants for easy usage

eldadcohen1:
I Exposed constants for easy usage.
For tracer configuration you should supply sampler type, it's a defined constant in the library.

### DIFF
--- a/entrypoint.js
+++ b/entrypoint.js
@@ -28,6 +28,7 @@ module.exports = {
   LoggingReporter: require('./dist/src/reporters/logging_reporter.js').default,
   NoopReporter: require('./dist/src/reporters/noop_reporter.js').default,
   RemoteReporter: require('./dist/src/reporters/remote_reporter.js').default,
+  Constants: require('./dist/src/constants'),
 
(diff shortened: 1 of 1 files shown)


yurishkuro:
I have a small concern with this change because it may expose some constants that ideally should be kept private to the module.

tiffon:
I agree with @yurishkuro, there are a lot of inner workings in the constants exports. 

Maybe something like the following would suffice?

```js
export const samplerTypes = {
  CONST: SAMPLER_TYPE_CONST,
  PROBABILISTIC: SAMPLER_TYPE_PROBABILISTIC
  RATE_LIMITING: SAMPLER_TYPE_RATE_LIMITING,
  LOWER_BOUND: SAMPLER_TYPE_LOWER_BOUND,
  REMOTE: SAMPLER_TYPE_REMOTE,
};

// ...

  samplerTypes: require('...').samplerTypes,
```

eldadcohen1:
OK I agree, I'll fix it.

yurishkuro:
closing due to inactivity.

PR #234, opened 2018-02-26
Re-add husky as dev dependency

### LINKED ISSUE #232: husky should not be an optional dependency
`optionalDependencies` are being installed for consumers of the library as well, resulting in having husky installed in my project and checking my commit messages now. I think it should better be a `devDependency` as well.

tiffon:
Fix #232.

This approach _might_ work. The CI script uninstalls husky before the `npm install`.


MarckK:
Unfortunately, [Yarn targets node > 4](https://github.com/yarnpkg/yarn/blob/aee005aa16a929a69730235e4f178f5a7bc5f81b/package.json#L95)

tiffon:
>Unfortunately, Yarn targets node > 4

@MarckK Great point, thanks :+1:

tiffon:
Closing in favor of #235.

PR #240, opened 2018-02-28
Run tests against dist/* in CI

### LINKED ISSUE #239: Tests in CI are not run against dist/
Tests run in CI do not run against `dist/...`. Running them against (or in) `dist/...` is preferable as those are the files published to npm.

tiffon:
Fix #239.

Only runs the non-transpiled tests for node 6.

Moves the entry point to `src/index` (`dist/src/index` in the transpiled code).

cc @MarckK


.travis.yml
```diff
@@ -47,12 +47,12 @@ before_install:
 - nvm use $TRAVIS_NODE_VERSION
 
 before_script:
+- npm uninstall -D husky lint-staged
```
yurishkuro: makes sense removing them upfront, but in practice Travis runs `npm install` as part of the `install` step (which we don't override), so it would make sense to move this command even higher, as the last one in `before_install`

package.json
```diff
@@ -8,7 +8,7 @@
   "license": "Apache-2.0",
   "keywords": [],
   "author": "Onwukike Ibe <oibe@uber.com>",
-  "main": "./entrypoint.js",
+  "main": "./dist/src/index.js",
```
yurishkuro: 👍

.travis.yml
tiffon: Hm, it doesn't seem to be working when I move this to the `before_install`:

The `uninstall` command runs: https://travis-ci.org/jaegertracing/jaeger-client-node/jobs/347115286#L628

But, husky is still installed: https://travis-ci.org/jaegertracing/jaeger-client-node/jobs/347115286#L1186

I'm going to move it back to the first step in the `before_script` so the CI builds pass for the older versions of node and this PR can move forward.

PR #245, opened 2018-03-27
Fix baggage formatting

### LINKED ISSUE #244: trimming space in parseCommaSeparatedMap
The comments here https://github.com/jaegertracing/jaeger-client-node/blob/master/src/propagators/baggage.js#L16-L19 
suggest that if I enter "key3 = value3 ", then it should return  {"key3": "value3"}
But it doesn't

related Issue in GO.  https://github.com/jaegertracing/jaeger-client-go/issues/271

```
function parseCommaSeparatedBaggage(baggage, values) {
    values.split(',').forEach((keyVal) => {
        let splitKeyVal = keyVal.trim().split('=');
        if (splitKeyVal.length == 2) {
            baggage[splitKeyVal[0]] = splitKeyVal[1];
        }
    });
}
t = {};
parseCommaSeparatedBaggage(t, "  key3  =  value3 ");
console.log(t);  // returns {"key3  ": "  value3"}
```

since trimming is not done after the = separation.
I'm not sure if this is intentional or a bug. If it's intentional, then the comments should be updated.

MarckK:
Fix #244

### DIFF
--- a/src/propagators/baggage.js
+++ b/src/propagators/baggage.js
@@ -24,7 +24,7 @@
  */
 export function parseCommaSeparatedBaggage(baggage: any, values: string): void {
   values.split(',').forEach(keyVal => {
-    let splitKeyVal: Array<string> = keyVal.trim().split('=');
+    let splitKeyVal: Array<string> = keyVal.split('=').map(str => str.trim());
     if (splitKeyVal.length == 2) {
       baggage[splitKeyVal[0]] = splitKeyVal[1];
     }


yurishkuro:
we should have a test for tolerating whitespace in baggage

MarckK:
Oh no, I lost track of this PR and yesterday killed my local branch of this, plus updated my local. Hmm, I'm going to end this PR and send up another. Sorry about that.

PR #246, opened 2018-04-02
Add throttler

black-adder:
Throttler is a rate limiter that frequently fetches credits for each operation and rate limits said operation if it has insufficient credits. This PR only adds the throttler, the actual rate limiting will be done in a subsequent PR. 


src/throttler/remote_thottler.js
```diff
@@ -0,0 +1,188 @@
...
+
+  setProcess(process: Process): void {
+    this._uuid = process.uuid;
+  }
+
+  isAllowed(operation: string): boolean {
+    if (!this._credits.has(operation)) {
+      this._credits.set(operation, 0);
+      // If seen for the first time, async fetch credits
+      this._refreshCredits();
```
black-adder: In golang, when we see the operation for the first time, we have the option of allowing service owners to synchronously fetch credits for the first call; this is useful for cases like yab. In node however, I spoke with @tiffon and he said it's not advisable (TBF, it's not advisable in golang either but I think necessary).

I think this is an ok compromise.

src/_flow/jaeger-thrift.js
```diff
@@ -24,6 +24,9 @@ declare type LogData = {
 declare type Process = {
   serviceName: string,
   tags: Array<Tag>,
+  // N.B. uuid uniquely identifies this instance of the client. It is only used by the client and is not
+  // propagated as part of the real jaeger thrift process.
+  uuid?: string,
```
black-adder: I made this optional to not break all the things. Should not be a breaking change.

package.json
```diff
@@ -60,18 +60,19 @@
   "scripts": {
     "check-license": "./scripts/check-license.sh",
     "check-ls": "npm ls --loglevel=http --parseable 1>/dev/null && echo '# npm is in a good state'",
-    "cover": "babel-node ./node_modules/.bin/babel-istanbul cover ./node_modules/.bin/_mocha -- test/ test/samplers/ test/baggage/",
+    "cover": "babel-node ./node_modules/.bin/babel-istanbul cover ./node_modules/.bin/_mocha -- test/ test/samplers/ test/baggage/ test/throttler/",
     "coveralls": "npm run cover && cat ./coverage/lcov.info | coveralls",
     "flow": "flow",
-    "format": "prettier --write '**/*.{js,json,md}'",
+    "format": "prettier --write '**/*.{js,json,md}' '!src/jaeger-idl/**'",
```
black-adder: added a rule to skip prettifying a git submodule

src/throttler/remote_thottler.js
```diff
@@ -0,0 +1,185 @@
...
+    this._refreshIntervalHandle = setInterval(this._refreshCredits.bind(this), this._refreshIntervalMs);
+  }
+
+  setProcess(process: Process): void {
+    this._uuid = process.uuid || '';
+  }
+
+  isAllowed(operation: string): boolean {
+    if (operation in this._credits) {
+      // TODO if credits is 0, is this false? Do I need to do an explicit check on undefined?
```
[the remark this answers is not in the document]
tiffon: Or is that the point of `initialDelayMs`? Seems like it won't prevent the above scenario, though.

src/throttler/remote_thottler.js
```diff
@@ -0,0 +1,185 @@
...
+      return this._isAllowed(operation);
+    }
+    this._credits[operation] = 0;
+    // If seen for the first time, async fetch credits
+    this._refreshCredits();
+    return false;
+  }
+
+  _isAllowed(operation: string): boolean {
+    // N.B. The -1 assignment is necessary, otherwise we need to type credits as ?number which
```
tiffon: What `-1` assignment?

src/throttler/remote_thottler.js
```diff
@@ -0,0 +1,185 @@
...
+   */
+  constructor(serviceName: string, options: any = {}) {
+    this._serviceName = serviceName;
+    this._logger = options.logger || new NullLogger();
+    this._metrics = options.metrics || new Metrics(new NoopMetricFactory());
+    this._refreshIntervalMs = options.refreshIntervalMs || DEFAULT_REFRESH_INTERVAL_MS;
+    this._host = options.host || DEFAULT_THROTTLER_HOST;
+    this._port = options.port || DEFAULT_THROTTLER_PORT;
+
+    this._credits = Object.create(null);
```
tiffon: What's the advantage of `Object.create(null)` vs `{}` in this case?

black-adder:
addressed some comments, looking into debouncing and other comments

test/throttler/remote_throttler.js
```diff
@@ -0,0 +1,163 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import { assert } from 'chai';
+import Metrics from '../../src/metrics/metrics.js';
+import MockLogger from '../lib/mock_logger';
+import ConfigServer from '../lib/config_server';
+import LocalMetricFactory from '../lib/metrics/local/metric_factory.js';
+import LocalBackend from '../lib/metrics/local/backend.js';
+import RemoteThrottler from '../../src/throttler/remote_thottler';
```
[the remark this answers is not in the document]
black-adder: that's a mighty sharp eye you have there

src/throttler/remote_thottler.js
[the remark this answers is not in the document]
black-adder: the credits can't ever be set something that's not null so the current check will suffice

src/throttler/remote_thottler.js
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  _fetchCredits(operations: any) {
+    let serviceName: string = encodeURIComponent(this._serviceName);
+    let uuid: string = encodeURIComponent(this._uuid);
+    let url: string = `/credits?service=${serviceName}&uuid=${uuid}`;
+
+    operations.forEach(operation => {
+      url = url + `&operation=${encodeURIComponent(operation)}`;
+    });
```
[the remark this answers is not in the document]
black-adder: This led me astray: https://stackoverflow.com/questions/13859543/fast-way-to-concatenate-strings-in-nodejs-javascript

Is that link incorrect? I changed over to using join

src/throttler/remote_thottler.js
black-adder: sorry old comment

src/throttler/remote_thottler.js
black-adder: I copied over legacy code that did this. Since we're using the keys of the object, won't Object.keys({}) return things like _proto which is something we don't want right? I'll check

src/throttler/remote_throttler.js
```diff
@@ -0,0 +1,180 @@
...
+    return true;
+  }
+
+  _refreshCredits() {
+    if (!this._uuid) {
+      this._logger.error(`UUID must be set to fetch credits`);
+      return;
+    }
+    const keys = Object.keys(this._credits);
+    if (keys.length == 0) {
```
tiffon: I recommend using `===` instead of `==` for all cases except comparisons with `null` or `undefined`.

Ref: [Loose equality using ==](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness#Loose_equality_using)

src/throttler/remote_throttler.js
```diff
@@ -0,0 +1,180 @@
...
+  _fetchCredits(operations: any) {
+    let serviceName: string = encodeURIComponent(this._serviceName);
+    let uuid: string = encodeURIComponent(this._uuid);
+    let ops = operations.map(encodeURIComponent).join('&operation=');
+    let url = `/credits?service=${serviceName}&uuid=${uuid}&operation=${ops}`;
+
+    let success: Function = body => {
+      this._parseCreditResponse(body);
+    };
+    let error: Function = err => {
```
tiffon: I'd recommend any `let` that doesn't change be switched to a `const`.

src/throttler/remote_thottler.js
tiffon: It should be fine:

>The object of which the enumerable's own properties are to be returned.

[MDN Object.keys](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys)

src/throttler/remote_thottler.js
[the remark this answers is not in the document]
tiffon: Seems to vary based on the strings, themselves. But, they're all fairly close in most scenarios.

test/throttler/remote_throttler.js
```diff
@@ -70,13 +70,20 @@ describe('RemoteThrottler should', () => {
...
   it('log an error if _refreshCredits is called prior to UUID being set', () => {
     throttler._refreshCredits();
     assert.equal(logger._errorMsgs.length, 1);
   });
 
+  it('not fetch credits if uuid is invalid', () => {
+    throttler = new RemoteThrottler(serviceName);
+    throttler.setProcess({ uuid: null });
+    throttler._refreshCredits();
+  });
```
tiffon: Should there be an assert, here?

src/metrics/metrics.js
```diff
@@ -120,5 +122,13 @@ export default class Metrics {
     });
 
     this.baggageTruncate = this._factory.createCounter('baggage-trucate');
+
+    this.throttlerUpdateSuccess = this._factory.createCounter('throttler-update', {
```
tiffon: I don't know the params for this factory, but should this string be something like `throttler-update-success` and the next one `throttler-update-failure`?

src/samplers/remote_sampler.js
```diff
@@ -88,36 +88,19 @@ export default class RemoteControlledSampler {
...
-          res.on('end', () => {
-            this._parseSamplingServerResponse(body);
-          });
-        }
-      )
-      .on('error', err => {
-        this._logger.error(`Error in fetching sampling strategy: ${err}.`);
-        this._metrics.samplerQueryFailure.increment(1);
-      });
+    let success: Function = body => {
```
tiffon: I recommend using `const` on values that don't change.

src/throttler/default_throttler.js
```diff
@@ -0,0 +1,27 @@
...
+// or implied. See the License for the specific language governing permissions and limitations under
+// the License.
+
+/**
+ * DefaultThrottler either throttles everything or nothing.
+ */
+export default class DefaultThrottler {
+  _throttleAll: boolean;
+
+  constructor(throttleAll: ?boolean) {
```
tiffon: I think this should be `throttleAll?: boolean`

src/throttler/remote_throttler.js
```diff
@@ -0,0 +1,179 @@
...
+
+  setProcess(process: Process): void {
+    this._uuid = process.uuid || '';
+  }
+
+  isAllowed(operation: string): boolean {
+    if (operation in this._credits) {
+      return this._isAllowed(operation);
+    }
+    // Credits for the operation will be asynchronously fetched
```
tiffon: Just curious, is relying on the interval consistent with the Go and Java throttlers?

src/throttler/remote_throttler.js
```diff
@@ -0,0 +1,179 @@
...
+      this._logger.error(`Error in updating credits: ${error}.`);
+      this._metrics.throttlerUpdateFailure.increment(1);
+      return;
+    }
+    if (this._onCreditsUpdate) {
+      this._onCreditsUpdate(this);
+    }
+  }
+
+  close(callback: ?Function): void {
```
tiffon: I believe this should be `callback?: Function`, or `callback?: () => void`

test/lib/config_server.js
```diff
@@ -13,38 +13,57 @@
...
   }
 
-  clearStrategies(): void {
+  addCredits(serviceName: string, response: Array<CreditResponse>): void {
+    this._credits[serviceName] = response;
+  }
+
+  clearConfigs(): void {
     this._strategies = Object.create(null);
+    this._credits = Object.create(null);
```
tiffon: I think all of these `Object.create(null)` should just be `{}`...

test/throttler/remote_throttler.js
```diff
@@ -0,0 +1,171 @@
...
+    throttler.setProcess({ uuid: uuid });
+    throttler._refreshCredits();
+    assert.equal(Object.keys(throttler._credits).length, 0);
+  });
+
+  it('refresh credits after _afterInitialDelay is called', done => {
+    throttler.setProcess({ uuid: uuid });
+    throttler._credits[operation] = 0;
+    server.addCredits(serviceName, [{ operation: operation, credits: 5 }]);
+    throttler._onCreditsUpdate = _throttler => {
```
tiffon: Looks like you can avoid overwriting the private(-ish) property by using the `options.onCreditUpdate`. This would avoid relying on the private property and breaking the public API would break the test.

```
let creditUpdatedHook;

beforeEach(() => {
  creditUpdatedHook = sinon.spy();
  throttler = new RemoteThrottler(serviceName, {
    // ...
    onCreditUpdate: (...args) => creditUpdatedHook(...args),
  });
});

it('....', () => {
  assertTrue(creditUpdatedHook.called);
});

it('...', done => {
  creditUpdatedHook = () => {
    // ...
    done();
  }
});
```

test/throttler/remote_throttler.js
```diff
@@ -0,0 +1,171 @@
...
+      assert.equal(_throttler._credits[operation], 4);
+      assert.isOk(_throttler.isAllowed(other_operation));
+      assert.equal(_throttler._credits[other_operation], 2);
+      assert.equal(LocalBackend.counterValue(metrics.throttlerUpdateSuccess), 1);
+      done();
+    };
+    throttler._refreshCredits();
+  });
+
+  it('emit failure metric on failing to query for credits', done => {
```
tiffon: For the failure scenarios, you might consider also ensuring the credit update callback does not fire.

src/throttler/remote_throttler.js
black-adder: yup

test/throttler/remote_throttler.js
black-adder: stupid setTimeout issue on travis strikes again. I think asserting that the logs were added are good enough for these tests.

test/lib/config_server.js
tiffon: :+1:

src/throttler/remote_throttler.js
tiffon: :+1:

src/throttler/remote_throttler.js
```diff
@@ -0,0 +1,179 @@
...
+      return;
+    }
+    if (this._onCreditsUpdate) {
+      this._onCreditsUpdate(this);
+    }
+  }
+
+  close(callback?: Function): void {
+    clearTimeout(this._initialDelayTimeoutHandle);
+    clearInterval(this._refreshIntervalHandle);
```
tiffon: There is an edge case of a credits fetch finishing after close is called.

PR #248, opened 2018-04-12
Use throttler for rate limiting

black-adder:


src/span.js
```diff
@@ -205,8 +205,10 @@ export default class Span {
    **/
   addTags(keyValuePairs: any): Span {
     if (opentracing.Tags.SAMPLING_PRIORITY in keyValuePairs) {
-      this._setSamplingPriority(keyValuePairs[opentracing.Tags.SAMPLING_PRIORITY]);
-      delete keyValuePairs[opentracing.Tags.SAMPLING_PRIORITY];
+      // Delete the SAMPLING_PRIORITY tag if it was not set
```
tiffon: If the span is already marked as debug (for instance from [`.isDebugIDContainerOnly`](https://github.com/jaegertracing/jaeger-client-node/blob/master/src/tracer.js#L217)) then the tag should be ignored so as to not hit the credits twice. Same for `. setTag(...)`, below.

src/tracer.js
```diff
@@ -96,7 +100,18 @@ export default class Tracer {
     this.registerInjector(opentracing.FORMAT_BINARY, binaryCodec);
     this.registerExtractor(opentracing.FORMAT_BINARY, binaryCodec);
 
-    this._reporter.setProcess(this._serviceName, Utils.convertObjectToTags(this._tags));
+    const uuid = Utils.getRandom64().toString('hex');
```
tiffon: `uuid` is a misnomer here and in remote throttler as the value is not actually a uuid.

src/tracer.js
```diff
@@ -96,7 +100,18 @@ export default class Tracer {
...
+    this._tags[constants.TRACER_CLIENT_ID_TAG_KEY] = uuid;
+    this._process = {
+      serviceName: serviceName,
+      tags: Utils.convertObjectToTags(this._tags),
+      uuid: uuid,
+    };
+    if (typeof this._debugThrottler.setProcess === 'function') {
```
tiffon: The default throttler can have `.setProcess(...)` as a noop. This would keep the interface consistent. Same for `.close(...)`.

test/init_tracer.js
```diff
@@ -21,6 +20,24 @@ import RemoteSampler from '../src/samplers/remote_sampler';
 import RateLimitingSampler from '../src/samplers/ratelimiting_sampler';
 import { initTracer } from '../src/index.js';
 import opentracing from 'opentracing';
+import RemoteThrottler from '../src/throttler/remote_throttler';
+import DefaultThrottler from '../src/throttler/default_throttler';
+
+const logger = {
```
tiffon: `const` !!! :+1:

test/init_tracer.js
```diff
@@ -167,4 +170,33 @@ describe('initTracer', () => {
...
+    expect(tracer._debugThrottler).to.be.an.instanceof(DefaultThrottler);
+  });
+
+  it('should use throttler passed in via options', () => {
+    let config = {
+      serviceName: 'test-service',
+    };
+    const throttler = new RemoteThrottler();
+    const tracer = initTracer(config, { throttler: throttler });
+    expect(tracer._debugThrottler).to.be.an.instanceof(RemoteThrottler);
```
tiffon: I believe this assertion should verify the instance is used:

```js
expect(tracer._debugThrottler).to.be(RemoteThrottler);
```

test/span.js
```diff
@@ -170,6 +170,35 @@ describe('span should', () => {
...
+      span._tags.length,
+      'The sampling.priority tag should not be set if throttled'
+    );
+
+    span = new Span(tracer, 'op-name', spanContext, tracer.now());
+
+    prevTagLength = span._tags.length;
+    const tags = {};
+    tags[opentracing.Tags.SAMPLING_PRIORITY] = 1;
+    span.addTags(tags);
```
tiffon: Seems like this would warrant being a separate test-case?

test/text_map_codec.js
```diff
@@ -68,4 +69,24 @@ describe('Text Map Codec should', () => {
     // metrics
     assert.isOk(LocalBackend.counterEquals(metrics.tracesStartedSampled, 1));
   });
+
+  it('not set debug flag when debug-id-header is received but operation is throttled', () => {
```
tiffon: Does this mean that headers indicating the trace is a debug trace will be ignored and subsequent spans will not be marked with the debug flag?

black-adder: Yes since it's throttled.

src/span.js
```diff
@@ -287,13 +286,18 @@ export default class Span {
...
+        // If the span is already in debug, no need to set it again
+        return false;
+      }
+      if (this._tracer._isDebugAllowed(this._operationName)) {
+        this._spanContext.flags = this._spanContext.flags | constants.SAMPLED_MASK | constants.DEBUG_MASK;
+        return true;
+      }
+      return false;
     }
+    this._spanContext.flags = this._spanContext.flags & ~constants.SAMPLED_MASK;
```
tiffon: Here you're updating the flag but returning false...?

src/throttler/default_throttler.js
```diff
@@ -24,4 +24,12 @@ export default class DefaultThrottler {
   isAllowed(operation: string): boolean {
     return !this._throttleAll;
   }
+
+  setProcess(process: Process): void {
+    // NOP
+  }
+
+  close(callback?: Function): void {
+    // NOP
```
tiffon: This should probably call the callback?

src/configuration.js
```diff
@@ -175,6 +207,7 @@ export default class Configuration {
       metrics: options.metrics,
       logger: options.logger,
       tags: options.tags,
+      debugThrottler: throttler,
```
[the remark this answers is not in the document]
tiffon: What's the rationale for naming the throttler throttler instead of debugThrottler?

src/tracer.js
```diff
@@ -96,7 +100,16 @@ export default class Tracer {
     this.registerInjector(opentracing.FORMAT_BINARY, binaryCodec);
     this.registerExtractor(opentracing.FORMAT_BINARY, binaryCodec);
 
-    this._reporter.setProcess(this._serviceName, Utils.convertObjectToTags(this._tags));
+    const uuid = Utils.getRandom64().toString('hex');
+    this._tags[constants.TRACER_CLIENT_ID_TAG_KEY] = uuid;
+    this._process = {
+      serviceName: serviceName,
+      tags: Utils.convertObjectToTags(this._tags),
+      uuid: uuid,
```
tiffon: Why is this not named processId as it's not a UUID its just a random number?

test/text_map_codec.js
tiffon: Sounds kind of ruthless... is this consistent with the other clients?

black-adder: this is the first place we're implementing it. How is being throttled via debug header vs throttled vs sampling.priority any different? We should be defending against both.

src/span.js
black-adder: I made this change on purpose but now that you mention it, semantically it doesn't make sense. Even though this returning true or false has no impact on the span (since we're disabling sampling so the flag being set or not will never be surfaced to the user). I'll revert

src/tracer.js
black-adder: This is somewhat unfortunate. I could pull in another dependency (npm uuid) and actually generate a uuid but given the collision can only occur on one host, I didn't see the need to generate a real uuid.

bah, ill try to look for a very light weight uuid generator maybe this (https://gist.github.com/jed/982883) and inline it. I rather not pull in a new dependency.

src/configuration.js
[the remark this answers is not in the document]
black-adder: at the moment we're only using it as a debug throttler but we can always use the throttler to throttle normal sampling decisions too (perhaps if the user is using const sampling and we want to throttle it). So I prefer to leave the name as "throttler" because it'll be a pain in the butt to call this debugThrottler and then rename it in the future.

tiffon:
@black-adder You're missing the DCO sign-off for something.

PR #249, opened 2018-04-18
Use different "good commit message" reference and inline it 

eundoosong:
refs https://github.com/jaegertracing/jaeger/issues/771

Signed-off-by: Eundoo Song <eundoo.song@gmail.com>


eundoosong:
@yurishkuro  Oops, I'll follow this from the next patch. thanks.

PR #257, opened 2018-05-08
Pass logger/metrics option to remote sampler,reporter

eundoosong:
When remote reporter and sampler are initialized via _getReporter and _getSampler, logger and metrics options are not passed into each constructor. 

Signed-off-by: eundoo-song <eundoo.song@samsung.com>

### DIFF
--- a/src/configuration.js
+++ b/src/configuration.js
@@ -53,7 +53,7 @@ let jaegerSchema = {
 };
 
 export default class Configuration {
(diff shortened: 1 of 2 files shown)


eundoosong:
Oh, the error disappeared! Thanks.

black-adder:
looks like it was a flaky test, I added a ticket for it.

PR #272, opened 2018-07-04
Upgrade flow to v0.75

TLadd:
Signed-off-by: Thomas Ladd <thomas.rye.ladd@gmail.com>

Resolves https://github.com/jaegertracing/jaeger-client-node/issues/228

Upgrades to the latest version of flow

Upgrade flow and make slight modifications to fix a handful of new type errors.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -9,7 +9,6 @@
 ./test/_flow/
 
 [options]
-unsafe.enable_getters_and_setters=true
 suppress_comment= \\(.\\|\n\\)*\\$FlowIgnore
 esproposal.class_static_fields=enable
 esproposal.class_instance_fields=enable
--- a/src/reporters/udp_sender.js
+++ b/src/reporters/udp_sender.js
@@ -59,7 +59,7 @@ export default class UDPSender {
 
   _calcBatchSize(batch: Batch) {
     return this._agentThrift.Agent.emitBatch.argumentsMessageRW.byteLength(
-      this._convertBatchToThriftMessage(this._batch)
+      this._convertBatchToThriftMessage()
     ).length;
   }
 
@@ -144,7 +144,7 @@ export default class UDPSender {
     const bufferLen = this._totalSpanBytes + this._emitSpanBatchOverhead;
     const thriftBuffer = new Buffer(bufferLen);
     const writeResult = this._agentThrift.Agent.emitBatch.argumentsMessageRW.writeInto(
-      this._convertBatchToThriftMessage(this._batch),
+      this._convertBatchToThriftMessage(),
       thriftBuffer,
       0
     );
--- a/src/samplers/per_operation_sampler.js
+++ b/src/samplers/per_operation_sampler.js
@@ -16,7 +16,7 @@ import * as constants from '../constants.js';
 import ProbabilisticSampler from './probabilistic_sampler.js';
 import GuaranteedThroughputSampler from './guaranteed_throughput_sampler.js';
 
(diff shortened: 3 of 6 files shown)


.flowconfig
```diff
@@ -9,7 +9,6 @@
 ./test/_flow/
 
 [options]
-unsafe.enable_getters_and_setters=true
```
TLadd: Option was removed and is enabled by default.

src/reporters/udp_sender.js
```diff
@@ -59,7 +59,7 @@ export default class UDPSender {
 
   _calcBatchSize(batch: Batch) {
     return this._agentThrift.Agent.emitBatch.argumentsMessageRW.byteLength(
-      this._convertBatchToThriftMessage(this._batch)
```
TLadd: Method does not actually take an argument. It uses this._batch itself.

src/samplers/per_operation_sampler.js
```diff
@@ -16,7 +16,7 @@ import * as constants from '../constants.js';
 import ProbabilisticSampler from './probabilistic_sampler.js';
 import GuaranteedThroughputSampler from './guaranteed_throughput_sampler.js';
 
-type SamplersByOperation = { [key: string]: GuaranteedThroughputSampler };
```
TLadd: https://github.com/facebook/flow/issues/4967

src/span_context.js
```diff
@@ -134,7 +134,7 @@ export default class SpanContext {
     this._spanIdStr = null;
   }
 
-  set parentId(parentId: Buffer): void {
```
TLadd: Necessary because of https://github.com/jaegertracing/jaeger-client-node/blob/master/src/tracer.js#L226

src/span_context.js
```diff
@@ -134,7 +134,7 @@ export default class SpanContext {
...
+  set parentId(parentId: Buffer | null): void {
```
yurishkuro: what does `| null` mean? It should not be legal to pass null to this method

TLadd: It is necessary because parentId is set to null here: https://github.com/jaegertracing/jaeger-client-node/blob/master/src/tracer.js#L226

TLadd: When doing `ctx.parentId = null`, this set function is called. [From MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/set): `The set syntax binds an object property to a function to be called when there is an attempt to set that property.` 

Thus, as it is written currently, this setter is called with null. Flow understands this and triggers an error. Here is a small example: https://flow.org/try/#0PTAEAEDMBsHsHcBQiDG0CGBnToAqBTTAFwGENtQBvRUUTfI0AB3QCd8A7IgSQBNQAFAEteALjpFWQjgHMAlFRq1QRABZDMAOhbsufUAF5QIgNy0lAX0RXUsDsRWFGRjvnh4nZLJgFzkRJ202Th5+IwAiHRC+cMQA4iDdUMNQDgBXaGggA

tiffon: @TLadd Thanks for upgrading flow! Awesome!

Just curious, why not mark the field as `?Buffer` ?

TLadd: It's not particularly important in this case either way. ?Buffer is equivalent to `Buffer | null | void`. So by specifying `parentId: Buffer | null`, it is not legal to do `ctx.parentId = undefined` whereas that would be fine with ?Buffer. Since we only had the one case of setting parentId to null, I figured not allowing undefined was reasonable.

PR #274, opened 2018-07-05
Pass object to prom-client metric to fix warning

eundoosong:
Signed-off-by: Eundoo Song <eundoo.song@gmail.com>

- This removes prom-client warning called when a metric is constructed.

- Passing a non-object to metrics constructor is deprecated. So it shows below warning when a metric is constructed.
```
(node:49239) DeprecationWarning: prom-client - Passing a non-object to metrics constructor is deprecated
```
prom-client code : https://github.com/siimon/prom-client/blob/v11.0.0/lib/counter.js#L30,L42

### DIFF
--- a/src/metrics/prometheus.js
+++ b/src/metrics/prometheus.js
@@ -75,7 +75,7 @@ export default class PrometheusMetricsFactory {
       name = this._namespace + '_' + name;
     }
     if (!(key in this._cache)) {
-      this._cache[key] = new metric(name, help, labelNames);
+      this._cache[key] = new metric({ name, help, labelNames });
     }
     return labelValues.length > 0 ? this._cache[key].labels(...labelValues) : this._cache[key];
   }


src/metrics/prometheus.js
```diff
@@ -75,7 +75,11 @@ export default class PrometheusMetricsFactory {
       name = this._namespace + '_' + name;
     }
     if (!(key in this._cache)) {
-      this._cache[key] = new metric(name, help, labelNames);
+      this._cache[key] = new metric({
+        name: name,
+        help: help,
+        labelNames: labelNames,
+      });
```
yurishkuro: iirc this can be simplified to `{ name, help, labelNames }`, and Node will define hash keys equal to variable names.

eundoosong: Yes I know it can be simplified.
But I just refered to existing code like [this](https://github.com/jaegertracing/jaeger-client-node/blob/master/src/span.js#L238) and make same.
It is not a big deal, I can change. :)

PR #280, opened 2018-07-31
Add HTTP Sender

### LINKED ISSUE #170: Support for HTTP sender
Because Java and Go clients already support HTTP senders to emit span requests directly to the collector, it would be a nice addition if Node.js client could be capable to transport spans over HTTP too.

keitwb:
This will use the Jaeger collector HTTP endpoint (assumed to be running
on localhost) by default, but is configurable to send Thrift-encoded
spans to any HTTP endpoint.

Spans are batched upto a certain amount (based on # of spans and not
byte lengths) and are also flushed on a regular interval by the
RemoteReporter just like with the UDPSender.

It also supports basic auth in case the Jaeger collector has some form
of authentication in front of it.

Fixes #170

Signed-off-by: Ben Keith <bkeith@signalfx.com>

### DIFF
--- a/src/configuration.js
+++ b/src/configuration.js
@@ -20,6 +20,7 @@ import RemoteSampler from './samplers/remote_sampler';
 import Metrics from './metrics/metrics';
 import Tracer from './tracer';
 import UDPSender from './reporters/udp_sender';
+import HTTPSender from './reporters/http_sender';
 import opentracing from 'opentracing';
 import * as constants from './constants.js';
 import RemoteThrottler from './throttler/remote_throttler';
@@ -45,8 +46,12 @@ let jaegerSchema = {
     reporter: {
       properties: {
         logSpans: { type: 'boolean' },
+        agentProtocol: { type: 'string' },
         agentHost: { type: 'string' },
         agentPort: { type: 'number' },
+        agentPath: { type: 'string' },
+        username: { type: 'string' },
+        password: { type: 'string' },
         flushIntervalMs: { type: 'number' },
       },
       additionalProperties: false,
@@ -104,6 +109,7 @@ export default class Configuration {
   static _getReporter(config, options) {
     let reporterConfig = {};
     let reporters = [];
+    let senderCls = UDPSender;
     let senderConfig = {
       logger: options.logger,
     };
@@ -116,6 +122,20 @@ export default class Configuration {
         reporterConfig['bufferFlushInterval'] = config.reporter.flushIntervalMs;
       }
 
+      if (config.reporter.agentProtocol === 'http' || config.reporter.agentProtocol === 'https') {
+        senderCls = HTTPSender;
+        senderConfig['useHTTPS'] = config.reporter.agentProtocol === 'https';
+
+        if (config.reporter.agentPath) {
+          senderConfig['path'] = config.reporter.agentPath;
+        }
+        if (config.reporter.username) {
+          senderConfig['username'] = config.reporter.username;
+        }
+        if (config.reporter.password) {
+          senderConfig['password'] = config.reporter.password;
+        }
+      }
       if (config.reporter.agentHost) {
         senderConfig['host'] = config.reporter.agentHost;
       }
@@ -126,7 +146,7 @@ export default class Configuration {
     }
     reporterConfig['metrics'] = options.metrics;
     reporterConfig['logger'] = options.logger;
-    let sender = new UDPSender(senderConfig);
+    let sender = new senderCls(senderConfig);
     let remoteReporter = new RemoteReporter(sender, reporterConfig);
     if (reporters.length == 0) {
       return remoteReporter;
--- a/src/configuration.js
+++ b/src/configuration.js
@@ -46,10 +46,9 @@ let jaegerSchema = {
     reporter: {
       properties: {
         logSpans: { type: 'boolean' },
-        agentProtocol: { type: 'string' },
         agentHost: { type: 'string' },
         agentPort: { type: 'number' },
-        agentPath: { type: 'string' },
+        collectorEndpoint: { type: 'string' },
         username: { type: 'string' },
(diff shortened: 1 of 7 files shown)


test/http_sender.js
```diff
@@ -0,0 +1,275 @@
...
+
+  describe('span reference tests', () => {
+    let tracer = new Tracer('test-service-name', new InMemoryReporter(), new ConstSampler(true));
+    let parentContext = tracer.startSpan('just-used-for-context').context();
+    let childOfContext = tracer.startSpan('just-used-for-context').context();
+    let childOfRef = new opentracing.Reference(opentracing.REFERENCE_CHILD_OF, childOfContext);
+    let followsFromContext = tracer.startSpan('just-used-for-context').context();
+    let followsFromRef = new opentracing.Reference(opentracing.REFERENCE_FOLLOWS_FROM, followsFromContext);
+
+    let options = [
```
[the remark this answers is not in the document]
tiffon: I see now that your approach is consistent with the UDPSender test. But, UDPSender doesn't do anything with the parent context or references, either. And, I can't find anything special going on with those fields, so the variations should be safe to strip. 

What do you think?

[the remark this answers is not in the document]
keitwb: Yes, I don't see any special logic in either the HTTP or UDP senders that really needs testing all of those combinations -- it just hands them off to ThriftRW.  I will leave the two basic combinations and clean up all of the shadowed variables, and do the same for the UDPSender.

tiffon:
@keitwb Looks great!

# Thank you! 🎉

src/reporters/http_sender.js
```diff
@@ -0,0 +1,140 @@
...
+    this._maxSpanBatchSize = options.maxSpanBatchSize || DEFAULT_MAX_SPAN_BATCH_SIZE;
+
+    this._logger = options.logger || new NullLogger();
+    this._jaegerThrift = new Thrift({
+      source: fs.readFileSync(path.join(__dirname, '../jaeger-idl/thrift/jaeger.thrift'), 'ascii'),
+      allowOptionalArguments: true,
+    });
+  }
+
+  setProcess(process: Process): void {
```
yurishkuro: is this not the same as in the UDP sender? Any reason not to extract into a base class?

src/reporters/http_sender.js
```diff
@@ -0,0 +1,140 @@
...
+
+    const result = this._jaegerThrift.Batch.rw.toBuffer(this._batch);
+    if (result.err) {
+      SenderUtils.invokeCallback(callback, numSpans, `Error encoding Thrift batch: ${result.err}`);
+      return;
+    }
+
+    this._reset();
+
+    const options = {
```
black-adder: can these options be generated once in the constructor?

src/reporters/http_sender.js
keitwb: There is quite a bit of common logic so I'll extract it to the SenderUtils class.

src/configuration.js
```diff
@@ -45,8 +46,12 @@ let jaegerSchema = {
     reporter: {
       properties: {
         logSpans: { type: 'boolean' },
+        agentProtocol: { type: 'string' },
```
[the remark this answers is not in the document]
keitwb: Ok, check out the latest commit.  I make `collectorEndpoint` a full URL which make it unnecessary to have the `protocol` field.

src/reporters/http_sender.js
```diff
@@ -0,0 +1,131 @@
...
+    this._maxSpanBatchSize = options.maxSpanBatchSize || DEFAULT_MAX_SPAN_BATCH_SIZE;
+
+    this._logger = options.logger || new NullLogger();
+    this._jaegerThrift = new Thrift({
+      source: fs.readFileSync(path.join(__dirname, '../jaeger-idl/thrift/jaeger.thrift'), 'ascii'),
+      allowOptionalArguments: true,
+    });
+
+    this._httpOptions = {
+      protocol: this._url.protocol,
```
black-adder: ahh, how super useful

black-adder:
There's probably some more refactoring that we can do here like we've done in the [java client](https://github.com/jaegertracing/jaeger-client-java/tree/master/jaeger-thrift/src/main/java/io/jaegertracing/thrift/internal/senders) but I haven't looked too closely yet.

keitwb:
See the discussion with @tiffon above about how they did not test anything
specific to the sender's.

tiffon:
@yurishkuro Looks like the setup code ended up executing uncovered code in tracer.js ([codecov](https://codecov.io/gh/jaegertracing/jaeger-client-node/pull/280/src/src/tracer.js)). 

I'd say it's not a good thing to rely on setup code in UDPSender and HTTPSender to test tracer.js. Worse still, removing the test cases had no effect on the coverage of the UDPSender or HTTPSender, which makes the code misleading. Its presence suggests it's relevant to the resources under test, but nothing in either class is sensitive to variations in the test cases. Further, because the test cases have no clear connection to the UDPSender, it leads to the assumption that it's there for a reason and the reason is simply not known. People are risk averse when it comes to these matters, so the test variations are propagated to new code. I'd say the extra test cases easily qualify as tech debt, tech debt that can spread. I suggest we keep the test cases removed. We can regain the coverage more directly via testing tracer.js.

yurishkuro:
@tiffon You may be 100% correct about the tests being misplaced, but the end result is the same - code that was previously tested now isn't. Removing tests is always a red flag to me. If you think tests are misplaced (I can't comment on that, don't know what they were covering), a better course of action is to refactor them to a more sensible place in a standalone PR, so that coverage is preserved and tech debt cleaned.

tiffon:
@yurishkuro You're right – that's a better course of action. I'll issue a PR, shortly.

tiffon:
Created #284 to address the reduction in coverage.

PR #284, opened 2018-08-07
Bolster tracer.js coverage per UDPSender cleanup

tiffon:

A regression in test coverage.

Bolsters `tracer.js` coverage after removing [test cases](https://github.com/jaegertracing/jaeger-client-node/pull/280/files#diff-4abc69534f2675ef27d324a3a3a87638L125) from UDPSender which did not test anything in UDPSender.

### DIFF
--- a/test/tracer.js
+++ b/test/tracer.js
@@ -139,36 +139,81 @@ describe('tracer should', () => {
     assert.equal(span._startTime, startTime);
   });
 
-  it('start a child span represented as a separate span from parent, using childOf and references', () => {
-    let traceId = Utils.encodeInt64(1);
-    let spanId = Utils.encodeInt64(2);
-    let parentId = Utils.encodeInt64(3);
-    let flags = 1;
-    let context = SpanContext.withBinaryIds(traceId, spanId, parentId, flags);
-    let startTime = 123.456;
+  describe('start a child span represented as a separate span from parent, using childOf and references', () => {
+    let nextId = 0;
+    const getId = () => Utils.encodeInt64(nextId++);
+    const traceId = getId();
+    const flags = 1;
 
-    let childOfParams = {
-      operationName: 'test-name',
-      childOf: context,
-      startTime: startTime,
-    };
+    const parentContext = SpanContext.withBinaryIds(traceId, getId(), null, flags);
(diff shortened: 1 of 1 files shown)


test/tracer.js
```diff
@@ -139,36 +139,81 @@ describe('tracer should', () => {
...
-      let span = tracer.startSpan('test-span', params);
-      assert.deepEqual(span.context().traceId, traceId);
-      assert.deepEqual(span.context().parentId, spanId);
-      assert.equal(span.context().flags, constants.SAMPLED_MASK);
-      assert.equal(span._startTime, startTime);
-    };
-
-    assertByStartSpanParameters(childOfParams);
-    assertByStartSpanParameters(referenceParams);
+  describe('start a child span represented as a separate span from parent, using childOf and references', () => {
```
yurishkuro: why "describe" when all other same-level methods are "it" ?

Side note: L28 `describe('tracer should',` should really be `describe('tracer ',`, then "should" goes under it().

tiffon: The six test cases are very related in what they're testing and lend themselves to being grouped.

```
  tracer should
    ✓ be able to override codec contextKey and extract context
    ✓ begin a new span given only baggage headers
    ✓ create a span correctly through _startInternalSpan
    ✓ report a span with no tracer level tags
    ✓ start a root span with proper structure
    ✓ inject plain text headers into carrier, and extract span context with the same value
    ✓ inject url encoded values into headers
    ✓ assert inject and extract throw errors when given an invalid format
    ✓ report spans
    ✓ set _process on initialization
    ✓ close _debugThrottler on close
    start a child span represented as a separate span from parent, using childOf and references
      ✓ starts a span based on childOf
      ✓ starts a span based on childOf, ignoring FOLLOWS_FROM
      ✓ starts a span based on childOf, ignoring CHILD_OF and FOLLOWS_FROM
      ✓ starts a span with parent falling back to the CHILD_OF ref
      ✓ starts a span with parent falling back to the FOLLOWS_FROM ref
      ✓ starts a span with parent falling back to the CHILD_OF ref and ignoring FOLLOWS_FROM
    Metrics
      ✓ startSpan
      ✓ emits counter when report called
```

What's the downside of using `describe(...)`?

(Side note: The Metrics are also using `describe(...)`.)

PR #286, opened 2018-08-10
Add HTTP sender documentation

### LINKED ISSUE #285: More detailed examples of configuration

In some cases data in spans are important, so UDP is not suitable for this.

There are only UDP examples for Tracer instance created by initTracer.
None of the other examples (TChannel) uses Tracer interface. So how can I use opentracing with them?
How can I create tracer which use HTTP/TCP for example?

black-adder:

Resolves #285

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,12 +14,41 @@ Please see [CONTRIBUTING.md](./CONTRIBUTING.md).
 
 ## Initialization
 
+The Tracer defaults to sending spans over UDP to the jaeger-agent running on localhost.
+
 ```javascript
 var initTracer = require('jaeger-client').initTracer;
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,12 +14,41 @@ Please see [CONTRIBUTING.md](./CONTRIBUTING.md).
...
+  tags: {
+    'my-awesome-service.version': '1.1.2',
+  },
+  metrics: metrics,
+  logger: logger,
+};
+var tracer = initTracer(config, options);
+```
+
+#### Reporting spans via HTTP
```
yurishkuro: I would add some explanation why someone might want that, and also clarify that in this case the data is not going through the agent.

PR #289, opened 2018-08-10
Rename 'throttler_update' metric to 'throttler_updates'

black-adder:

- Fix inconsistency seen here: https://github.com/jaegertracing/jaeger-client-node/pull/279/files#r209324272

No tests need to updated because we mock metrics, we would need to introduce a local metric store (maybe use prometheus?) to test against the actual metric.

### DIFF
--- a/src/metrics/metrics.js
+++ b/src/metrics/metrics.js
@@ -113,7 +113,7 @@ export default class Metrics {
 
     this.throttledDebugSpans = this._factory.createCounter('jaeger:throttled_debug_spans');
 
-    this.throttlerUpdateSuccess = this._factory.createCounter('jaeger:throttler_update', {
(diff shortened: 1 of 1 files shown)


src/metrics/metrics.js
```diff
@@ -113,7 +113,7 @@ export default class Metrics {
 
     this.throttledDebugSpans = this._factory.createCounter('jaeger:throttled_debug_spans');
 
-    this.throttlerUpdateSuccess = this._factory.createCounter('jaeger:throttler_update', {
+    this.throttlerUpdateSuccess = this._factory.createCounter('jaeger:throttler_updates', {
```
yurishkuro: no test breaking?

black-adder: I addressed this in the blurb above

PR #294, opened 2018-08-16
Fix HTTP sender to work with TLS

keitwb:
The http.Agent instance doesn't work with HTTPS and requires a separate type of agent.

### DIFF
--- a/src/reporters/http_sender.js
+++ b/src/reporters/http_sender.js
@@ -13,6 +13,7 @@
 
 import fs from 'fs';
 import http from 'http';
+import https from 'https';
 import path from 'path';
 import * as URL from 'url';
 import { Thrift } from 'thriftrw';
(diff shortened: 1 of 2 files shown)


keitwb:
Yes, I committed a self-signed cert/key pair to the repo that we can use to start up an TLS server.

yurishkuro:
fails with 
```
1) http sender should handle HTTPS collectors:
     Error: ENOENT, no such file or directory '/home/travis/build/jaegertracing/jaeger-client-node/dist/test/server.key'
      at Object.fs.openSync (fs.js:439:18)
      at Object.fs.readFileSync (fs.js:290:15)
      at Context.<anonymous> (dist/test/http_sender.js:314:25)
```

Maybe consider just embedding the cert/key as string constants in the test file, as otherwise they need to be copied to dist, unnecessary churn.

keitwb:
Hmm...I'm going to have to figure out something else for Node 0.10 to support HTTPS.

yurishkuro:
> Hmm...I'm going to have to figure out something else for Node 0.10 to support HTTPS.

I think it's fine to have fewer features for 0.10 - throw explicit exception that HTTPS is not supported there and have a branched unit tests for that.

yurishkuro:
also, if you do exclude 0.10 for https, please add a note in the readme

keitwb:
@yurishkuro I was able to make it work by using the specific `https.request` method where appropriate.

yurishkuro:
Thanks!

Issue #282, opened 2018-08-06
Support ENV variables for configuration


black-adder:
## Requirement - what kind of business use case are you trying to solve?

Java and golang clients support tracer initialization from environment variables, would be nice for node to fall into line.

Related to #152

claudio-viola:
Related https://github.com/jaegertracing/jaeger-client-node/issues/293

eundoosong:
@black-adder 
I'd like to work on this issue and think simply we can use `||` operator like this.
```node
static _getSampler(config, options) {
    let type = config.sampler.type || process.env.JAEGER_SAMPLER_TYPE;
    let param = config.sampler.param || process.env.JAEGER_SAMPLER_PARAM;
    let host = config.sampler.host || process.env.JAEGER_SAMPLER_HOST;
    let port = config.sampler.port || process.env.JAEGER_SAMPLER_PORT;
    let refreshIntervalMs = config.sampler.refreshIntervalMs || process.env.JAEGER_SAMPLER_REFRESH_INTERVAL;
```

yurishkuro:
Go for it

PR #296, opened 2018-10-04
Support ENV variables for configuration

eundoosong:

- node client supports tracer initialization from environment variables, as the other clients such as java and go.
- Resolves: #282

- Add configuration_env.js for tracer initialization from environment variables.
- Allow direct configuration setting which overrides environment variables like this:
```python
    process.env.JAEGER_SERVICE_NAME = 'test-service';
    process.env.JAEGER_DISABLE = false;

    let config = {
      serviceName: 'test-service-arg',
    };
    let tracer = initTracerFromEnv(config);
    assert.equal(tracer._serviceName, 'test-service-arg');
```
- Update readme.md


src/configuration_env.js
```diff
@@ -0,0 +1,171 @@
...
+// is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+// or implied. See the License for the specific language governing permissions and limitations under
+// the License.
+
+import Configuration from './configuration.js';
+import Utils from './util.js';
+
+export default class ConfigurationEnv {
+  static _getConfigValue(obj, key, env) {
+    return obj ? obj[key] || env : env;
```
yurishkuro: this seems more intuitive: `return obj && obj[key] || env`. Also, I'd replace `env` with `defaultValue`

eundoosong: That is more intuitive. Thanks!

PR #302, opened 2018-10-14
Fix HTTPSender ignoring query parameters

### LINKED ISSUE #301: jaeger-collector returning 'Unsupported format type'

Trying to use jaeger-client-node together with istio and ambassador

Jaeger is not respecting the defined `format` in HTTPSender.
Right now I use `collectorEndpoint: 'http://jaeger-collector.istio-system:14268/api/traces?format=jaeger.thrift',` and the `?format=jaeger.thrift` is not sent with any of the flushes.
Further more it's quite hard to debug, as there's no errors when Jaeger responds with `Unsupported format type:`, it's just silently ignored.

I propose to change https://github.com/jaegertracing/jaeger-client-node/blob/206e4c1c4562c0ccbf6e4f3617da5d72baf9bb39/src/reporters/http_sender.js#L66
Instead of path: `this._url.pathname,` I believe it should be `path: this._url.path,`.

I've tried the fix above locally, and it fixes my issues.

Multiply:

Resolves #301 

Changed HTTPSender to include query parameters from `collectorEndpoint`.

### DIFF
--- a/src/reporters/http_sender.js
+++ b/src/reporters/http_sender.js
@@ -63,7 +63,7 @@ export default class HTTPSender {
       protocol: this._url.protocol,
       hostname: this._url.hostname,
       port: this._url.port,
-      path: this._url.pathname,
+      path: this._url.path,
       method: 'POST',
       auth: this._username && this._password ? `${this._username}:${this._password}` : undefined,
       headers: {


Multiply:
@yurishkuro I might be running an older version of jaeger-collector, in that case. :(

src/reporters/http_sender.js
```diff
@@ -63,7 +63,7 @@ export default class HTTPSender {
       protocol: this._url.protocol,
       hostname: this._url.hostname,
       port: this._url.port,
-      path: this._url.pathname,
+      path: this._url.path,
```
yurishkuro: I am not seeing `path` being a valid property of the URL object: https://nodejs.org/api/url.html#url_class_url

yurishkuro:
Do you have problems with upgrading?

Multiply:
@yurishkuro I am investigating how big a deal it is. It's currently running in 3 docker images that we're not in control of, so I might have to wait for those to be upgraded to latest jaeger.

src/reporters/http_sender.js
Multiply: https://nodejs.org/api/url.html#url_urlobject_path

Multiply:
Closing this, as I don't see a point in supporting older versions of jaeger.

PR #303, opened 2018-10-19
Add setProcess method to LoggingReporter (#300)

### LINKED ISSUE #300: function setProcess required for logging reporter
I want to use logging reporter if there no jaeger server exists.

logging reporter currently cannot work. 

image

according to reporter interface, we should add setProcess function to logging reporter.
image

isayme:

- fix #300 

-


yurishkuro:
Thanks!

PR #305, opened 2018-10-24
Change Zipkin Codec to Not Inject Missing Parent

### LINKED ISSUE #304: Zipkin codec should not inject parentspanid if it was not extracted
I am trying to use Jaeger with Istio, but I do not want to forward headers explicitly. I want to make use of the OpenTracing API wherever possible.

The Zipkin codec injects an empty string for the parentspanid even when the object that it extracts from does not have x-b3-parentspanid specified.

This problem is low impact. I can just delete the property from the object before using it as my HTTP headers.

I suggest to check to make sure that the parentspanid is not falsy before injecting into an object with the zipkin carrier.

I don't know if this is something that should be generalized for other fields as well, but the parentspanid is the only one that affected me. Should this same check be applied to, for example, x-b3-traceid?

adinunzio84:
The Zipkin b3 TextMap codec was injecting an empty string for the
x-b3-parentspanid field when one was not present in the object it
extracted the SpanContext from. This caused gaps in the traces.

This change first checks the span context's parentIdStr value,
and does not inject it into the carrier if it is falsy.

Signed-off-by: Andrew DiNunzio <adinunzio84@gmail.com>

Resolves #304

Checks the parentspanid to only inject into the carrier if it is not falsy in the SpanContext.

### DIFF
--- a/src/propagators/zipkin_b3_text_map_codec.js
+++ b/src/propagators/zipkin_b3_text_map_codec.js
@@ -155,7 +155,9 @@ export default class ZipkinB3TextMapCodec {
 
   inject(spanContext: SpanContext, carrier: any): void {
     carrier[ZIPKIN_TRACE_HEADER] = spanContext.traceIdStr;
-    carrier[ZIPKIN_PARENTSPAN_HEADER] = spanContext.parentIdStr;
+    if (spanContext.parentIdStr) {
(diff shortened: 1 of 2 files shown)


adinunzio84:
I made this change pretty specific to the problem I was facing. If a more generalized solution is better (where any missing fields should not be injected with empty strings), I can try to do that.

PR #308, opened 2018-11-01
Add missed contextKey option to initTracer

### LINKED ISSUE #254: The static method initTracer in configuration.js missing support for contextKey
It looks like support for an alternative to the default TRACER_STATE_HEADER_NAME defined in constant.js has been added to tracer.js and text_map_code.js by allowing a 'contextKey' to be passed as an option. However, the initTracer method in configuration.js does not include support for this when creating a new Tracer object. The current code reads like this:

`
return new Tracer(config.serviceName, reporter, sampler, {
      metrics: options.metrics,
      logger: options.logger,
      tags: options.tags,
    });
`

Fixing this seems like a simple update - something like this:

`
return new Tracer(config.serviceName, reporter, sampler, {
      metrics: options.metrics,
      logger: options.logger,
      tags: options.tags,
      contextKey: options.contextKey
    });
`

Is there a reason that contextKey is not supported in the initTracer method?

yepninja:
Resolves #254

### DIFF
--- a/src/configuration.js
+++ b/src/configuration.js
@@ -225,6 +225,7 @@ export default class Configuration {
     }
 
     return new Tracer(config.serviceName, reporter, sampler, {
+      contextKey: options.contextKey,
       metrics: options.metrics,
       logger: options.logger,
       tags: options.tags,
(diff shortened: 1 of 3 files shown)


yurishkuro:
@yepninja if your company is using Jaeger, please consider commenting on https://github.com/jaegertracing/jaeger/issues/207

PR #311, opened 2018-11-15
Use common env variables

yepninja:
Rename variables according to common names from https://www.jaegertracing.io/docs/1.8/client-features/


yurishkuro:
Fixes #313

yepninja:
@yurishkuro @tiffon

PR #314, opened 2018-11-16
fixed span.log return this

### LINKED ISSUE #312: Span.log should return this

doubret:
Resolves #312 

Changed return type of Span.log from void to Span.
Added test to check returned value from Span.log is equal to the span itself.


yurishkuro:
please make sure to sign all your commits. Easiest is to squash them into one.

doubret:
Sorry i messed up signing the commits. Hope it's ok for you.

PR #315, opened 2018-11-16
fix Span.log returns this

### LINKED ISSUE #312: Span.log should return this

doubret:
Signed-off-by: Charles-Edouard Breteche <doubret@yahoo.co.uk>

Resolves #312 

fix Span.log returns this
added test to check return form span.log is equal to the span itself


doubret:
Still not

yurishkuro:
> Author: doubret, Committer: doubret; Expected "doubret doubret@yahoo.co.uk", but got "Charles-Edouard Breteche doubret@yahoo.co.uk".

Sometimes DCO bot is picky about your GitHub name matching git signature

PR #316, opened 2018-11-16
fix Span.log returns this

### LINKED ISSUE #312: Span.log should return this

doubret:
Signed-off-by: Charles-Edouard Breteche <doubret@yahoo.co.uk>

Resolves #312

fix Span.log should return this
added test to check span.log returns the span itself


doubret:
Making the PR took longer than writing the code changes ;-)

doubret:
Crossdock tests are failing.
Not sure how to fix that.

Issue #313, opened 2018-11-16
Environment variable names do not match other Jaeger clients & docs


koliyo:
## Requirement - what kind of business use case are you trying to solve?

Monitoring

## Problem - what in Jaeger blocks you from solving the requirement?

The environment variables in the node client does not match the documentation, or the golang client.

Golang and node documentation says `JAEGER_AGENT_HOST`, but the code use `JAEGER_REPORTER_AGENT_HOST`. Same for a couple of other variables.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Update the implementation to match documentation. Good if it matches the golang client!

yurishkuro:
There is a fix pending #311

yepninja:
List of wrong env variables:
- [ ] `JAEGER_REPORTER_AGENT_PORT` -> `JAEGER_AGENT_PORT`
- [ ] `JAEGER_REPORTER_AGENT_HOST` -> `JAEGER_AGENT_HOST`
- [ ] `JAEGER_REPORTER_ENDPOINT` -> `JAEGER_ENDPOINT`
- [ ] `JAEGER_REPORTER_USER` -> `JAEGER_USER`
- [ ] `JAEGER_REPORTER_PASSWORD` -> `JAEGER_PASSWORD`
- [ ] `JAEGER_SAMPLER_HOST` & `JAEGER_AGENT_PORT` -> `JAEGER_SAMPLER_MANAGER_HOST_PORT`

yurishkuro:
@yepninja can this be closed?

yurishkuro:
But is #321 already addressed by #320 (which is merged)?

yepninja:
#320 now old variables are also supported. #321 is about removing them in a major release.

yurishkuro:
ok so it sounds like this issue should be closed? since the "official" (same as other clients) variables are not supported. Removing the deprecated variables is a much lower priority and is tracked by ticket #321.

yepninja:
Yes, this issue should be closed.

PR #320, opened 2018-12-02
Mismatching env variables

yepninja:
Support common env variables instead of mismatching variables.
Doc: https://www.jaegertracing.io/docs/1.8/client-features/
Fixes issue: #313 
Outdated PR: #311


yepninja:
Force pushed to the previous PR

Issue #328, opened 2018-12-12
4.0.0 Release


black-adder:
Ticket to keep track of breaking changes and other necessary changes for 4.0.0 release

- [ ] Remove unsupported ENV vars #311, #321

yepninja:
ENV vars are not removed. Both true and wrong vars are supported. Tests are only added. So it can be a minor release.
For a major release, it is better to remove support of mismatching variables.

PR #330, opened 2018-12-18
Remove deprecated env variables (4.x release)

### LINKED ISSUE #321: Remove deprecated environment variables
Unify configuration with env variables

Mismatching environment variable naming (#313)

Bring the configuration with env variables according to jaeger [doc](https://www.jaegertracing.io/docs/1.8/client-features/) in the next major releas

List of env variables to remove:
- [x] `JAEGER_REPORTER_AGENT_PORT`
- [x] `JAEGER_REPORTER_AGENT_HOST`
- [x] `JAEGER_REPORTER_ENDPOINT`
- [x] `JAEGER_REPORTER_USER`
- [x] `JAEGER_REPORTER_PASSWORD`
- [x] `JAEGER_SAMPLER_HOST`
- [x] `JAEGER_SAMPLER_PORT`
- [x] `JAEGER_DISABLE`

yepninja:
Resolves #321

### DIFF
--- a/src/configuration_env.js
+++ b/src/configuration_env.js
@@ -39,12 +39,12 @@ export default class ConfigurationEnv {
       samplerConfig.hostPort = value;
     }
 
-    value = ConfigurationEnv._getConfigValue(config.sampler, 'host', process.env.JAEGER_SAMPLER_HOST);
+    value = ConfigurationEnv._getConfigValue(config.sampler, 'host');
     if (value) {
       samplerConfig.host = value;
     }
 
-    value = ConfigurationEnv._getConfigValue(config.sampler, 'port', process.env.JAEGER_SAMPLER_PORT);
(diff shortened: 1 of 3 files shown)


yepninja:
Release #328

black-adder:
Thanks for this! Let's put this on ice for a while until we can figure out what other breaking changes we can introduce given how precious a major release is.

yurishkuro:
indeed.  In fact, I would suggest a different approach - we should merge another PR that clearly tags the deprecated constants somehow as "to be removed in 4.x". Then it will be a lot easier to remove them closer to 4.x than to keep this PR open and possibly do complicated merges.

yurishkuro:
Very old PR, closing. Given that this SDK is not being actively evolved further, it's not so bad to continue supporting old names.

PR #332, opened 2018-12-22
Warning for mismatching env variables

yepninja:
Warning for a minor release.

### DIFF
--- a/src/configuration_env.js
+++ b/src/configuration_env.js
@@ -35,7 +35,7 @@ export default class ConfigurationEnv {
       .keys(deprecatedEnvVars)
       .forEach(env => {
         if (process.env[env]) {
-          console.warn(`You are using deprecated env variable ${env}. Use ${deprecatedEnvVars[env]} instead. \nMismatching env variable will be removed in the next major release (4.x.x)`)
+          console.warn(`You are using deprecated env variable ${env}. Use ${deprecatedEnvVars[env]} instead. \nDeprecated env variable will be removed in the next major release (4.x.x)`)
         }
       })
 
--- a/src/configuration_env.js
+++ b/src/configuration_env.js
@@ -32,7 +32,7 @@ export default class ConfigurationEnv {
   static _getSamplerFromEnv(config) {
 
     Object
-      .keys(wrongEnvVars)
+      .keys(deprecatedEnvVars)
       .forEach(env => {
         if (process.env[env]) {
           console.warn(`You are using deprecated env variable ${env}. Use ${deprecatedEnvVars[env]} instead. \nMismatching env variable will be removed in the next major release (4.x.x)`)
--- a/src/configuration_env.js
+++ b/src/configuration_env.js
@@ -13,7 +13,7 @@
 import Configuration from './configuration.js';
 import Utils from './util.js';
 
-const wrongEnvVars = {
+const deprecatedEnvVars = {
   JAEGER_SAMPLER_HOST: 'JAEGER_SAMPLER_MANAGER_HOST_PORT',
   JAEGER_SAMPLER_PORT: 'JAEGER_SAMPLER_MANAGER_HOST_PORT',
   JAEGER_REPORTER_ENDPOINT: 'JAEGER_ENDPOINT',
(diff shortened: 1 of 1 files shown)


src/configuration_env.js
```diff
@@ -13,12 +13,32 @@
...
     return (obj && obj[key]) || defaultValue;
   }
 
   static _getSamplerFromEnv(config) {
+
+    Object
+      .keys(wrongEnvVars)
+      .forEach(env => {
+        if (process.env[env]) {
+          console.warn(`You are using wrong missmatching env variable ${env}. Use ${wrongEnvVars[env]} instead. \nMismatching env variable will be removed in the next major release (4.x.x)`)
```
yepninja: Maybe better to collect warnings, and than log them with the logger. But not everybody uses logger.

yurishkuro: Console is fine, I think. Replace “wrong mismatching” with “deprecated”

src/configuration_env.js
```diff
@@ -13,12 +13,32 @@
 import Configuration from './configuration.js';
 import Utils from './util.js';
 
+const wrongEnvVars = {
```
yurishkuro: DeprecatedEnvVars

src/configuration_env.js
```diff
@@ -13,12 +13,32 @@
...
+  JAEGER_SAMPLER_HOST: 'JAEGER_SAMPLER_MANAGER_HOST_PORT',
+  JAEGER_SAMPLER_PORT: 'JAEGER_SAMPLER_MANAGER_HOST_PORT',
+  JAEGER_REPORTER_ENDPOINT: 'JAEGER_ENDPOINT',
+  JAEGER_REPORTER_USER: 'JAEGER_USER',
+  JAEGER_REPORTER_PASSWORD: 'JAEGER_PASSWORD',
+  JAEGER_REPORTER_AGENT_HOST: 'JAEGER_AGENT_HOST',
+  JAEGER_REPORTER_AGENT_PORT: 'JAEGER_AGENT_PORT',
+  JAEGER_DISABLE: 'JAEGER_DISABLED'
+};
```
yurishkuro: Why does this need to be a map and not an array?

yurishkuro: Or, nvm

src/configuration_env.js
yepninja: Done

src/configuration_env.js
yepninja: Done

src/configuration_env.js
```diff
@@ -13,12 +13,32 @@
...
     return (obj && obj[key]) || defaultValue;
   }
 
   static _getSamplerFromEnv(config) {
+
+    Object
+      .keys(deprecatedEnvVars)
+      .forEach(env => {
+        if (process.env[env]) {
+          console.warn(`You are using deprecated env variable ${env}. Use ${deprecatedEnvVars[env]} instead. \nMismatching env variable will be removed in the next major release (4.x.x)`)
```
yurishkuro: s/Mismatching/Deprecated in the second sentence as well.

src/configuration_env.js
```diff
@@ -13,12 +13,32 @@
...
+};
+
 export default class ConfigurationEnv {
   static _getConfigValue(obj, key, defaultValue) {
     return (obj && obj[key]) || defaultValue;
   }
 
   static _getSamplerFromEnv(config) {
+
+    Object
```
yurishkuro: why is this check inside `_getSamplerFromEnv` function?

yepninja: Node.js gets env variables in runtime (call of `getenv` for each variable). So check is when process.env is used.

Related issues:
- https://github.com/nodejs/node/issues/3104
- https://github.com/nodejs/node/issues/22960

src/configuration_env.js
yepninja: Fixed

yepninja:
Opps. Almost there is no habit to code without eslint.
The build is fixed!

yepninja:
Nope. That issue is related to PR #330

PR #336, opened 2019-01-10
Use the ip and hostname tags if provided

ledor473:
Fixes https://github.com/jaegertracing/jaeger-client-node/issues/335

- Check if there's already a value before using the default logic


ledor473:
Oops I miss-click on `close`...

But yeah the CI seems to be waiting for a password 🤔

black-adder:
I think it's related to: https://github.com/jaegertracing/jaeger/pull/153. This fix would be to add a check for `"$TRAVIS_SECURE_ENV_VARS" == "true"` like we did before or a quick workaround is to use another branch `ticketmaster:master` -> `ticketmaster:jaeger-fix`

ledor473:
@black-adder backporting the fix worked 👍

black-adder:
awesome, thanks!

PR #339, opened 2019-01-24
[WIP] Remove package.json from dist so npm-publish works

tiffon:
The `package.json` file in `dist/` was causing problems with publish.


package.json
```diff
@@ -13,14 +13,20 @@
...
+    "Yuri Shkuro <ys@uber.com>",
+    "Won Jun Jang <wjang@uber.com>"
+  ],
   "dependencies": {
     "node-int64": "^0.4.0",
-    "thriftrw": "^3.5.0",
-    "xorshift": "^0.2.0",
     "opentracing": "^0.13.0",
-    "uuid": "^3.2.1"
+    "prom-client": "11.0.0",
```
yurishkuro: Wasn't there before

src/tracer.js
```diff
@@ -31,6 +30,14 @@ import BaggageSetter from './baggage/baggage_setter';
 import DefaultThrottler from './throttler/default_throttler';
 import uuidv4 from 'uuid/v4';
 
+let pjson;
+try {
+  pjson = require('../package.json');
+} catch (_) {
+  // $FlowIgnore
+  pjson = require('../../package.json');
```
yurishkuro: I think this is dangerous. If we do this, let's add an assert that the package name is actually jaeger.

tiffon:
This PR was executed in a haste. A more thorough approach is warranted.

PR #346, opened 2019-02-19
Support false values for the B3 Sampled header

sebnow:
The sampling state headers can have three possible values; force sampling, deny sampling, and defer sampling decision. The B3 propagator only looks at the existence of the header, and thus always forces sampling, even when the value is falsy.

> An accept sampling decision is encoded as `X-B3-Sampled: 1` and a deny as `X-B3-Sampled: 0`. Absent means defer the decision to the receiver of this header. For example, a Sampled header might look like: `X-B3-Sampled: 1`.
> 
> Note: Before this specification was written, some tracers propagated `X-B3-Sampled` as `true` or `false` as opposed to `1` or `0`. While you shouldn't encode `X-B3-Sampled` as `true` or `false`, a lenient implementation may accept them.

https://github.com/openzipkin/b3-propagation#sampling-state-1

Check that the value is true before setting the sampling flag.

### DIFF
--- a/src/propagators/zipkin_b3_text_map_codec.js
+++ b/src/propagators/zipkin_b3_text_map_codec.js
@@ -112,7 +112,9 @@ export default class ZipkinB3TextMapCodec {
             traceId = this._decodeValue(carrier[key]);
             break;
           case ZIPKIN_SAMPLED_HEADER:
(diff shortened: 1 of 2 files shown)


black-adder:
thanks!

PR #347, opened 2019-02-19
Update make build-node to feed version to src/tracer.js

### LINKED ISSUE #338: 3.14.0 build on npm is missing essential build artifacts and is therefore broken

Install the node jaeger client.

The published `dist` dir is missing the relevant contents see [unpkg](https://unpkg.com/jaeger-client@3.14.0/dist/src/). The `package.json` defines a main file of `"./dist/src/index.js"`; this file doesn't exist in the current published version (`3.14.0`).

**Impact**: `jaeger-client` is broken for all new installs with semver ranges satisfied by `3.14.0`.

I do not know.

N/A

everett980:
- Resolves #338 
- Makefile creates dist/src/tracer.js with version from package.json instead of a hardcoded value.

- Add src/version.js with stand in version.
- Change the version in dist/src/version.js to be version from package.json.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,11 +45,11 @@ check-node-6:
 
 .PHONY: build-node
 build-node: check-node-6 node-modules
-	sed -i.bak "s|module.exports = '.*'|module.exports = '$(shell node -p 'require("./package.json").version')'|" src/version.js && rm src/version.js.bak
 	rm -rf ./dist/
 	node_modules/.bin/babel --presets env --plugins transform-class-properties --source-maps -d dist/src/ src/
 	node_modules/.bin/babel --presets env --plugins transform-class-properties --source-maps -d dist/test/ test/
 	node_modules/.bin/babel --presets env --plugins transform-class-properties --source-maps -d dist/crossdock/ crossdock/
+	cat src/version.js | sed "s|VERSION_TBD|$(shell node -p 'require("./package.json").version')|g" > dist/src/version.js
 	cp -R ./test/thrift ./dist/test/thrift/
 	cp package.json ./dist/
 	cp -R ./src/jaeger-idl ./dist/src/
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ check-node-6:
 
 .PHONY: build-node
 build-node: check-node-6 node-modules
-	sed -i.bak "s|const pjsonVersion = '.*'|const pjsonVersion = '$(shell node -p 'require("./package.json").version')'|" src/tracer.js && rm src/tracer.js.bak
+	sed -i.bak "s|module.exports = '.*'|module.exports = '$(shell node -p 'require("./package.json").version')'|" src/version.js && rm src/version.js.bak
 	rm -rf ./dist/
 	node_modules/.bin/babel --presets env --plugins transform-class-properties --source-maps -d dist/src/ src/
 	node_modules/.bin/babel --presets env --plugins transform-class-properties --source-maps -d dist/test/ test/
--- a/Makefile
+++ b/Makefile
@@ -45,6 +45,7 @@ check-node-6:
 
 .PHONY: build-node
 build-node: check-node-6 node-modules
+	sed -i.bak "s|const pjsonVersion = '.*'|const pjsonVersion = '$(shell node -p 'require("./package.json").version')'|" src/tracer.js && rm src/tracer.js.bak
 	rm -rf ./dist/
 	node_modules/.bin/babel --presets env --plugins transform-class-properties --source-maps -d dist/src/ src/
 	node_modules/.bin/babel --presets env --plugins transform-class-properties --source-maps -d dist/test/ test/
--- a/src/tracer.js
+++ b/src/tracer.js
@@ -29,8 +29,7 @@ import os from 'os';
 import BaggageSetter from './baggage/baggage_setter';
 import DefaultThrottler from './throttler/default_throttler';
 import uuidv4 from 'uuid/v4';
-
-const pjsonVersion = '3.14.4';
+import version from './version';
 
 export default class Tracer {
   _serviceName: string;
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -45,6 +45,7 @@ check-node-6:
 
 .PHONY: build-node
 build-node: check-node-6 node-modules
+	sed -i.bak "s|const pjsonVersion = '.*'|const pjsonVersion = '$(shell node -p 'require("./package.json").version')'|" src/tracer.js && rm src/tracer.js.bak
```
everett980: creating the `.bak` file and removing it is necessary for cross-platform support for in-place edits. Different versions of `sed -i` require a different syntax to specify not to create a backup file. Either `sed -i''` or `sed -i ''`. That space makes all the difference.

yurishkuro:
I would prefer a separate, fully generated file to using sed. What do you mean by needing to git-add it? We could have a placeholder file checked in, which defines some nonsense version, to be overwritten by build.

everett980:
> I would prefer a separate, fully generated file to using sed. What do you mean by needing to git-add it? We could have a placeholder file checked in, which defines some nonsense version, to be overwritten by build.

@yurishkuro As soon as I hit submit I realized this has the same issue. I was trying to avoid adding a file in the build process since it would result in untraced files. Editing a file in the `makefile` will also have a change, it'll just be tracked.
Since the [release process](https://github.com/jaegertracing/jaeger-client-node/blob/master/RELEASE.md) requires issuing a PR after `npm publish`, I suppose that PR can also include updating `src/version.js`.

yurishkuro:
I would rather generate the new file in `/dist` only. You can still use `sed` but without the non-portable `-i` flag, e.g. `cat src/version.js | sed 's/VERSION-TBD/$version/g' > dist/version.js`. We don't want to check-in the file with the updated version, I think it's ok to have it as an artifact of the build & publish.

everett980:
That works nicely. I had assumed that the version in `src/version.js` would need to be a valid version for development / testing but without that requirement this is much simpler. I'll update the PR description for the sake of record keeping.

yurishkuro:
Please test it by publishing an unofficial version (similar to how @black-adder did the last time). I tested the published version by importing it from the internal `jaeger-client-adapter` lib.

tiffon:
@everett980 There's also a babel plugin worth considering, it inlines environment variables: [babel-plugin-transform-inline-environment-variables](https://babeljs.io/docs/en/babel-plugin-transform-inline-environment-variables)

>In
>```js
>// assuming process.env.NODE_ENV is actually "development"
>process.env.NODE_ENV;
>```
>
>Out
>```js
>"development";
>```

So, the change to Makefile would be

```make
build-node: PACKAGE_VERSION = $(shell node -p 'require('./package).version') 
build-node: check-node-6 node-modules
	# unchanged...
```

And to tracer.js

```js
this._tags[constants.JAEGER_CLIENT_VERSION_TAG_KEY] = `Node-${process.env.PACKAGE_VERSION}`;

**Comment by everett980** [comment]
@tiffon I had to do `PACKAGE_VERSION = $(shell ...)` on the same line as building `dist/src` otherwise it was considered undefined. But otherwise it works well!

For reasons unknown getting flow to ignore `process.env` was harder than it should've been.

File: src/tracer.js [hunk 2]
```diff
@@ -64,7 +64,9 @@ export default class Tracer {
     options: any = {}
   ) {
     this._tags = options.tags || {};
-    this._tags[constants.JAEGER_CLIENT_VERSION_TAG_KEY] = 'Node-3.14.4';
+
+    // $FlowIgnore - stop complaining about property `env` not found
+    this._tags[constants.JAEGER_CLIENT_VERSION_TAG_KEY] = `Node-${process.env['PACKAGE_VERSION']}`;
```
yurishkuro: I don't understand this - who is going to provide this env variable when the tracer is used in the end-user's application?

yurishkuro: or is it going to be resolved at transpile-time?

everett980: babel-plugin-transform-inline-environment-variables changes the built file to be `'Node-' + '3.14.4'` (whatever the version is when babel transpiles). it is defined by the makefile.

everett980: I packed the module (`make node-build && yarn pack`)and installed the tar in other folder and it has the version at build time in the dist:
image

everett980:
@yurishkuro I'm not sure what crossdock is or why the build failed when the env variables had `CROSSDOCK=1`
Is it possible to trigger a rebuild since the build seems to be a timeout in travis which could be unrelated.

yurishkuro:
restarted build

yurishkuro:
@everett980 I think crossdock test failure is related to your change. Expand after_failure in Travis log and look for node_1 logs (may need to download and grep), I think it complains about this new Babel pkg.

.makenodebabelrc
```diff
@@ -0,0 +1,15 @@
...
+    [
+      "transform-inline-environment-variables",
+      {
+        "include": [
+          "PACKAGE_VERSION"
+        ]
+      }
+    ]
+  ]
+}
```
yurishkuro: can you elaborate why this cannot be the default .babelrc and you need to copy it back and forth?

everett980: @yurishkuro
When running tests in Node 0.10, the `babel-plugin-transform-inline-environment-variables` cannot be used because it is only compatible with Node 6+. Additionally, this rule cannot be defined as a cli argument because the `includes` option must be defined otherwise different environment variables are replaced with `undefined`.
Is there somewhere in the codebase I should add this comment?

yurishkuro: > When running tests in Node 0.10

Tests with 0.10 are run against transpiled code, at which point the replacement should have happened already.

everett980: The tests themselves also need to transpile at test execution time. This plugin doesn't work in 0.10 so it is used at build time (Node 6), and not used the rest of the time (variable node version).

yurishkuro: so why go through all these hoops then?  Your earliest solution of just writing `version.js` file during build was much simpler imo. Simple unix commands instead of plugin always get my vote.

everett980: I leaned towards the plugin to avoid making a file in `dist` off of a file in `src`, but since the file is phenomenally simple (export a single string) it does seem easier than maintaining two `.bablerc` files or figuring out why a `babel.config.js` file isn't working.

I reverted the branch back to the previous, simpler state.

PR #350, opened 2019-03-12
WIP: Allow configuring http client timeout

sabrehagen:

- Users are unable to configure http client timeout

- Added support for `JAEGER_REPORTER_TIMEOUT` environment variable. Updated docs. Added tests.

## Status

Not ready for merge. Awaiting `configuration_env.js` test.


src/reporters/http_sender.js
```diff
@@ -45,7 +45,7 @@ export default class HTTPSender {
     this._url = URL.parse(options.endpoint);
     this._username = options.username;
     this._password = options.password;
-    this._timeoutMS = options.timeoutMS || DEFAULT_TIMEOUT_MS;
+    this._timeoutMs = options.timeoutMs || DEFAULT_TIMEOUT_MS;
```
yurishkuro: let's keep this backwards compatible: `options.timeoutMs || options.timeoutMS || DEFAULT_TIMEOUT_MS;`

src/configuration_env.js
```diff
@@ -21,18 +21,20 @@ const deprecatedEnvVars = {
...
-        if (process.env[env]) {
-          console.warn(`You are using deprecated env variable ${env}. Use ${deprecatedEnvVars[env]} instead. \nDeprecated env variable will be removed in the next major release (4.x.x)`)
-        }
-      })
+    Object.keys(deprecatedEnvVars).forEach(env => {
+      if (process.env[env]) {
+        console.warn(
+          `You are using deprecated env variable ${env}. Use ${
+            deprecatedEnvVars[env]
+          } instead. \nDeprecated env variable will be removed in the next major release (4.x.x)`
```
yurishkuro: This feels like a very weird formatting of an extrapolated expression. Why do we need this change in this PR?

src/configuration.js
```diff
@@ -143,6 +144,10 @@ export default class Configuration {
       if (config.reporter.agentPort) {
         senderConfig['port'] = config.reporter.agentPort;
       }
+
+      if (config.reporter.timeout) {
```
yurishkuro: timeout or timeoutMs? Is there a test?

src/reporters/http_sender.js
sabrehagen: Sure, can do. `timeoutMS` was not referenced elsewhere in this codebase, so I felt safe changing it. Is it possible that it's used by code outside the codebase that passes `timeoutMS`?

yurishkuro: http sender is not hidden under some `internal` package, so technically it's a public API, and I'd prefer to not break anyone who may be wiring components manually.

src/configuration_env.js
sabrehagen: Sorry for the confusion, I should have commented this change. This is happening when `prettier` [lints](https://github.com/jaegertracing/jaeger-client-node/blob/master/package.json#L81) the staged files. How would you like this to be resolved? Alter prettier config or accept new styling?

src/reporters/http_sender.js
sabrehagen: Thanks for the info. Will update as requested.

src/configuration_env.js
```diff
@@ -185,9 +196,10 @@ export default class ConfigurationEnv {
    * @param {Object} options - options, see Configuration.initTracer
    */
   static initTracer(config = {}, options = {}) {
-    ConfigurationEnv._validateEnv()
+    ConfigurationEnv._validateEnv();
```
sabrehagen: These 3 lines are changed automatically by prettier on `git add`.

src/configuration.js
sabrehagen: Good catch. Only `configuration_env.js` was tested. Added a test for `configuration.js` [here](https://github.com/jaegertracing/jaeger-client-node/pull/350/files#diff-0563bbe36f9d317234592eb362aa4115R196).

src/configuration_env.js
sabrehagen: To have a clean diff for now, I have overridden prettier and reverted these changes. Awaiting your final advice on the issue.

yurishkuro:
```
1 failing
  1) initTracerFromENV should respect http reporter options from mismatching env:
     AssertionError: expected undefined to equal 225
      at Context.<anonymous> (dist/test/init_tracer.js:533:18)
```

yurishkuro:
replaced by #465

PR #351, opened 2019-03-15
Wait statusCode 2xx for reset the batch spans

sguilly:

- when the http sender fail to send the span batches, spans are lose because they have been clear without waiting the statusCode of the request

- Call reset only when we receive a 2xx status code

### DIFF
--- a/src/reporters/http_sender.js
+++ b/src/reporters/http_sender.js
@@ -111,7 +111,7 @@ export default class HTTPSender {
 
     const req = requester(this._httpOptions, resp => {
       
-      if(resp.statusCode && resp.statusCode === 200)
+      if(resp.statusCode && (resp.statusCode >= 200 || resp.statusCode < 227))
       {
         this._reset();  
       }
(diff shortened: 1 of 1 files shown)


src/reporters/http_sender.js
```diff
@@ -107,11 +107,15 @@ export default class HTTPSender {
...
-    this._reset();
-
     const requester = this._url.protocol === 'https:' ? https.request : http.request;
 
     const req = requester(this._httpOptions, resp => {
+      
+      if(resp.statusCode && (resp.statusCode >= 200 || resp.statusCode < 227))
+      {
+        this._reset();  
+      }
```
yurishkuro: I dont think this is correct behavior. First, if we dont clear the buffer then the same spans might be submitted again due to concurrency if this request gets stuck. Second, if you want to build in error handling then the code can not proceed as usual and report success metrics.

yurishkuro:
No follow-up from the author, closing.

PR #354, opened 2019-04-01
- sockets fix

yrsh:
PR solves this issue:
https://github.com/jaegertracing/jaeger-client-node/issues/353

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "jaeger-client",
-  "version": "3.14.4",
+  "version": "3.14.5",
   "description": "Jaeger binding for OpenTracing Node",
   "engines": {
     "node": ">=0.10"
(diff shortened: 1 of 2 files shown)


yrsh:
@yurishkuro In fact, now sockets are just hanging in open state, and every next request opens new socket. please take a look at this issue https://github.com/jaegertracing/jaeger-client-node/issues/353, the issue is documented in details.

esimons:
We're also seeing this issue. Each span sent over HTTP appears to leave a connection open, eventually (in our case) leading to ECONNRESET due to an Istio circuit breaking on max connections.

Dadabhvai-Kashyap:
Why the socket not close does, its affect my application say that the pool connection have exceeded the limit server, I can connect not a new socket. Istio not do a connect to server, numbers of socket limit

esimons:
Alternate implementation offered in #356

isbalashov:
Is there any chance this PR could be merged eventually? This gives me lots of troubles now, had to fork library in order to continue using jaeger

yurishkuro:
there's more discussion happening in #356, and that fix looks safer than this PR, so I am going to close this, please watch the other one.

Issue #355, opened 2019-04-02
Add a MAINTAINERS/CODEOWNERS file


jpkrohling:
See jaegertracing/jaeger#254

Issue #344, opened 2019-02-11
Memory leak in HTTP sender


baldmaster:
After some time of using jaeger in production a memory leak detected thanks to [heapdump](https://github.com/bnoordhuis/node-heapdump). And here is why.

According to official [documentation](https://nodejs.org/dist/latest-v10.x/docs/api/http.html#http_class_http_clientrequest)

> If no 'response' handler is added, then the response will be entirely discarded. However, **if a 'response' event handler is added, then the data from the response object must be consumed**, either by calling response.read() whenever there is a 'readable' event, or by adding a 'data' handler, or by calling the .resume() method. Until the data is consumed, the 'end' event will not fire. Also, **until the data is read it will consume memory that can eventually lead to a 'process out of memory' error.** 
> 
[furthermore](https://nodejs.org/dist/latest-v10.x/docs/api/http.html#http_http_request_url_options_callback)

> The optional callback parameter will be added as a one-time listener for the **'response' event**. 
> 

It's pretty easy to reproduce. Start up any server, set HTTP jaeger endpoint and make a batch of requests.

before fix
![11-02-2019-19 09 17_1920x1080](https://user-images.githubusercontent.com/6952333/52580856-e6aec780-2e31-11e9-823e-f1fd193a105f.png)

after fix
![11-02-2019-19 10 07_1920x1080](https://user-images.githubusercontent.com/6952333/52582115-1c08e480-2e35-11e9-8d05-26ba81bfaba2.png)

comparison
![11-02-2019-19 10 40_1920x1080](https://user-images.githubusercontent.com/6952333/52582341-a5b8b200-2e35-11e9-91b1-2216aeb0b87a.png)

baldmaster:
Fixed by #343

viktorstaikov:
Is this fix published?

tiffon:
@viktorstaikov No, we just realized this fix fell through the cracks! We will be publishing an update, soon.

PR #356, opened 2019-04-02
Abort request on socket timeout

### LINKED ISSUE #353: HTTP transport problem
## Software versions
For our tests we used:
* jaeger-client-node: v3.12.x - v3.14.4
* jaeger-collector: v1-9 - v1.11

Client library doesn't close sockets, thus with a big number of request parallyzing network communication leading to **EADDRNOTAVAIL** error

```
Failed to flush spans in reporter: error sending spans over HTTP: Error: connect EADDRNOTAVAIL 172.30.20.40:14268 - Local (10.110.2.33:0)
error sending spans over HTTP: Error: connect EADDRNOTAVAIL 172.30.20.40:14268 - Local (10.130.2.33:0)
Failed to flush spans in reporter: error sending spans over HTTP: Error: connect EADDRNOTAVAIL 172.30.20.40:14268 - Local (10.110.2.33:0)
error sending spans over HTTP: Error: connect EADDRNOTAVAIL 172.30.20.40:14268 - Local (10.130.2.33:0)
Failed to flush spans in reporter: error sending spans over HTTP: Error: connect EADDRNOTAVAIL 172.30.20.40:14268 - Local (10.130.2.33:0)
error sending spans over HTTP: Error: connect EADDRNOTAVAIL 172.30.20.40:14268 - Local (10.130.2.33:0)
```

**HTTP transport problem**
When http used as main ransport, jaeger-client-node doesn't cleanup sockets,
with lots of  requests (e.g 100k) Linux faces socket shortage, number of opensockets increases up to max limit of 65k and that parallazyes network communication

image

To prevent socket shortage for entire hosts, max socket limit for http_sender could be introduced,
but this has a downside: sender quickly reaches limit and stops sending traces.

esimons:
Signed-off-by: Evan Simons <esimons@ebsco.com>

Resolves #353

Manually abort request (closing socket) when timeout event is received. See: https://nodejs.org/api/http.html#http_event_timeout
> Emitted when the underlying socket times out from inactivity. This only notifies that the socket has been idle. **The request must be aborted manually.**

Tested this manually in a consuming service, however, and observing that the problematic behavior is resolved. Apologies there's no new unit test for this, wasn't sure quite how to implement a relevant test case; would be willing to implement with some guidance, but hoping this can be merged expeditiously if deemed valuable.

I don't believe this is a complete solution -- I think it's still creating a connection per span reported over HTTP -- but it at least handles the timeout event and closes the socket, rather than leaking.

### DIFF
--- a/src/reporters/http_sender.js
+++ b/src/reporters/http_sender.js
@@ -122,6 +122,9 @@ export default class HTTPSender {
       this._logger.error(error);
       SenderUtils.invokeCallback(callback, numSpans, error);
     });
+    req.on('timeout', () => {
+      req.abort();
+    });
     req.write(result.value);
     req.end();
   }


esimons:
@yurishkuro wondering if I can get your eyes on this (very small) PR as it's fairly important to my team. Sorry to ping directly, as #355 points out there's no CODEOWNERS, so not sure if you're getting notifications for review requests.

src/reporters/http_sender.js
```diff
@@ -122,6 +122,9 @@ export default class HTTPSender {
       this._logger.error(error);
       SenderUtils.invokeCallback(callback, numSpans, error);
     });
+    req.on('timeout', () => {
+      req.abort();
```
yurishkuro: What happens to on-error path in this case? Is it still executed? If not, we need to execute the same logic

esimons: An error event should be emitted following the timeout event (error in this case would be "socket hang up").

yrsh: Hmm... we don't need wait timeout event, because the data has already been sent and the socket will not be reused - so we can abort connection, as soon as we get the response.

yurishkuro: @yrsh why abort the connection after every send? I'd think it's better to keep it open and reuse for future sends.

tiffon:
Looking through `http(s?).Agent` docs ([0.10.x](https://nodejs.org/docs/latest-v0.10.x/api/http.html), [10.x](https://nodejs.org/dist/latest-v10.x/docs/api/http.html)), seems like we might also make some headway by adjusting the agent config.

>`maxSockets` `<number>` Maximum number of sockets to allow per host. Each request will use a new socket until the maximum is reached. Default: Infinity.
>
> [Node 10.x docs](https://nodejs.org/dist/latest-v10.x/docs/api/http.html#http_new_agent_options)

In older versions of node (e.g. 0.10.x), the max number of sockets was set to 5 ([ref](https://nodejs.org/docs/latest-v0.10.x/api/http.html#http_agent_maxsockets)), which is infinity in newer versions, as quoted above.

Sounds like leaving this at infinity and using `{ keepAlive: true }` might be part of the issue. Have you tried setting this to something sane?

esimons:
@tiffon that sounds reasonable to me, would 5 constitute a sane default for jaeger-client? Can add that change to the PR, but not familiar with the expected characteristics under load.

src/reporters/http_sender.js
yrsh: @yurishkuro because it will not be reused in current case, so it’s better to close connections than to fail with socket limit error.

yrsh:
@tiffon I checked this scenario, it is not working. Because free sockets end, current open sockets hangs in open state, so new connections do not create and the traces stop sending.

isbalashov:
If you are not convinced please look at picture below

image

esimons:
@tiffon just tested locally with `maxSockets` set to 5, and the agent does indeed create connections up until 5 at which point no new connections are created... however, I'm observing the same thing @yrsh points out, that the agent fails to push any additional spans once it hits that connection limit (seems to fail silently for me).
@yurishkuro It's still not clear to me what's wrong with the usage of http.agent here that it's not reusing connections, and instead allowing them to pile up in ESTABLISHED state. It pains me to not know but I'm now tapped out on time for analysis, so if this is low-risk I'd like to ask if this PR could be merged as-is, as this issue is causing significant pain for our team.

tiffon:
Thanks for your time and attention to this issue!

Thanks also for trying out the suggestion of setting maxSockets.

IIRC, this PR was discussed in the bi-weekly, yesterday, so I'll defer to @yurishkuro but want to share a couple of thoughts.

From what I read, it seems like this package might be a reliable solution: [agentkeepalive](https://github.com/node-modules/agentkeepalive). Because we need to support node 0.10, we might want to look into the [2.x](https://github.com/node-modules/agentkeepalive/tree/2.x) version.

Regarding using the builtin `http.Agent`, are sockets being reused, at all?

If they are being reused but they end up in a bad state, maybe the issue is with the `_timeoutMs`? The timeout gets passed on to the socket, and if it is not being removed after a request finishes, maybe it's firing after being idle even when there isn't a request being processed? (This would only matter if a socket firing the `timeout` event actually puts the socket into a bad state, which seems unlikely.) In this case, removing the timeout after a request is processed might help.

```js
const req = requester(this._httpOptions, resp => {
  // consume response data to free up memory
  resp.resume();
  resp.end(() => {
  	req.socket.setTimeout(0);
  })
  SenderUtils.invokeCallback(callback, numSpans);
});
```

If sockets are not being reused then `keepAlive: true` is actually problematic (apparently) and removing it will probably help.

esimons:
> Regarding using the builtin http.Agent, are sockets being reused, at all?

In my observation they do not appear to be reused at all.

src/reporters/http_sender.js
yurishkuro: clarification from last project call - "abort connection" above is actually "abort current request",  which makes sense. Without aborting the request the connection is kept open.

yurishkuro:
> > Regarding using the builtin http.Agent, are sockets being reused, at all?

> In my observation they do not appear to be reused at all.

Not reused in master, or with the change proposed here to abort the request?

Isn't there a canonical way to use `http.request` that doesn't leak connection?

src/reporters/http_sender.js
tiffon: This will abort the request and destroy the connection.

https://nodejs.org/api/http.html#http_request_abort

tiffon:
Can you provide guidance on how to reproduce and measure this, locally?

esimons:
> Not reused in master, or with the change proposed here to abort the request?

Both, as the change made in this PR doesn't have any impact on that aspect. I was hoping only to contain the leak with a minimal, low-impact change that adds cleanup of idle sockets (which I imagine you'd want in any case as a matter of etiquette).

> Isn't there a canonical way to use `http.request` that doesn't leak connection?

Presumably! But I have to admit it's not an area of experience for me, and the fact that [agentkeepalive](https://www.npmjs.com/package/agentkeepalive) has been netting ~1.5m downloads a week indicates that maybe there's something to be desired in the base Node API. :smile:
There's also some potentially complicating network factors (Envoy sidecar proxy, for instance) that could be having impact.

> Can you provide guidance on how to reproduce and measure this, locally?

I was just using netstat to monitor connection build-up. If I find some time I might try setting up a demonstration repo, but honestly not sure when that might be.

tiffon:
@esimons @isbalashov After a few iterations and some banging my head against the wall, I've concluded the issue is caused by [#344](https://github.com/jaegertracing/jaeger-client-node/issues/344).

The only way I was able to reliably reproduce the issue is by removing the [`resp.resume();`](https://github.com/jaegertracing/jaeger-client-node/blob/a12692c2d5f9d19379ec63756096de1624816814/src/reporters/http_sender.js#L115-L116).

And, :fire:very unfortunately:fire:, this fix is on master but has not yet been released.

The proposed fixes of `req.abort()` do circumvent the issue, but the `resp.resume()` is the underlying fix.

We will cut a release, but I wanted to see if you can confirm this addresses the issue?

@esimons Thanks for the tip for reproducing this.

esimons:
@tiffon thanks for the follow-up. Ran a very small-scale manual test now and the issue does appear to be resolved locally if the current state of master were to be released.

tiffon:
@esimons Thank you for verifying! I will release the client tomorrow.

tiffon:
@esimons, @isbalashov, @yrsh [Version 3.15.0](https://github.com/jaegertracing/jaeger-client-node/releases/tag/v3.15.0) of the node [jaeger-client](https://www.npmjs.com/package/jaeger-client) has just been published. This should resolve the issues observed in this PR.

Please let us know if anything looks awry!

tiffon:
Closing as this should be resolved with the recent release.

Issue #143, opened 2017-07-31
Check the status code before parsing the agent response


black-adder:
https://github.com/uber/jaeger-client-node/blob/master/src/samplers/remote_sampler.js#L127

To return a more comprehensible log: https://github.com/uber/jaeger-client-node/issues/141#issuecomment-318661353

PR #360, opened 2019-04-22
check statuscode before firing success callback

guanw:
Signed-off-by: Jude Wang <judew@uber.com>

- #143 

- Validate statusCode before firing success callbacks for throttler and adaptive sampling

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -132,7 +132,6 @@ export default class Utils {
         res => {
           // explicitly treat incoming data as utf8 (avoids issues with multi-byte chars)
           res.setEncoding('utf8');
-          let statusCode = res.statusCode;
           // incrementally capture the incoming response body
           let body = '';
           res.on('data', chunk => {
@@ -140,10 +139,10 @@ export default class Utils {
           });
 
           res.on('end', () => {
-            if (statusCode == 200) {
+            if (res.complete && res.statusCode == 200) {
               success(body);
             } else {
-              error('StatusCode: ', statusCode);
+              error(new Error(`HTTP GET failed: ${res.statusCode}`));
             }
           });
         }
--- a/src/util.js
+++ b/src/util.js
@@ -139,7 +139,7 @@ export default class Utils {
           });
 
           res.on('end', () => {
-            if (res.complete && res.statusCode == 200) {
+            if (res.statusCode == 200) {
               success(body);
             } else {
               error(new Error(`HTTP GET failed: ${res.statusCode}`));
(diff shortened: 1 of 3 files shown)


src/util.js
```diff
@@ -132,15 +132,19 @@ export default class Utils {
...
+          let statusCode = res.statusCode;
           // incrementally capture the incoming response body
           let body = '';
           res.on('data', chunk => {
             body += chunk;
           });
 
           res.on('end', () => {
-            success(body);
+            if (statusCode == 200) {
```
tiffon: Why use a local var instead of `res.statusCode`?

tiffon: You may also want to check [`res.complete`](https://nodejs.org/api/http.html#http_message_complete).

tiffon: Also, because the `error` callback is called from the [stream `'error'` event](https://nodejs.org/api/stream.html#stream_event_error_1), it seems like it's usually called with an `Error` object.

So, it might make sense to generate pass an error into the callback rather than a string.

```js
const err = new Error(`HTTP GET failed: ${res.statusMessage}`);
err.code = res.statusMessage;
err.response = res;
error(err);
```

Or maybe something simpler:

```js
error(new Error(`HTTP GET failed: ${res.statusCode}`));
```

guanw: Updated

yurishkuro:
can you add a test?

guanw:
> can you add a test?

We have test already in sampler and throttler that calls out to `Utils.httpGet` in different scenarios, could you elaborate?

yurishkuro:
You're making a semantic change where depending on the status code the callback to success() can be replaced with a call to error(). Yet no test changes are included to reflect this semantic change, which makes me think that even if the tests do exist they do not set correct expectations or not checking which callback is called.

guanw:
I've thought about it some more. Maybe we should just close this request. I don't find the original error message `Error in parsing sampling strategy` to be annoying as that's what happened when agent/collector is down. I think we need to make sure all client acting consistently instead.

yurishkuro:
Well, I think it's a good change, because not checking response for status code looks very much like a bug. I just want to have unit tests that simulate that condition so that the same bug doesn't creep in again.

guanw:
> Well, I think it's a good change, because not checking response for status code looks very much like a bug. I just want to have unit tests that simulate that condition so that the same bug doesn't creep in again.

Gotcha. @tiffon Do you know if we have a good way to mimic 5xx error for node client? I was thinking just return 500 error for certain service name but am not sure this is the best way to do it.

image

yurishkuro:
if you're running a real server, no need to mimic 5xx, just return them as in your snippet.

yurishkuro:
already fixed in another PR

PR #361, opened 2019-04-26
Adding 128 bit traceId support

### LINKED ISSUE #295: SpanContext cannot handle 128-bit traceId

We have a mix of tracing instrumentation libraries in our environment, some of which generate 128-bit trace ids. Those trace ids cannot be handled by the jaeger-client-node library and we end up with disconnected traces.

The SpanContext class is attempting to parse the traceId as a 64-bit number, which means the 128-bit number is not handled properly and the trace ends up disconnected.

https://github.com/jaegertracing/jaeger-client-node/blob/master/src/span_context.js#L69-L74

The trace id is a stored as a buffer, can we just use a 128-bit buffer? I'm not sure the ramifications of that though.

PaulMiami:
Signed-off-by: Paul Biccherai <pbiccherai@hotmail.com>

Resolves #295

Adding 128 bit support to the traceId

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,8 @@ matrix:
       - docker
   - env: TEST_NODE_VERSION=6 LINT=1 COVER=1
   - env: TEST_NODE_VERSION=4
+  - env: TEST_NODE_VERSION=--lts
+  allow_failures:
   - env: TEST_NODE_VERSION=node
 
 env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 sudo: required
+dist: trusty
 
 # Travis will use .nvmrc, by default
 language: node_js
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,7 @@
 sudo: required
 
+# Travis will use .nvmrc, by default
 language: node_js
-
-node_js:
-- '6'
 
 cache:
   directories:
@@ -19,12 +17,9 @@ matrix:
     sudo: required
     services:
       - docker
-  - env: TEST_NODE_VERSION=6 LINT=1 COVER=1
   - env: TEST_NODE_VERSION=4
-  - env: TEST_NODE_VERSION=--lts
-  - env: TEST_NODE_VERSION=node
-  allow_failures:
-  - env: TEST_NODE_VERSION=node
+  - env: TEST_NODE_VERSION=6
+  - env: TEST_NODE_VERSION=10 LINT=1 COVER=1
 
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,7 @@ matrix:
   - env: TEST_NODE_VERSION=6 LINT=1 COVER=1
   - env: TEST_NODE_VERSION=4
   - env: TEST_NODE_VERSION=--lts
+  - env: TEST_NODE_VERSION=node
   allow_failures:
   - env: TEST_NODE_VERSION=node
 
--- a/src/_flow/tracer.js
+++ b/src/_flow/tracer.js
@@ -20,4 +20,5 @@ declare type startSpanOptions = {
   references?: Array<Reference>,
   tags?: any,
   startTime?: number,
+  traceId128bit?: boolean,
 };
--- a/src/_flow/tracer.js
+++ b/src/_flow/tracer.js
@@ -20,5 +20,4 @@ declare type startSpanOptions = {
   references?: Array<Reference>,
   tags?: any,
   startTime?: number,
-  traceId128bit?: boolean,
 };
--- a/src/span_context.js
+++ b/src/span_context.js
@@ -41,7 +41,7 @@ export default class SpanContext {
   ) {
     this._traceId = traceId;
     this._spanId = spanId;
-    this._parentId = parentId;
+    this._parentId = parentId || null;
     this._traceIdStr = traceIdStr;
     this._spanIdStr = spanIdStr;
     this._parentIdStr = parentIdStr;
--- a/src/span_context.js
+++ b/src/span_context.js
@@ -68,8 +68,10 @@ export default class SpanContext {
 
   get traceId(): any {
     if (this._traceId == null && this._traceIdStr != null) {
+      // make sure that the HEX has an even number of digits, node is expecting 2 HEX character per byte
+      // https://github.com/nodejs/node/issues/21242
       const safeTraceIdStr = this._traceIdStr.length % 2 == 0 ? this._traceIdStr : '0' + this._traceIdStr;
-      const tmpBuffer = Utils.newBuffer(safeTraceIdStr, 'hex');
+      const tmpBuffer = Utils.newBufferFromHex(safeTraceIdStr);
       const size = tmpBuffer.length > 8 ? 16 : 8;
       this._traceId = Utils.newBuffer(size);
       tmpBuffer.copy(this._traceId, size - tmpBuffer.length);
--- a/src/span_context.js
+++ b/src/span_context.js
@@ -15,8 +15,7 @@ import * as constants from './constants.js';
 import Utils from './util.js';
 
 export default class SpanContext {
-  _traceIdLow: any;
-  _traceIdHigh: any;
+  _traceId: any;
   _spanId: any;
   _parentId: any;
   _traceIdStr: ?string;
@@ -44,8 +43,7 @@ export default class SpanContext {
   _samplingFinalized: boolean;
 
   constructor(
-    traceIdLow: any,
-    traceIdHigh: any,
+    traceId: any,
     spanId: any,
     parentId: any,
     traceIdStr: ?string,
@@ -56,8 +54,7 @@ export default class SpanContext {
     debugId: ?string = '',
     samplingFinalized: boolean = false
   ) {
-    this._traceIdLow = traceIdLow;
-    this._traceIdHigh = traceIdHigh;
+    this._traceId = traceId;
     this._spanId = spanId;
     this._parentId = parentId;
     this._traceIdStr = traceIdStr;
@@ -69,33 +66,16 @@ export default class SpanContext {
     this._samplingFinalized = samplingFinalized;
   }
 
-  get traceIdLow(): any {
-    this._tryParseTraceIdStr();
-    return this._traceIdLow;
-  }
-
-  get traceIdHigh(): any {
-    this._tryParseTraceIdStr();
-    return this._traceIdHigh;
-  }
-
-  _tryParseTraceIdStr() {
-    if (this._traceIdLow == null && this._traceIdHigh == null && this._traceIdStr != null) {
-      if (this._traceIdStr.length > 16) {
-        const safeTraceIdStr = this._traceIdStr.length % 2 == 0 ? this._traceIdStr : '0' + this._traceIdStr;
-        const traceId = new Buffer(safeTraceIdStr, 'hex');
-        this._traceIdHigh = traceId.slice(-16, -8);
-        if (this._traceIdHigh.length != 8) {
-          const tmpBuffer = new Buffer(8);
-          tmpBuffer.fill(0);
-          this._traceIdHigh.copy(tmpBuffer, 8 - this._traceIdHigh.length);
-          this._traceIdHigh = tmpBuffer;
-        }
-        this._traceIdLow = traceId.slice(-8);
-      } else {
-        this._traceIdLow = Utils.encodeInt64(this._traceIdStr);
-      }
+  get traceId(): any {
+    if (this._traceId == null && this._traceIdStr != null) {
+      const safeTraceIdStr = this._traceIdStr.length % 2 == 0 ? this._traceIdStr : '0' + this._traceIdStr;
+      const tmpBuffer = new Buffer(safeTraceIdStr, 'hex');
+      const size = tmpBuffer.length > 8 ? 16 : 8;
+      this._traceId = new Buffer(size);
+      this._traceId.fill(0);
+      tmpBuffer.copy(this._traceId, size - tmpBuffer.length);
     }
+    return this._traceId;
   }
 
   get spanId(): any {
@@ -113,14 +93,8 @@ export default class SpanContext {
   }
 
   get traceIdStr(): ?string {
-    if (this._traceIdStr == null && this._traceIdLow != null) {
-      if (this._traceIdHigh != null) {
-        this._traceIdStr = Utils.removeLeadingZeros(
-          this._traceIdHigh.toString('hex') + this._traceIdLow.toString('hex')
-        );
-      } else {
-        this._traceIdStr = Utils.removeLeadingZeros(this._traceIdLow.toString('hex'));
-      }
+    if (this._traceIdStr == null && this._traceId != null) {
+      this._traceIdStr = Utils.removeLeadingZeros(this._traceId.toString('hex'));
     }
     return this._traceIdStr;
   }
@@ -155,13 +129,8 @@ export default class SpanContext {
     return this._samplingFinalized;
   }
 
-  set traceIdLow(traceIdLow: Buffer): void {
-    this._traceIdLow = traceIdLow;
-    this._traceIdStr = null;
-  }
-
-  set traceIdHigh(traceIdHigh: Buffer): void {
-    this._traceIdHigh = traceIdHigh;
+  set traceId(traceId: Buffer): void {
+    this._traceId = traceId;
     this._traceIdStr = null;
   }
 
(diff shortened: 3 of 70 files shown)


mwieczorek:
@PaulMiami Any news on resolving the Travis errors?
Sorry for bothering but I'm also interested in this feature.

PaulMiami:
> @PaulMiami Any news on resolving the Travis errors?
> Sorry for bothering but I'm also interested in this feature.

@mwieczorek I am not sure if I can fix that one on my own, travis only fails trying to run the tests with node 12. I haven't looked at it extensively but it seems that it won't work until node 12 is more stable.

PaulMiami:
I am not sure if it's acceptable but I have changed the travis config to build node lts instead of latest, which is unstable at this time

.travis.yml
```diff
@@ -21,6 +21,8 @@ matrix:
       - docker
   - env: TEST_NODE_VERSION=6 LINT=1 COVER=1
   - env: TEST_NODE_VERSION=4
+  - env: TEST_NODE_VERSION=--lts
+  allow_failures:
```
yurishkuro: what is this for, and it is indented correctly?

PaulMiami: The latest version of node seems to be unstable at this time, so I added the lts version to the build and move the latest to the 'allow_failures' sections which will still build it but won't fail the entire build if it fails

src/span_context.js
```diff
@@ -124,8 +155,13 @@ export default class SpanContext {
     return this._samplingFinalized;
   }
 
-  set traceId(traceId: Buffer): void {
-    this._traceId = traceId;
+  set traceIdLow(traceIdLow: Buffer): void {
```
yurishkuro: why do we need to expose low/high fields explicitly?  It seems like the Buffer type can represent the full ID as a single buffer, and do it transparently to the user.

.travis.yml
yurishkuro: makes total sense, thanks

src/span_context.js
PaulMiami: ok, I'll work on that

mwieczorek:
Hi, I'm sorry for bothering once again.
But I guess @PaulMiami addressed comments from review. 
@yurishkuro - could you take a look?

scott2449:
This update is critical to our usage w/ Istio @dowjones

src/tracer.js
```diff
@@ -188,6 +188,8 @@ export default class Tracer {
    *        the created Span object. The time should be specified in
    *        milliseconds as Unix timestamp. Decimal value are supported
    *        to represent time values with sub-millisecond accuracy.
+   * @param {number} [options.traceId128bit] - generate root span with a
```
everett980: I believe this is a boolean based on the config and a test, so it should be `@param {boolean} ...` instead

src/span_context.js
```diff
@@ -68,7 +68,12 @@ export default class SpanContext {
 
   get traceId(): any {
     if (this._traceId == null && this._traceIdStr != null) {
-      this._traceId = Utils.encodeInt64(this._traceIdStr);
+      const safeTraceIdStr = this._traceIdStr.length % 2 == 0 ? this._traceIdStr : '0' + this._traceIdStr;
+      const tmpBuffer = new Buffer(safeTraceIdStr, 'hex');
+      const size = tmpBuffer.length > 8 ? 16 : 8;
+      this._traceId = new Buffer(size);
```
everett980: It seems as `new Buffer` [is deprecated](https://nodejs.org/api/buffer.html#buffer_new_buffer_array). We should probably use the new API. So this line and the subsequent line can jointed into:
```
this._traceId = Buffer.alloc(size)
```
Then we can rely on babel to provide backwards compatibility.

src/tracer.js
```diff
@@ -239,7 +241,11 @@ export default class Tracer {
         ctx.baggage = parent.baggage;
       }
 
-      ctx.traceId = randomId;
+      if (options.traceId128bit) {
+        ctx.traceId = new Buffer.concat([Utils.getRandom64(), randomId]);
```
everett980: I believe you can just drop the `new` to avoid the deprecated constructor

test/span_context.js
```diff
@@ -90,6 +106,26 @@ describe('SpanContext should', () => {
...
+    assert.equal(null, context.parentId);
+    assert.equal(1, context.flags);
+
+    // test large numbers
+    context = SpanContext.fromString('ffffffffffffffffffffffffffffffff:ffffffffffffffff:5:1');
+    assert.equal('ffffffffffffffffffffffffffffffff', context.traceIdStr);
+    assert.equal('ffffffffffffffff', context.spanIdStr);
+    assert.deepEqual(Buffer.concat([LARGEST_64_BUFFER, LARGEST_64_BUFFER]), context.traceId);
+    assert.deepEqual(LARGEST_64_BUFFER, context.spanId);
+    assert.deepEqual(LARGEST_64_BUFFER, context.spanId);
```
everett980: This line is the same as the previous line.

src/span_context.js
```diff
@@ -68,7 +68,12 @@ export default class SpanContext {
 
   get traceId(): any {
     if (this._traceId == null && this._traceIdStr != null) {
-      this._traceId = Utils.encodeInt64(this._traceIdStr);
+      const safeTraceIdStr = this._traceIdStr.length % 2 == 0 ? this._traceIdStr : '0' + this._traceIdStr;
+      const tmpBuffer = new Buffer(safeTraceIdStr, 'hex');
```
everett980: This should be [`Buffer.from(safeTraceIdStr, 'hex');`](https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_string_encoding)

src/span_context.js
tiffon: @everett980 I don't think babel will provide backwards compatibility for the `Buffer.from` because it's not a syntax issue.

src/tracer.js
tiffon: In this case I think the use of `new` is unnecessary, regardless of the constructor being deprecated.

PaulMiami:
No problem!
Check out the changes and if you could please re-run the build because it failed with a socket error.
Thanks

tiffon:
Thanks for the changes!

The build for 0.10 keeps hitting a [test failure](https://travis-ci.org/jaegertracing/jaeger-client-node/jobs/544988472#L1555-L1561). A test which covers an error scenario is failing due to an emitted error having a different underlying OS error: `ESRCH` instead of `ENOTFOUND`.

https://github.com/jaegertracing/jaeger-client-node/blob/8d0b1f766846b80b397f1ca2f809e070fc20bba2/test/http_sender.js#L241-L256

```
  1) http sender should gracefully handle errors emitted by socket.send:
     Uncaught AssertionError: expected 'error sending spans over HTTP: Error: getaddrinfo ESRCH' to contain 'error sending spans over HTTP: Error: getaddrinfo ENOTFOUND'
      at dist/test/http_sender.js:305:38
      at Function.invokeCallback (dist/src/reporters/sender_utils.js:33:9)
      at ClientRequest.<anonymous> (dist/src/reporters/http_sender.js:148:32)
      at Socket.socketErrorListener (http.js:1610:9)
      at net.js:834:16
```

This is passing on [master](https://travis-ci.org/jaegertracing/jaeger-client-node/jobs/530913640#L1630) (I reran it Friday), so it seems like it's related to the diff.

I think revising the `expect(...)` to be more generic would entail a breaking change (via a different error message) and expose folks to different error messages based on the version of node they're running. 

Do you know what the issue might be?

@yurishkuro Any thoughts on what the issue might be?

PaulMiami:
@tiffon apparently travis is [defaulting to ubuntu xenial now ](https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment) and master is building of trusty, I changed the travis file to force trusty.

tiffon:
@PaulMiami 	Thanks!

src/_flow/tracer.js
```diff
@@ -20,4 +20,5 @@ declare type startSpanOptions = {
   references?: Array<Reference>,
   tags?: any,
   startTime?: number,
+  traceId128bit?: boolean,
```
yurishkuro: why is this needed? `startSpan()` is an OpenTracing API function, we cannot change its signature in Jaeger. This flag is only needed as tracer-level flag, not span-level.

src/span_context.js
```diff
@@ -68,7 +68,11 @@ export default class SpanContext {
 
   get traceId(): any {
     if (this._traceId == null && this._traceIdStr != null) {
-      this._traceId = Utils.encodeInt64(this._traceIdStr);
+      const safeTraceIdStr = this._traceIdStr.length % 2 == 0 ? this._traceIdStr : '0' + this._traceIdStr;
```
yurishkuro: What is the objective of checking for even string length? Please add a comment when this could be an issue.

src/span_context.js
```diff
@@ -68,7 +68,11 @@ export default class SpanContext {
...
+      const tmpBuffer = Utils.newBuffer(safeTraceIdStr, 'hex');
+      const size = tmpBuffer.length > 8 ? 16 : 8;
+      this._traceId = Utils.newBuffer(size);
```
yurishkuro: why do we need to allocate two buffers?

src/tracer.js
```diff
@@ -239,7 +241,11 @@ export default class Tracer {
         ctx.baggage = parent.baggage;
       }
 
-      ctx.traceId = randomId;
+      if (options.traceId128bit) {
+        ctx.traceId = Buffer.concat([Utils.getRandom64(), randomId]);
```
yurishkuro: this can probably made more efficient by allocating 16-bytes buffer, copying randomId, and filling high bytes with more random values. Saves at least one allocation.

src/util.js
```diff
@@ -16,6 +16,17 @@ import Int64 from 'node-int64';
...
+  | 'ascii'
+  | 'utf8'
+  | 'utf-8'
+  | 'utf16le'
+  | 'ucs2'
+  | 'ucs-2'
+  | 'base64'
+  | 'latin1'
+  | 'binary'
+  | 'hex';
```
yurishkuro: why do we need all of these?

src/util.js
```diff
@@ -148,4 +159,28 @@ export default class Utils {
         error(err);
       });
   }
+
+  /**
+   * @param {string|number} input - a string to store in the buffer
```
yurishkuro: There are no common code paths based on this param, and `encoding` is only used for string. I think this should be split into separate functions.

test/tracer.js
```diff
@@ -418,6 +450,17 @@ describe('tracer should', () => {
     });
   });
 
+  it('start a root span with 128 bit traceId', () => {
+    let span = tracer.startSpan('test-name', {
+      traceId128bit: true,
+    });
```
yurishkuro: this is not OpenTracing-compliant code.

src/span_context.js
PaulMiami: In node hex strings have to be of [even length](https://github.com/nodejs/node/issues/21242) for Buffer.From(), I'll add a comment

src/_flow/tracer.js
PaulMiami: didn't know that, I'll move it

src/span_context.js
PaulMiami: I need to create a 64 or 128 bits buffer and decode the hex string at the same, I don't think that can be done in one shot

test/tracer.js
PaulMiami: sorry didn't know that

src/util.js
PaulMiami: I'll split the code

PaulMiami:
@tiffon I made the changes. The code coverage went down a little because the deprecated buffer code does get hit

https://github.com/PaulMiami/jaeger-client-node/blob/a00710ef336e049a85daf0f9861b854cfa1c37c1/src/util.js#L178

https://github.com/PaulMiami/jaeger-client-node/blob/a00710ef336e049a85daf0f9861b854cfa1c37c1/src/util.js#L189-L191

src/util.js
```diff
@@ -148,4 +165,29 @@ export default class Utils {
...
       });
   }
+
+  /**
+   * @param {string|number} input - a hex encoded string to store in the buffer.
+   * @return {Buffer} - returns a buffer representing the hex encoded string.
+   **/
+  static newBufferFromHex(input: string): Buffer {
+    const encoding = 'hex';
+    if (Buffer.from && Buffer.from !== Uint8Array.from) {
```
yurishkuro: please add a comment about this check, the purpose is not clear

src/span_context.js
yurishkuro: It was unfortunate oversight that Jaeger clients allowed shortened string representation of trace/span IDs by dropping leading zeroes (an optimization that only affects ~1/16th of all traces, wasn't worth it), instead of always insisting on exactly 16 or 32 characters. I booked a meta-issue for this: https://github.com/jaegertracing/jaeger/issues/1657

To maintain backwards compatibility we do need to be able to accept strings <16 or <32 chars, but always output strings of the exact length.

The overall implementation would probably be simpler if internal representation always uses a buffer of exact length, as you're doing above (an alternative would be use use the buffer of whatever length the input string is). But we should still avoid double-allocating the buffer. For example, in L73 instead of padding  with a single zero, you could pad with as many zeros as needed to reach 16/32 length, and then your `Utils.newBufferFromHex()` function will return the right-sized buffer in one allocation. We still lose an allocation due to padding, but that will only affect about 1/16th of all traces, and once https://github.com/jaegertracing/jaeger/issues/1657 is fixed for all languages padding won't be happening in practice.

PaulMiami:
@yurishkuro I made the changes but I am trying to figure out what's wrong with the build
```
  1) http sender should gracefully handle errors emitted by socket.send:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
  
  2) udp sender should gracefully handle errors emitted by socket.send:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

yurishkuro:
I am not sure. Are they repeatable locally, or just in a one-off Travis run?

PaulMiami:
@yurishkuro it didn't fail initially on my local but after running it repeatedly it did once in a while.
I started to run mocha only for 'udp sender should gracefully handle errors emitted by socket.send' with no timeout and once in a while the time took over 2 seconds to complete.
```
...
✓ should gracefully handle errors emitted by socket.send
1 passing (39ms)
✓ should gracefully handle errors emitted by socket.send
1 passing (37ms)
✓ should gracefully handle errors emitted by socket.send (3020ms)
1 passing (3s)
```
I have increase the timeout on those 2 test to 5 seconds. It still failed for the cross docker test but yet seems random as the previous run worked for that test.

PaulMiami:
@yurishkuro I did the same experiment from master locally and the same thing happens, once in a while those 2 tests (http and udp 'sender should gracefully handle errors emitted by socket.send') take a little over 3 seconds to complete

yurishkuro:
if it's a flaky test we shouldn't block this PR. I restarted the build, let's see if it passes.

jmoney8080:
@yurishkuro did the build pass?  Any way I can help with this PR.  I'm very interested in seeing this be approved and merged.

rafabene:
> @yurishkuro did the build pass? Any way I can help with this PR. I'm very interested in seeing this be approved and merged.

I'm also waiting for this PR so I can use this library together with Istio.
What is it needed to be merged?

yurishkuro:
I will look at it soon, since we're doing a lot of work in Node client right now.

src/span_context.js
```diff
@@ -68,7 +68,20 @@ export default class SpanContext {
...
+      // corresponding 64 or 128 bits buffer by padding the start with zeros if necessary
+      // https://github.com/jaegertracing/jaeger/issues/1657
+      // At the same time this will enforce that the HEX has an even number of digits, node is expecting 2 HEX characters per byte
+      // https://github.com/nodejs/node/issues/21242
+      const traceIdExactLength = this._traceIdStr.length > 16 ? 32 : 16;
+      if (this._traceIdStr.length === traceIdExactLength) {
+        this._traceId = Utils.newBufferFromHex(this._traceIdStr);
+      } else {
+        const paddings = ['0000000000000000', '00000000000000000000000000000000'];
+        const paddingIndex = traceIdExactLength === 16 ? 0 : 1;
```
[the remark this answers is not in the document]
PaulMiami: ok

PaulMiami:
@yurishkuro I merged master

PR #372, opened 2019-08-22
All spans of a trace share sampling state

tiffon:

Lays the ground work for making sampling decisions based on span tags. These sampling decisions should affect all spans of a trace (within a given process) regardless of which span triggered the decision.

### Sampler API

Enhance the sampler API to be able to make sampling decisions on:

- Span creation
- Setting the operation name (aside from the initial value)
- Setting tags

Add `src/_flow/legacy_sampler_v1.js` to capture the legacy sampler API.

Redefine `src/_flow/sampler.js` to match the new sampler API.

Implement `src/samplers/v2/const_sampler.js` with the new API.

Implement `src/samplers/adapt_sampler.js` to wrap legacy samplers with the new API.

### Share sampled state within a trace

Share sampling state across all span contexts of the same trace.

Add `src/samplers/v2/sampling_state.js` to share sampling state among spans of the same trace.

Revise `src/span_context.js` to use `SamplingState` instead of `_flags` and `_samplingFinalized`. Update various classes that referred to these fields.

### DIFF
--- a/src/_flow/legacy_sampler_v1.js
+++ b/src/_flow/legacy_sampler_v1.js
@@ -1,5 +1,5 @@
 // @flow
-// Copyright (c) 2016 Uber Technologies, Inc.
+// Copyright (c) 2019 Uber Technologies, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
 // in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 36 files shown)


src/_flow/legacy_sampler_v1.js
```diff
@@ -0,0 +1,41 @@
+// @flow
+// Copyright (c) 2016 Uber Technologies, Inc.
```
vprithvi: Should this be `2019`?

tiffon: The file has been renamed.

tiffon: Will update it.

yurishkuro:
closing in favor of #377

PR #377, opened 2019-08-26
All spans of a trace share sampling state

yurishkuro:
Replaces #372


src/tracer.js
```diff
@@ -218,53 +226,39 @@ export default class Tracer {
...
+          ctx._setDebug(true);
           internalTags[constants.JAEGER_DEBUG_HEADER] = parent.debugId;
         }
         // baggage that could have been passed via `jaeger-baggage` header
         ctx.baggage = parent.baggage;
       }
-
-      ctx.traceId = randomId;
-      ctx.spanId = randomId;
-      ctx.parentId = null;
```
vprithvi: We have a test in the adaptor that explicitly checks that `parentId` is `null` (as opposed to `undefined`) for a newly created span - do you think this behavior is expected and relied upon elsewhere else?

yurishkuro: we should restore it, probably assign null explicitly in the constructor

Issue #379, opened 2019-08-29
Support delayed sampling


yurishkuro:
Per meta-ticket jaegertracing/jaeger#1861

## Requirement - what kind of business use case are you trying to solve?

* Develop custom sampler that can make sampling decisions based on tags of the span that can be set after span creation (similar to late binding of setOperation, but even later, possibly after child spans have been already created)
* Allow the samplers to set the firehose mode

## Problem - what in Jaeger blocks you from solving the requirement?

Deferred sampling (aka "finalizing sampling decision") currently only extends until setOperation() is called or span context is used for any other purpose like injecting it into a carrier or creating a child span.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Allow un-finalized sampling decision to be indefinite. When creating child spans or propagating the context this has the effect of reporting sampling=no, but it can later change to yes for future spans.

## Related issues

#260

yurishkuro:
Released in 3.16.0

PR #380, opened 2019-08-29
Support delayed sampling

yurishkuro:

- Support delayed sampling #379

- Sampling `finalized` mode is no longer early triggered by read operations like injecting context or starting a child span. Instead, sampling may remain unfinalized indefinitely if at least one of the samplers returns `retryable=true`.
- Introducing `remote` flag in SpanContext that is set when the context has been extracted from a carrier/request. When a child span has a remote context as a parent, the child sampling is finalized immediately, thus preserving the existing behavior of respecting the upstream sampling decision.
- Most samplers are upgraded to V2
  - Remote and PerOperation samplers are completely switched to V2, because they always delegate to the underlying sampler and isSampled() cannot be adequately delegated to V2 API
  - Primitive samplers implement both V1 and V2
  - GuaranteedThroughputSampler is kept at V1 because it delegates to primitive ones using V1 API (there's no benefit for it to use V2)
- Primitive samplers now return `retryable=false` from `onCreateSpan`. If someone wants a deferred sampling behavior, then some samplers should be returning `retryable=true`.
- Adding experimental tag-sampler and priority-sampler


src/samplers/probabilistic_sampler.js
```diff
@@ -11,12 +11,14 @@
...
-import * as constants from '../constants.js';
+import * as constants from '../constants';
+import LegacySamplerV1Base from './_adapt_sampler';
 
-export default class ProbabilisticSampler implements LegacySamplerV1 {
+export default class ProbabilisticSampler extends LegacySamplerV1Base implements LegacySamplerV1 {
   _samplingRate: number;
 
   constructor(samplingRate: number) {
+    super('ProbabilisticSampler');
```
vprithvi: Is calling by name the preferred way of invoking super? It seems pretty fragile

src/samplers/probabilistic_sampler.js
```diff
@@ -26,6 +28,14 @@ export default class ProbabilisticSampler implements LegacySamplerV1 {
     this._samplingRate = samplingRate;
   }
 
+  update(samplingRate: number): boolean {
+    if (this._samplingRate == samplingRate) {
```
vprithvi: Is there an equality function which takes in an epsilon for float comparisons?

src/span_context.js
```diff
@@ -26,6 +26,7 @@ export default class SpanContext {
   _baggage: any;
   _debugId: ?string;
   _samplingState: SamplingState;
+  _remote: boolean; // indicates that span context represents a remote parent
```
vprithvi: Replace with `isRemoteParent`?

src/tracer.js
```diff
@@ -246,7 +246,9 @@ export default class Tracer {
       hasValidParent = true;
       let spanId = Utils.getRandom64();
       ctx = parent._makeChildContext(spanId);
-      ctx.finalizeSampling(); // will finalize sampling for all spans sharing this traceId
+      if (parent.isRemote()) {
+        ctx.finalizeSampling(); // will finalize sampling for all spans sharing this traceId
```
vprithvi: IMHO the comment is redundant when the reader knows ctx is shared

test/samplers/delayed_sampling.js
```diff
@@ -0,0 +1,223 @@
...
+
+    _getState(span: Span): PrioritySamplerState {
+      const store = span.context()._samplingState.extendedState();
+      const stateKey = 'PrioritySampler'; // TODO ideally should be uniqueName() per BaseSamplerB2
+      let state: ?PrioritySamplerState = store[stateKey];
+      if (!state) {
+        state = {
+          samplerFired: Array(this._delegates.length),
+        };
+        // cannot use array.fill() in Node 0.10
```
vprithvi: Are tests written in `Node 0.10`?

test/samplers/delayed_sampling.js
```diff
@@ -0,0 +1,223 @@
...
+  class PrioritySampler implements Sampler {
+    apiVersion = SAMPLER_API_V2;
+
+    _delegates: Array<Sampler>;
+
+    constructor(samplers: Array<Sampler | LegacySamplerV1>) {
+      this._delegates = samplers.map(s => adaptSamplerOrThrow(s));
+    }
+
+    _getState(span: Span): PrioritySamplerState {
```
vprithvi: `s/_getState/_getOrCreateState`?

test/samplers/delayed_sampling.js
```diff
@@ -0,0 +1,223 @@
...
+      const store = span.context()._samplingState.extendedState();
+      const stateKey = 'PrioritySampler'; // TODO ideally should be uniqueName() per BaseSamplerB2
+      let state: ?PrioritySamplerState = store[stateKey];
+      if (!state) {
+        state = {
+          samplerFired: Array(this._delegates.length),
+        };
+        // cannot use array.fill() in Node 0.10
+        for (let i = 0; i < this._delegates.length; i++) {
+          state.samplerFired[i] = false;
```
vprithvi: Can't this be encapsulated within `PrioritySamplerState`

test/samplers/delayed_sampling.js
```diff
@@ -0,0 +1,223 @@
...
+
+    close(callback: ?Function): void {
+      if (callback) {
+        callback();
+      }
+    }
+  }
+
+  declare type PrioritySamplerState = {
+    samplerFired: Array<boolean>,
```
vprithvi: How is `samplerFired` different than `retryable`?

test/samplers/delayed_sampling.js
```diff
@@ -0,0 +1,223 @@
...
+
+    beforeEach(() => {});
+
+    it('should not sample or finalize new span without tags', () => {
+      let span = tracer.startSpan('opName');
+      assert.isFalse(span._spanContext.isSampled(), 'sampled');
+      assert.isFalse(span._spanContext.samplingFinalized, 'finalized');
+    });
+
+    it('should sample and finalize created span with tag', () => {
```
vprithvi: Might be better to nest these under a "matching tag(s)" describe block

test/span.js
```diff
@@ -244,166 +252,180 @@ describe('span should', () => {
     assert.equal(prevTagLength, span._tags.length, 'The sampling.priority tag should only be set once');
   });
 
-  describe('adaptive sampling tests for span', () => {
```
yurishkuro: moved to top level test below

test/samplers/delayed_sampling.js
```diff
@@ -0,0 +1,223 @@
...
+      let span = tracer.startSpan('opName', { tags: { theWho: 'Bender' } });
+      assert.isTrue(span._spanContext.isSampled(), 'sampled');
+      assert.isTrue(span._spanContext.samplingFinalized, 'finalized');
+    });
+
+    it('should sample and finalize span after setTag', () => {
+      let span = tracer.startSpan('opName');
+      assert.isFalse(span._spanContext.isSampled(), 'sampled');
+      assert.isFalse(span._spanContext.samplingFinalized, 'finalized');
+      span.setTag('theWho', 'Leela');
```
vprithvi: Is this wip? Seems to be missing asserts

vprithvi:
I think the API looks pretty usable

src/samplers/probabilistic_sampler.js
yurishkuro: it simply passes the "name" that is used in toString and uniqueName() functions.

src/samplers/probabilistic_sampler.js
yurishkuro: not that I know off. I don't think it matters here. If the rate is coming from JSON strategy, then float representation will be the same as long as the string representation is the same.

src/span_context.js
yurishkuro: In Java Beans naming we don't add `is` to the field names. And `parent` would be incorrect here because it's _this span context_ that is "remote".

test/samplers/delayed_sampling.js
yurishkuro: +1

test/samplers/delayed_sampling.js
yurishkuro: +1

test/samplers/delayed_sampling.js
yurishkuro: samplerFired is set once the given sampler returns retryabe=false. I will add a type comment.

test/samplers/delayed_sampling.js
```diff
@@ -0,0 +1,223 @@
...
+
+  /**
+   * PrioritySampler contains a list of samplers that it interrogates in order.
+   * Sampling methods return as soon as one of the samplers returns sample=true.
+   * The retryable state for each underlying sampler is stored in the extended context
+   * and once retryable=false is returned by one of the delegates it will never be
+   * called against.
+   *
+   * TODO: since sampling state is shared across all spans of the trace, the execution
+   *       of the sampler should probably only happen on the local-root span.
```
yurishkuro: open question

test/samplers/delayed_sampling.js
yurishkuro: we test with all versions

test/samplers/delayed_sampling.js
yurishkuro: fixed

src/util.js
```diff
@@ -148,4 +148,18 @@ export default class Utils {
         error(err);
       });
   }
+
+  /**
+   * Creates a callback function that only delegates to passed <code>callback</code>
+   * after <code>limit</code> invocations.
```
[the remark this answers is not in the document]
yurishkuro: +1

src/tracer.js
yurishkuro: and if the reader forgets it, here's a reminder which is critical to understanding this piece

src/samplers/experimental/tag_equals_sampler.js
```diff
@@ -0,0 +1,95 @@
...
+  onCreateSpan(span: Span): SamplingDecision {
+    const tag: ?Tag = this._findTag(span.getTags());
+    if (tag) {
+      return this._decide(span, tag.value);
+    }
+    return this._undecided;
+  }
+
+  onSetOperationName(span: Span, operationName: string): SamplingDecision {
+    return this.onCreateSpan(span);
```
yurishkuro: @vprithvi  maybe this should be always `return this._undecided`, because this sampler is not sensitive to operation name.

test/samplers/exp_extend_remote_sampler.js
```diff
@@ -0,0 +1,109 @@
...
+      if (strategy.strategyType == 'extended' && strategy.tagEqualsStrategy) {
+        let tagSampler = TagEqualsSampler.fromStrategy(strategy.tagEqualsStrategy);
+        if (this._sampler instanceof PrioritySampler) {
+          this._sampler = this._sampler._delegates[1];
+        }
+        super._updateSampler(strategy);
+        this._sampler = new PrioritySampler([tagSampler, this._sampler]);
+        return true;
+      }
+      return super._updateSampler(strategy);
```
yurishkuro: @vprithvi this is working, but is hacky, because `super._updateSampler` works in-place, so I have to do tricks with overriding `this._sampler` multiple times. Also, the tag & priority samplers are replaced every time, which is not necessarily a problem since they are both stateless and we don't run updates frequently.

test/samplers/exp_extend_remote_sampler.js
```diff
@@ -0,0 +1,109 @@
...
+    });
+  });
+
+  afterEach(() => {
+    remoteSampler.close();
+  });
+
+  it('should parse extended strategy response', function(done) {
+    server.addStrategy('service1', {
+      strategyType: 'extended',
```
yurishkuro: @vprithvi thoughts?

test/samplers/exp_extend_remote_sampler.js
```diff
@@ -0,0 +1,109 @@
...
+      metrics: metrics,
+      logger: logger,
+    });
+  });
+
+  afterEach(() => {
+    remoteSampler.close();
+  });
+
+  it('should parse extended strategy response', function(done) {
```
yurishkuro: as a POC this works, but we probably need to add more tests to ensure that the extended remote sampler can switch back and forth between extended and normal adaptive behavior.

test/samplers/exp_extend_remote_sampler.js
vprithvi: These trade offs seem reasonable for our use case

test/samplers/exp_extend_remote_sampler.js
vprithvi: Not a fan of this in the short term bc it would mean that the backend implementation would need to modify the payload from the thrift sampling object (or modify it after it has been converted to JSON).

yurishkuro: The easiest way to avoid it is to nest the "classic" response under a property.

src/samplers/experimental/tag_equals_sampler.js
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  /**
+   * Creates the sampler from a JSON configuration of the following form:
+   * <code>
+   *   {
+   *     key: 'taKey',
+   *     values: {
+   *       'tagValue1': {
+   *         firehose: true,
```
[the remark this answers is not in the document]
yurishkuro: Usually "disabled" flags are good when you have a historical default of "true" and you want to make it configurable. In this case we don't have any default, and I don't know if there's a clear winner which one should be the default. It feels like firehose is an extra functionality that use should specifically request (our internal use case notwithstanding).

test/samplers/exp_extend_remote_sampler.js
yurishkuro: just made that change

yurishkuro:
Part of #379.

PR #381, opened 2019-09-03
feat(propagator): inject value into carrier through set method of carrier if exists

edvardchen:

- Carrier like [grpc metadata](https://github.com/grpc/grpc-node/blob/master/packages/grpc-native-core/index.d.ts#L529) can only be inserted value through its `set` or `update` methods. We should respect them.

-  inject value into carrier through set method of carrier if exists

### DIFF
--- a/src/propagators/text_map_codec.js
+++ b/src/propagators/text_map_codec.js
@@ -59,6 +59,14 @@ export default class TextMapCodec {
     }
   }
 
+  _set(carrier: any, key: string, value: string) {
+    if (carrier.set) {
+      carrier.set(key, value);
+    } else {
+      carrier[key] = value;
+    }
+  }
+
   extract(carrier: any): ?SpanContext {
     let spanContext = new SpanContext();
     let baggage = {};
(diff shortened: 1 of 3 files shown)


yurishkuro:
This would not be compliant with the OpenTracing API that says the carrier is always a Javascript Object. 

https://github.com/opentracing/opentracing-javascript/blob/master/src/constants.ts#L18

We cannot assume that `set` method is actually for setting keys and not some other business method.

edvardchen:
I thought the carrier was not actually identical to the `SpanContexts`.

Why `set` method? I had kook the implementation of [jaeger-go](https://github.com/jaegertracing/jaeger-client-go/blob/master/propagation.go#L121) which injected value through `set` method. 
Turn out that the `API` [opentracing-go](https://github.com/opentracing/opentracing-go/blob/master/propagation.go#L95) that `jaeger-client-go` follows is a little different than `opentracing-javascript`.

Anyway, I think that adding a little code to respect `set` method is not harmful.

Otherwise people who use this repo with [grpc-node](https://github.com/grpc/grpc-node/tree/master/packages/grpc-native-core) would have to write something like this:

```javascript
      tracer.inject(
        span.context(),
        FORMAT_HTTP_HEADERS,
        new Proxy(grpcMeta, {
          set(target, key: string, value) {
            target.set(key, value);
            return true;
          },
        })
      );
```
Looks awful, isn't it?

yurishkuro:
> Anyway, I think that adding a little code to respect set method is not harmful.

It is harmful because it is unexpected behavior not prescribed by the OT API. 

Using a proxy doesn't seem awful at all, it's what most OT implementations use in many languages, because OT APIs define only a single carrier interface, which always needs to be adapted to the actual API of the "headers" object in the request. In many HTTP frameworks in Javascript and Python the headers happen to be just a dictionary, so the OT API defines carrier interface as a dictionary (but in Go it defines a dedicated interface).

What is your concern with using the Proxy?

yurishkuro:
No follow-up from the author, closing.

PR #384, opened 2019-09-05
Support configuring sampling url

vprithvi:

- Allows users to specify a custom sampling endpoint on the agent so that they may experiment easily.

- Add a configuration that takes in sampling path, defaulting to `/sampling`


src/samplers/remote_sampler.js
```diff
@@ -118,7 +122,7 @@ export default class RemoteControlledSampler implements Sampler {
       this._logger.error(`Error in fetching sampling strategy: ${err}.`);
       this._metrics.samplerQueryFailure.increment(1);
     };
-    Utils.httpGet(this._host, this._port, `/sampling?service=${serviceName}`, success, error);
+    Utils.httpGet(this._host, this._port, `${this._samplingPath}?service=${serviceName}`, success, error);
```
yurishkuro: +1

but fwiw, this is soooo broken - what if service name has a space in it? We should be urlencoding URL params.

PR #396, opened 2019-09-13
udp6 support

doochik:

Jaegler client supports udp4 socket only.

I've added `agentSocketType` to reporter config to add support for IPv6 udp sockets.
https://nodejs.org/dist/latest-v10.x/docs/api/dgram.html#dgram_dgram_createsocket_type_callback

Example
```
const jaegerConfig = {
    reporter: {
        agentHost: 'jaeger-agent.query.consul',
        agentPort: 5775,
        agentSocketType: 'udp6',
    },
    ....
};
```


doochik:
> Love the PR.
> 
> Q: what would be the reasons for someone to prefer v6 over v4? Does it matter on the localhost for UDP? If there are good reasons, perhaps mention them in the readme when describing the new env var, as recommendation to the user.

I have external (not localhost) ipv6-only jaeger-agent. So I can't use udp4.

src/configuration_env.js
```diff
@@ -149,6 +149,15 @@ export default class ConfigurationEnv {
       reporterConfig.agentPort = parseInt(value);
     }
 
+    value = ConfigurationEnv._getConfigValue(
+      config.reporter,
+      'agentSocketType',
+      process.env.JAEGER_AGENT_SOCKET_TYPE || process.env.JAEGER_REPORTER_AGENT_SOCKET_TYPE
```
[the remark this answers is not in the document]
doochik: Yes, it's copy/paste. Which name is legacy?

test/udp_sender.js
```diff
@@ -274,3 +275,87 @@ describe('udp sender', () => {
     });
   }).timeout(5000);
 });
+
+describe('udp6 sender', () => {
```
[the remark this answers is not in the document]
doochik: ok, I'll fix it

src/configuration_env.js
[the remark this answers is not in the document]
yurishkuro: `JAEGER_REPORTER_*`

yurishkuro:
> I have external (not localhost) ipv6-only jaeger-agent. So I can't use udp4.

We don't recommend non-local agents, the agent wasn't designed for this. For sending spans to a remote received it's better to use TCP-based protocols, e.g. HTTP directly to the collector.

doochik:
@yurishkuro I've fixed all of you comments

src/configuration.js
```diff
@@ -143,6 +144,10 @@ export default class Configuration {
       if (config.reporter.agentPort) {
         senderConfig['port'] = config.reporter.agentPort;
       }
+
+      if (config.reporter.agentSocketType) {
+        senderConfig['socketType'] = config.reporter.agentSocketType;
```
yurishkuro: could we add a test for this?

src/configuration_env.js
```diff
@@ -149,6 +149,15 @@ export default class ConfigurationEnv {
       reporterConfig.agentPort = parseInt(value);
     }
 
+    value = ConfigurationEnv._getConfigValue(
+      config.reporter,
+      'agentSocketType',
+      process.env.JAEGER_AGENT_SOCKET_TYPE
+    );
+    if (value) {
+      reporterConfig.agentSocketType = value;
```
yurishkuro: could we add a test for this?

test/udp_sender.js
```diff
@@ -26,251 +26,269 @@ import ThriftUtils from '../src/thrift.js';
...
-        tracer.close(done);
-      }
+    it('should return 0,undefined on flush() with no spans', () => {
+      sender.flush(assertCallback(0, undefined));
     });
-  }).timeout(5000);
-});
+
+    it('should gracefully handle errors emitted by socket.send', function(done) {
+      // EAI_AGAIN - nodejs received invalid DNS response. E.g. resolver doesn't support IPv6.
```
yurishkuro: When would this error happen in the test?

test/udp_sender.js
```diff
@@ -26,251 +26,269 @@ import ThriftUtils from '../src/thrift.js';
...
+      // ENOTFOUND - nodejs received valid DNS response but domain not found
+      // ESRCH - nodejs v0.10 error
```
yurishkuro: is that specifically for udp6? Since before the tests were passing with just ENOTFOUND

test/udp_sender.js
doochik: This errors occurs when dns resolver doesn't support IPv6 or host doesn't have ipv6 resolver. I've got this error in travis for upd6 and `foo.bar.xyz` instead of `ENOTFOUND`.

test/udp_sender.js
doochik: Yes, is specifically for upd6

test/udp_sender.js
doochik: For example,
on my local Mac I've got `ENOTFOUND` for this domain because I have dns resolver.
In travis I've got `EAI_AGAIN`

doochik:
Red tests because of this regression https://github.com/Raynos/error/pull/19

deps: jaeger-client-node -> tchannel, thriftrw -> bufrw -> error

doochik:
Tests are green

doochik:
@yurishkuro could you release new version please?

PR #397, opened 2019-09-16
initTracer should pass options.traceId128bit

doochik:

128 bit traceid was added at https://github.com/jaegertracing/jaeger-client-node/pull/361
But I can't do like this
`const tracer = require('jaeger-client').initTracer(config, { traceId128bit: true });`.

- pass `options.traceId128bit` to `new Tracer`


src/configuration.js
```diff
@@ -191,6 +191,8 @@ export default class Configuration {
    * @param {Object} [options.logger] - a logger (see ./_flow/logger.js)
    * @param {Object} [options.tags] - set of key-value pairs which will be set
    *        as process-level tags on the Tracer itself.
+   * @param {boolean} [options.traceId128bit] - generate root span with a 128bit traceId.
+   *        BaggageRestrictionManager API from ./baggage.js.
```
yurishkuro: what is this line?

doochik: Oops )
Remove this line

PR #398, opened 2019-09-17
Do not strip leading zeros

### LINKED ISSUE #391: Do not strip leading zeros
Parent issue describes why: https://github.com/jaegertracing/jaeger/issues/1657

doochik:
Fixes #391


src/span_context.js
```diff
@@ -53,19 +53,7 @@ export default class SpanContext {
...
-      // https://github.com/nodejs/node/issues/21242
-      const traceIdExactLength = this._traceIdStr.length > 16 ? 32 : 16;
-      if (this._traceIdStr.length === traceIdExactLength) {
-        this._traceId = Utils.newBufferFromHex(this._traceIdStr);
-      } else {
-        const padding = traceIdExactLength === 16 ? '0000000000000000' : '00000000000000000000000000000000';
-        const safeTraceIdStr = (padding + this._traceIdStr).slice(-traceIdExactLength);
-        this._traceId = Utils.newBufferFromHex(safeTraceIdStr);
-      }
+      this._traceId = Utils.newBufferFromHex(this._traceIdStr);
```
yurishkuro: I don't think I like this part. We have backwards compatibility requirement to parse strings with stripped zeros (potentially coming from other clients). That means `_traceidStr` could be such trimmed ID. If we convert it directly to byte buffer, the buffer could be of random length. I liked the code above that made sure the buffer is always of precisely known length.

doochik: Do you mean case where I manually set `trace.context().traceIdStr = 'abc'` ?

In other cases `SpanContext.withStringIds()` guarantees valid `traceIdStr`

yurishkuro: Fair enough. But wouldn't it be better to perform the padding in the constructor, instead of `SpanContext.withStringIds()`?

test/tracer.js
```diff
@@ -51,10 +51,10 @@ describe('tracer should', () => {
     };
 
     let mycontext = mytracer.extract(opentracing.FORMAT_HTTP_HEADERS, headers);
-    assert.equal(mycontext.toString(), headers[ck]);
+    assert.equal(mycontext.toString(), '000000000000000a:000000000000000b:000000000000000c:d');
```
yurishkuro: +1

brandon-leapyear:
Any updates on this?

PR #410, opened 2019-10-22
Do not apply adaptive sampler to child spans

yurishkuro:

Per-operation adaptive sampling should only apply to service endpoints, not to operation names of the inner/children spans. Recent changes (e.g. #380) broke that functionality and traces can now be sampled by the per-operation sampler triggering on children spans. This is a problem for adaptive sampling backend that only records operations from the root spans, meaning all child operations are getting the default probability, which may be higher than the calculated probabilities.

- Add a test that demonstrates the problem.
- Change per-operation sampler to only allow sampling "local root spans", which is equivalent to the actual root span of the trace because spans with remote parent are finalized right away.

### DIFF
--- a/src/samplers/per_operation_sampler.js
+++ b/src/samplers/per_operation_sampler.js
@@ -96,7 +96,8 @@ export default class PerOperationSampler implements Sampler {
       return { sample: false, retryable: false, tags: outTags };
     }
     let isSampled = this.isSampled(span.operationName, outTags);
-    // returning retryable=true since we can change decision after setOperationName().
+    // returning retryable=true since we can change the sampling decision
+    // after the first call to setOperationName()
     return { sample: isSampled, retryable: true, tags: outTags };
   }
 
--- a/src/samplers/per_operation_sampler.js
+++ b/src/samplers/per_operation_sampler.js
@@ -92,13 +92,19 @@ export default class PerOperationSampler implements Sampler {
 
   onCreateSpan(span: Span): SamplingDecision {
     const outTags = {};
-    const isSampled = this.isSampled(span.operationName, outTags);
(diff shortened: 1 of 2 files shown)


test/samplers/remote_sampler.js
```diff
@@ -192,6 +194,50 @@ describe('RemoteSampler', () => {
...
+
+      let sp0 = tracer.startSpan('op2');
+      assert.isTrue(sp0.context().isSampled(), 'op2 should be sampled on the root span');
+
+      sp0 = tracer.startSpan('op1');
+      assert.isFalse(sp0.context().isSampled(), 'op1 should not be sampled');
+      sp0.setOperationName('op2');
+      assert.isTrue(sp0.context().isSampled(), 'op2 should be sampled on the root span');
+
+      let sp1 = tracer.startSpan('op1', 'op1 should not be sampled');
```
vprithvi: Could we use more descriptive names? parent/child or some such?

src/samplers/per_operation_sampler.js
```diff
@@ -92,13 +92,19 @@ export default class PerOperationSampler implements Sampler {
 
   onCreateSpan(span: Span): SamplingDecision {
     const outTags = {};
-    const isSampled = this.isSampled(span.operationName, outTags);
-    // NB: return retryable=true here since we can change decision after setOperationName().
+    if (!span.context()._samplingState.isLocalRootSpan(span.context())) {
+      return { sample: false, retryable: false, tags: outTags };
+    }
+    let isSampled = this.isSampled(span.operationName, outTags);
+    // returning retryable=true since we can change decision after setOperationName().
```
vprithvi: I think this is a bit confusing because we set retryable 'false' onSetOperationName - perhaps "the decision can change the first time setOperationName is called?"

PR #415, opened 2019-12-24
[HTTPSender] Clear span buffer if Thrift encoding fails

### LINKED ISSUE #414: HTTPSender.flush: spans hasn't been emptied when encode failed
Hi, we got an OOM in production env recently and get lots of logs like below through enable the logger.

`Failed to append spans in reporter: Error encoding Thrift: operationName: missing required field "operationName" with id 5 on Span`

There is no doubt, the span has the problem. 

However, the HTTPSender should empty the send queue when catch error. If not, the send queue will be bigger and bigger beacause the RemoteReporter continue to append span to send queue.

himStone:
Resolves #414 

HTTPSender empty the send queue when catch error to avoid OOM

### DIFF
--- a/src/reporters/http_sender.js
+++ b/src/reporters/http_sender.js
@@ -102,13 +102,12 @@ export default class HTTPSender {
     }
 
     const result = this._jaegerThrift.Batch.rw.toBuffer(this._batch);
+    this._reset(); // clear buffer for new spans, even if Thrift conversion fails
+
     if (result.err) {
-      this._reset();
       SenderUtils.invokeCallback(callback, numSpans, `Error encoding Thrift batch: ${result.err}`);
       return;
     }
-
-    this._reset();
 
     const requester = this._url.protocol === 'https:' ? https.request : http.request;
 
--- a/src/reporters/http_sender.js
+++ b/src/reporters/http_sender.js
@@ -103,6 +103,7 @@ export default class HTTPSender {
 
     const result = this._jaegerThrift.Batch.rw.toBuffer(this._batch);
     if (result.err) {
+      this._reset();
       SenderUtils.invokeCallback(callback, numSpans, `Error encoding Thrift batch: ${result.err}`);
       return;
     }
--- a/test/http_sender.js
+++ b/test/http_sender.js
@@ -229,6 +229,7 @@ describe('http sender', () => {
     sender.flush((numSpans, err) => {
       assert.equal(numSpans, 1);
       expect(err).to.have.string('Error encoding Thrift batch:');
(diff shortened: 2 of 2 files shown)


himStone:
@yurishkuro here is the unit test that demonstrates this issue

```
  it('should avoid OOM on failed buffer conversion', done => {
    function createSpan(name) {
      const span = tracer.startSpan(name);
      span.finish(); // finish to set span duration
      return span
    }

    const maxSpanBatchSize = 5

    sender = new HTTPSender({
      endpoint: serverEndpoint,
      maxSpanBatchSize,
    });
    sender.setProcess(reporter._process);

    // simulate the server use jaeger client
    let num = 0
    const intervalHandle = setInterval(() => {
      if (num === 14) {
        const wrongSpan = createSpan() // name is required, make a mistake in create span, which will cause serialization error in flush
        sender.append(ThriftUtils.spanToThrift(wrongSpan))
      } else {
        sender.append(ThriftUtils.spanToThrift(createSpan(num)))
      }
      num++
    }, 1);

    setTimeout(() => {
      clearInterval(intervalHandle)

      // check the send queue
      assert.equal(sender._batch.spans.length < maxSpanBatchSize, true);

      sender.flush((numSpans, err) => {
        assert.equal(sender._batch.spans.length, 0);
        done();
      });
    }, 1000);
  });
```

Thanks for given such awesome tracing system to community. 
I think this pr will make the Node.js client more robust. Is there anything else I may do?

yurishkuro:
can you add this test to `test/http_sender.js`?

himStone:
@yurishkuro done

src/reporters/http_sender.js
```diff
@@ -103,6 +103,7 @@ export default class HTTPSender {
 
     const result = this._jaegerThrift.Batch.rw.toBuffer(this._batch);
     if (result.err) {
+      this._reset();
```
yurishkuro: I think it's better to move this to the next line after toBuffer() call, and remove the same statement  from L111.

test/http_sender.js
```diff
@@ -283,4 +284,44 @@ describe('http sender', () => {
       done();
     });
   });
+
+  it('should avoid OOM on failed buffer conversion', done => {
```
yurishkuro: On closer inspection, I don't think this test adds much in addition to the assert in L232, and it also runs for a long time (5s), so let's remove this.

himStone: That's what I thought at first. So I both add the assert in L232 and put this unit test as a comment which may be helpful to understand the issue.

PR #424, opened 2020-03-02
Generate JavaScript module from thrift IDLs

jamietre:
#303 
 Which problem is this PR solving?

Resolves [Issue 421](https://github.com/jaegertracing/jaeger-client-node/issues/421)  (client can't be used in webpack builds)

- Add module to read thrift IDLs and generate a JavaScript module with an export named for each IDL path
- Add node CLI bootstrap to convert all IDLs in the client to a module using aforementioned module
- Add npm script to invoke node CLI to perform codegen
- Remove steps from makefile that copy IDLs to build destination
- Add step to makefile to run npm script
- update .gitignore to exclude codegen target "src/generated"

This PR adds JavaScript as part of the build process, which doesn't have an established convention for file locations in this repo. I took a stab, but this might not make sense, so I welcome guidance.

The code that parses the IDL into a JavaScript module includes a pretty trivial mechanism to strip out comments from the IDL source. They would become part of the memory footprint of this client when loaded statically. I believe this to be safe based on my reading of the IDL spec, but since this mechanism doesn't actually create an AST but just uses regexps, and the spec could change, this could be considered a risk. It seems likely tests would fail if this ever broke the IDL source, but the reason why might not be obvious. If there are any concerns about this, it's a pretty minor optimization which I'm can remove. Alternatives might be only remove very simple comments (line starts with hash), or alternatively, investigate some way to pre-parse the IDLs and distribute a compiled or optimized version. I couldn't see an obvious way to do this with `thriftrw-node` but perhaps there is.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,8 @@ let server = new TChannel({ serviceName: 'server' });
 server.listen(4040, '127.0.0.1');
 let serverThriftChannel = TChannelAsThrift({
   channel: server,
-  entryPoint: path.join(__dirname, 'thrift', 'echo.thrift'), // file path to a thrift file
+  entryPoint: 'thrift/echo.thrift', //  path to a thrift IDL
+  idls: ThriftData,
 });
 
 let perProcessOptions = {};
@@ -168,6 +169,7 @@ Outbound calls can be made in two ways, shown below.
 
 ```javascript
 import { TChannelBridge } from 'jaeger-client';
+import ThriftData from './generated/thrift';
 
 let bridge = new TChannelBridge(tracer);
 // Create the toplevel client channel.
@@ -181,7 +183,8 @@ let clientSubChannel = client.makeSubChannel({
 
 let encodedThriftChannel = TChannelAsThrift({
   channel: clientSubChannel,
-  entryPoint: path.join(__dirname, 'thrift', 'echo.thrift'), // file path to a thrift file
+  entryPoint: 'thrift/echo.thrift', // path to a thrift IDL
+  idls: ThriftData,
 });
 
 // wrap encodedThriftChannel in a tracing decorator
--- a/README.md
+++ b/README.md
@@ -148,8 +148,7 @@ let server = new TChannel({ serviceName: 'server' });
 server.listen(4040, '127.0.0.1');
 let serverThriftChannel = TChannelAsThrift({
   channel: server,
-  entryPoint: 'thrift/echo.thrift', //  path to a thrift IDL
-  idls: ThriftData,
+  entryPoint: path.join(__dirname, 'thrift', 'echo.thrift'), // file path to a thrift file
 });
 
 let perProcessOptions = {};
@@ -169,7 +168,6 @@ Outbound calls can be made in two ways, shown below.
 
 ```javascript
 import { TChannelBridge } from 'jaeger-client';
-import ThriftData from './generated/thrift';
 
 let bridge = new TChannelBridge(tracer);
 // Create the toplevel client channel.
@@ -183,8 +181,7 @@ let clientSubChannel = client.makeSubChannel({
 
 let encodedThriftChannel = TChannelAsThrift({
   channel: clientSubChannel,
-  entryPoint: 'thrift/echo.thrift', // path to a thrift IDL
-  idls: ThriftData,
+  entryPoint: path.join(__dirname, 'thrift', 'echo.thrift'), // file path to a thrift file
 });
 
 // wrap encodedThriftChannel in a tracing decorator
--- a/crossdock/src/tchannel_server.js
+++ b/crossdock/src/tchannel_server.js
@@ -36,14 +36,10 @@ export default class TChannelServer {
     this._helpers = new Helpers(this._tracer);
 
     let serverChannel = TChannel({ serviceName: 'node' });
-
     let tchannelThrift = TChannelThrift({
       channel: serverChannel,
       source:
-        ThriftData[crossdockSpecPath] ||
-        (() => {
-          throw new Error(`${crossdockSpecPath} not found`);
-        }),
+        ThriftData[crossdockSpecPath],
     });
     let context = new DefaultContext();
 
--- a/crossdock/src/tchannel_server.js
+++ b/crossdock/src/tchannel_server.js
@@ -24,6 +24,7 @@ import TChannelBridge from '../../src/tchannel_bridge';
 import TChannel from 'tchannel';
 import TChannelThrift from 'tchannel/as/thrift';
 import Utils from '../../src/util.js';
+import ThriftData from '../../src/generated/thrift';
 
 let DEFAULT_THRIFT_PATH = '/crossdock/tracetest.thrift';
 export default class TChannelServer {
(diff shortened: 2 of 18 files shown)


jamietre:
Hmm, I'm getting a build failure from a crossdock test that doesn't fail for me locally, I'm guessing it's not running. Also a license check failure. Guessing the latter is `mock-fs` dependency I added; I could avoid this by testing in the real file system easily enough. Will need to look into the failed test.

crossdock/src/tchannel_server.js
```diff
@@ -35,9 +36,14 @@ export default class TChannelServer {
...
     let serverChannel = TChannel({ serviceName: 'node' });
+
     let tchannelThrift = TChannelThrift({
       channel: serverChannel,
-      entryPoint: crossdockSpecPath,
+      source:
+        ThriftData[crossdockSpecPath] ||
+        (() => {
+          throw new Error(`${crossdockSpecPath} not found`);
+        }),
```
yurishkuro: does this need trailing `()` in order for the function to be executed?

scripts/lib/javascriptifier.js
```diff
@@ -0,0 +1,164 @@
+// Copyright (c) 2017, The Jaeger Authors
```
yurishkuro: 2020

scripts/lib/javascriptifier.js
```diff
@@ -0,0 +1,164 @@
...
+}
+
+// for windows
+function slash(somePath) {
+  return somePath.replace(/\\/g, '/');
+}
+
+const commentRegexp = /^\s*(#|\/\*|\*|\*\/|\/\/).*$/;
+const inlineCommentRegexp = /^([^"']+)\s(#|\/\/).*$/;
+function stripCommentsFromThrift(text) {
```
yurishkuro: What is the objective of stripping comments? To make the embedded string literal smaller?

src/thrift.js
```diff
@@ -11,15 +11,16 @@
...
 
-import fs from 'fs';
 import opentracing from 'opentracing';
 import path from 'path';
 import { Thrift } from 'thriftrw';
 import Utils from './util.js';
+import ThriftData from './generated/thrift';
 
 export default class ThriftUtils {
   static _thrift = new Thrift({
```
yurishkuro: Seems like we're loading the same Thrift IDLs over and over, perhaps it's worth extracting everything into a single module (although not strictly in scope of this PR).

test/javascriptifier.js
```diff
@@ -0,0 +1,161 @@
+// Copyright (c) 2017, The Jaeger Authors
```
yurishkuro: 2020

README.md
```diff
@@ -148,7 +148,8 @@ let server = new TChannel({ serviceName: 'server' });
 server.listen(4040, '127.0.0.1');
 let serverThriftChannel = TChannelAsThrift({
   channel: server,
-  entryPoint: path.join(__dirname, 'thrift', 'echo.thrift'), // file path to a thrift file
+  entryPoint: 'thrift/echo.thrift', //  path to a thrift IDL
```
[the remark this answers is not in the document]
jamietre: I suppose not - I was just trying to make the usage here align with the examples; but yeah it might just be more confusing without context. Will revert.

crossdock/src/tchannel_server.js
jamietre: This was inline debugging I forgot to remove. It would fail anyway if any invalid path was used (and we don't do any kind of entry-point sanity check anywhere else) so definitely no needed.

src/thrift.js
jamietre: Agree, and probably should be another PR to avoid conflating the refactor with these changes.

scripts/lib/javascriptifier.js
jamietre: I mention this is the initial comment - yeah, it's an optimization. This change has the impact of making the IDLs become part of the memory footprint of this client, whereas they would not be when loaded & parsed dynamically. So the comments matter, and they are actually a fairly big part of the size of these files. 

It's probably not a big deal in the grand scheme - but seemed easy to just get rid of them. That said, there's a potential risk since this is a trivial way to remove them and doesn't involve actually parsing the syntax. I believe it to be safe, but if the spec for thrift IDL changes ever and there were some syntax (such as a non-comment leading slash or hash, possibly part of a multi-line statement?) then it might break. A better way seems like it would be to pre-parse the IDLs with `thriftrw` and emit some representation of the compiled AST instead of doing all this at all, but I didn't see any obvious way to "get compiled in and out" out of the thriftrw-node module, so that seemed to get way beyond the straightforward way of solving the basic problem that this PR seeks to solve.

In any event - stripping comments seems like a small risk, and I think if if failed, then tests would break, but not sure 🤷‍♂ ? Happy to do whatever you think is best here.

jamietre:
> There should be a way to exclude some files.

Ah OK - I misunderstood that error, thought that was a license check on a new npm package I consumed.  Should I just add the preamble to the generated code, or is there some way to exclude?

src/thrift.js
yurishkuro: Up to you. I would do it in this PR, since you're changing all the call sites already, so rather than passing file paths, you could simply import a symbol from a helper module.

scripts/lib/javascriptifier.js
yurishkuro: ok

yurishkuro:
looks like if there's a word "generated" in the file it will be skipped
https://github.com/jaegertracing/jaeger-client-node/blob/18a2d3479ff8a05b2cd853d3ca139c28bb224159/scripts/check-license.sh#L5

src/thrift.js
jamietre: That's a fair point. I'll give it a shot.

jamietre:
> looks like if there's a word "generated" in the file it will be skipped
> https://github.com/jaegertracing/jaeger-client-node/blob/18a2d3479ff8a05b2cd853d3ca139c28bb224159/scripts/check-license.sh#L5

Ha - the word GENERATED  *is* already in there but looks like it appears on line 4 which is just past what this checks. Will fix.

yurishkuro:
build is failing on crossdock tests with :
```
1) Endtoend Handler should "before each" hook for "report spans to local server":
     AssertionError [ERR_ASSERTION]: thriftrw-idl/agent.thrift: Thrift must be constructed with either a complete set of options.idls, options.asts, or options.fs access
      at Thrift._parse [as loadSync] (node_modules/thriftrw/thrift.js:216:16)
      at new Thrift (node_modules/thriftrw/thrift.js:145:14)
      at Context.<anonymous> (crossdock/test/endtoend_handler.js:34:14)
```

jamietre:
> build is failing on crossdock tests with :

I really apologize - we've had a crazy week like I imagine many have :/ and I haven't had time to get back to this. I noticed this test failing when I first made the PR, the test didn't fail for me locally. I the the crossdock tests aren't running for me locally. I don't think this will be hard to fix, but I probably won't be able to get back to it for a few more days until things settle down at work.

jamietre:
I'm going to close this until I have time to get back to it so it's out of your queue.

joneshf-cn:
@jamietre we ran into this when setting up Jaeger the other day. It looks like you did the majority of the work. Would you be okay with someone taking this the rest of the way? Happy to take it up, if so.

jamietre:
@joneshf-cn no problem at all! I am slammed at work so probably would be a while till I got back to it

Meemaw:
Any update on this? This is preventing various use cases for this library.

joneshf-cn:
Sorry, ended up not doing it an just copying the files over for our setup.

PR #441, opened 2020-08-14
Always read thrift defs from `./` to better support bundling

unstubbable:
Always reading the jaeger thrift definition file from the root dir (i.e. `src`) allows consumers to bundle `jaeger-client`, e.g. for uploading code to an AWS lambda function.

To bundle `jaeger-client` with `webpack` for example, the thrift file needs to be copied into the output directory of the bundle like this:

```js
{
  plugins: [
    new CopyPlugin([
      {
        from: require.resolve(
          'jaeger-client/dist/src/jaeger-idl/thrift/jaeger.thrift'
        ),
        to: 'jaeger-idl/thrift/jaeger.thrift'
      }
    ])
  ]
}
```

With this in place `fs.readFileSync` will be able to find the file.

Signed-off-by: Hendrik Liebau <mail@hendrik-liebau.de>

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,6 +267,23 @@ tracer.registerExtractor(opentracing.FORMAT_HTTP_HEADERS, codec);
 
 This can prove useful when compatibility with existing Zipkin tracing/instrumentation is desired.
 
+### Webpack Compatibility
+
+In order to bundle the library using webpack, e.g. for uploading code to an AWS Lambda function, it is required to copy the Jaeger thrift definition file into the output directory of the bundle:
+
+```js
+{
+  plugins: [
+    new CopyPlugin([
+      {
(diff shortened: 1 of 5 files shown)


unstubbable:
With my change it is guaranteed that the thrift file is always loading with a path that is relative to `__dirname` and continues with `./` (vs. `../` as before in two occurences). When a node application that requires `jaeger-client` is bundled into a single file, say `dist/my-bundle.js`, the thrift file can be resolved relative to the `__dirname` of the bundle, i.e. `dist` in my example. Therefore we can just copy the thrift file into the `dist` directory at `dist/jaeger-idl/thrift/jaeger.thrift` from where it can be read using `fs`.

yurishkuro:
released as https://github.com/jaegertracing/jaeger-client-node/releases/tag/v3.18.1

PR #448, opened 2020-11-06
Add config for maxUdpPacketSize

singron:

- The max udp packet size can differ between platforms. See https://github.com/jaegertracing/jaeger-client-node/issues/124#issuecomment-324222456
- In that thread, they increased the maximum packet size on the system, but sometimes it's easier to just have clients produce smaller packets if they can.

- Make the maximum udp packet size configurable.

The tests were also failing, so there is also a small commit to fix some invalid hostnames to be invalid.

### DIFF
--- a/src/configuration.js
+++ b/src/configuration.js
@@ -148,6 +148,9 @@ export default class Configuration {
       if (config.reporter.agentSocketType) {
         senderConfig['socketType'] = config.reporter.agentSocketType;
       }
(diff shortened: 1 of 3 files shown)


yurishkuro:
Any tests that can be added? Any changes to documentation required? IIRC there's a section in the README that specifically talks about packet sizes, may be worth mentioning this new config option there.

yurishkuro:
No response from the author, closing.

PR #459, opened 2021-04-06
Report HTTP errors when flushing spans

### LINKED ISSUE #458: HTTP errors on flush are not reported

When using the HTTP sender to send spans to jaeger and the endpoint is not correctly configured, as a developer I want to surface this information so I may address it.

If you have configured an HTTP sender that has a server on the other end that is receiving the request, but rejects the request with an HTTP error either due to a client error (incorrect path, incorrect payload) or a server problem (internal server error, gateway timeouts), no error is logged and the callback on the flush method does not indicate any error.

I propose HTTP errors (status code >= 400) are reported the same way as socket errors: `logger.error(message)`, as well as with the flush callback.

None that I can think of.

rexxars:

Fixes the issue where the server might respond with an error (4xx, 5xx) but no feedback is given to the developer that can help them diagnose the issue.

Resolves #458

- Uses the same pattern as on socket errors: builds an error message and calls the `logger.error()` method, as well as providing the error to the flush callback function.

### DIFF
--- a/src/reporters/http_sender.js
+++ b/src/reporters/http_sender.js
@@ -115,7 +115,14 @@ export default class HTTPSender {
     const req = requester(this._httpOptions, resp => {
       // consume response data to free up memory
       resp.resume();
-      SenderUtils.invokeCallback(callback, numSpans);
+
+      let error;
+      if (resp.statusCode >= 400) {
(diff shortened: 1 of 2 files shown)


yurishkuro (approved):
Thanks!

PR #461, opened 2021-04-22
Fix env parsing of falsy values (#460)

### LINKED ISSUE #460: ConfigurationEnv doesn't allow falsy values
We wanted to disable the tracer based on environment variables and some docs suggested to use `config.sampler = {type: 'const', param: 0}` instead of `config.disable = true`. Trying this we found that the parsing method `ConfigurationEnv._getConfigValue` doesn't handle falsy values (like `param: 0` above) very well:

https://github.com/jaegertracing/jaeger-client-node/blob/4c739a26d57c615fb4ac53eae921b9fd5ca53df2/src/configuration_env.js#L40-L42

The function will fall back to the default value, which in most cases is `process.env.*`, which in turn will be `undefined` if you don't provide it explicitly. `undefined` however is not an allowed value for many options (including `sampler.param`) which will result in errors. 

It would probably make more sense to explicitly check for `null` and / or `undefined` instead of doing an implicit boolean evaluation of the value.

Gerrit-K:
Signed-off-by: Gerrit Kieffer <kieffer.gerrit@gmail.com>

- Resolves #460

- Replace implicit boolean evaluation with explicit check in ternary operator

### DIFF
--- a/src/configuration_env.js
+++ b/src/configuration_env.js
@@ -38,7 +38,7 @@ export default class ConfigurationEnv {
   }
 
   static _getConfigValue(obj, key, defaultValue) {
-    return (obj && obj[key]) || defaultValue;
+    return obj && key in obj ? obj[key] : defaultValue;
   }
 
   static _getSamplerFromEnv(config) {
@@ -49,7 +49,7 @@ export default class ConfigurationEnv {
     }
 
     value = ConfigurationEnv._getConfigValue(config.sampler, 'param', process.env.JAEGER_SAMPLER_PARAM);
-    if (value) {
+    if (!isNaN(value)) {
       samplerConfig.param = parseFloat(value);
     }
 
--- a/test/init_tracer.js
+++ b/test/init_tracer.js
@@ -546,4 +546,27 @@ describe('initTracerFromENV', () => {
     assert.equal(tracer._tags['KEY2'], 'VALUE2');
     tracer.close(done);
   });
+
+  it('should parse falsy values from direct config setting', done => {
+    let config = {
+      serviceName: 'test-service-arg',
+      sampler: {
+        type: 'const',
+      },
+    };
+
+    let tracer;
+
+    expect(() => {
+      tracer = initTracerFromEnv(config);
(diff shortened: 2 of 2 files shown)


Gerrit-K:
Please note that I didn't test this thoroughly (apart from `npm test`) and I have absolutely no glimpse of how impactful this change can be, as I just started using Jaeger this week. I appreciate any feedback! Please let me know if I should add or change anything.

src/configuration_env.js
```diff
@@ -38,7 +38,7 @@ export default class ConfigurationEnv {
   }
 
   static _getConfigValue(obj, key, defaultValue) {
-    return (obj && obj[key]) || defaultValue;
+    return obj && key in obj ? obj[key] : defaultValue;
```
yurishkuro, line 41: could you please add a unit test that would fail without this change?

Gerrit-K: Done. And actually a very good hint as it uncovered another issue, which I fixed right away.

I now added a test with a `const` sampler and three essential assertions:
* the first one makes sure that the test still fails if no `param` is specified at all (so that my changes don't break this unintentionally)
* the second test checks that the original issue is now fixed (this one would fail prior to my commit)
* and the last one asserts that the value is parsed correctly

Would that be sufficient?

yurishkuro (approved):
Thanks

Gerrit-K:
I had to push the work again because I accidentally amended the commit with a different account, leaving the "committed by" field not matching the sign-off. Sorry for the inconvenience :(

And while I'm commenting: are the failing crossdock and node 4 checks something I should look for or is this currently broken in the main branch?

yurishkuro:
nothing should be failing, previous runs on master were green.

btw, I strongly recommend using named branches from PRs. The master branch in the fork should remain in sync with upstream. This could potentially confuse crossdock tests (they might try to upload docker images and fail on credentials which are not present in PR).

Gerrit-K:
Oh okay, I see. I'll change that, close this PR and re-open another one, linking to this.

PR #530, opened 2021-10-08
Simplify bundling udp_sender

punkfest:
Bundling with webpack / esbuild is hard due to the use of `__dirname + "../"` for the udp sender.

Add a helper function in src/thrift.js that returns an absolute path to
agent.thrift, without using "../". This is done to simplify bundling with
webpack or esbuild: assume an app is bundled to /app/index.js, if we resolve
a path to "../thrift-idl/etc", the file will be loaded from /thrift-idl/etc,
which is at the root of the filesystem.

Adjust README to support udp-sender.

This commit is a continuation of #441.


yurishkuro (approved):

punkfest:
Thank you! May I ask how often you tag and release changes to npm? No pressure, iirc I can patch the locally installed package with some override

yurishkuro:
released 3.19

